PR #14, opened 2024-06-07
Add trigger moderate full example as a doc

jonatas:
I was not sure if we start with an `examples` folder or just adding as docs. Probably generating some mkdocs could also be amazing or deploying the content directly to docs.timescale.com.

Let me know if I should also append the full script.

### DIFF
--- a/docs/trigger_moderate.md
+++ b/docs/trigger_moderate.md
@@ -1,0 +1,105 @@
+# Trigger moderate
+
+Let's say you want to moderate comments using OpenAI.
+
+You can get the full example in the [trigger_moderate.sql](examples/trigger_moderate.sql) file.
+
+First, let's create the extension and set the API key:
+
+```sql
+create extension if not exists ai cascade;
+```
+
+To set the API key, you can use the following command:
+```sql
+select set_config('ai.openai_api_key', :'OPENAI_API_KEY', false) is not null as set_config;
+```
+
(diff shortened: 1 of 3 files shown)


docs/trigger_moderate.md
```diff
@@ -0,0 +1,105 @@
...
+
+First, let's create the extension and set the API key:
+
+```sql
+create extension if not exists ai cascade;
+```
+
+To set the API key, you can use the following command:
+```sql
+select set_config('ai.openai_api_key', :'OPENAI_API_KEY', false) is not null as set_config;
```
jgpruitt, line 15: oops. I set a bad example in the tests.sql. We should use $1 and \bind to help minimize the risk of having the API key printed to logs etc. 

```suggestion
\getenv openai_api_key OPENAI_API_KEY
select set_config('ai.openai_api_key', $1, false) is not null as set_openai_api_key
\bind :openai_api_key
\g
```

jgpruitt (changes requested):
This is super cool!

jgpruitt:
Yeah, the `docs` directory may be better named `examples`.  Let's get the content in and worry about organization later.

docs/trigger_moderate.md
jgpruitt: Fixing tests.sql https://github.com/timescale/pgai/pull/15

jgpruitt (approved):

PR #30, opened 2024-06-14
chore: add maintenance window note.

billy-the-fish:


jonatas (approved):
That's great! I didn't know the maintenance would run during pause! Very smart!

jgpruitt (approved):

PR #31, opened 2024-06-14
Example mixing timescaledb background jobs

jonatas:
I'm expanding the examples introducing timescaledb too.

Probably we should find a better way to handle the api key, but for now, at least we have the templates.


docs/delayed_embed.md
```diff
@@ -0,0 +1,116 @@
...
+via payload.
+
+If you store it in the payload, make sure you're using a secure connection and
+also that your secrets are renewed frequently.
+
+You can also setup this settings at your server level, for example, you can
+add a new config to the system:
+
+```bash
+echo "ai.openai_api_key=${API_KEY}" >> /path/to/postgresql.conf
```
jonatas, line 24: still not sure if this is a valid option.

jgpruitt: I think maybe we can go ahead and merge this if we add a big warning/disclaimer here that putting the API key in the conf file means

1. your API key is in a plain text file
2. any database user can `show ai.openai_api_key` and see it

"use your own judgment" and "proceed at your own risk"

docs/delayed_embed.md
```diff
@@ -0,0 +1,116 @@
...
+  END LOOP;
+END;
+$$ language plpgsql;
+```
+
+Schedule the job to run every 10 seconds:
+
+```
+SELECT add_job('populate_embedding','10s', fixed_schedule => true,
+  config => format('{"api_key": "%s"}', :'OPENAI_API_KEY')::jsonb);
```
jgpruitt, line 97: Need a warning that this puts the API key in the job definition table in plain text

docs/delayed_embed.md
```diff
@@ -0,0 +1,116 @@
...
+  r record;
+  api_key text;
+BEGIN
+  api_key := config->>'api_key';
+  FOR r IN select id, contents from document_embedding where embedding IS NULL LOOP
+    update document_embedding
+    set embedding = openai_embed('text-embedding-ada-002', r.contents, api_key)
+    where id = r.id;
+  END LOOP;
+END;
```
jgpruitt, line 89: I don't know if it complicates the example too much, but consider using the "select for update skip locked" logic that is in [pgvectorizer](https://github.com/timescale/python-vector/blob/main/timescale_vector/pgvectorizer.py#L98) which ought to perform much better.

docs/delayed_embed.md
```diff
@@ -0,0 +1,116 @@
+# Delayed embed
+
+You may decide to combine pgai with [Timescaledb Background Actions][tsdb_actions].
+and you can also use the `pgai` extension to run background actions.
```
jgpruitt, line 4: Should we also provide instructions for using pg_cron?

docs/moderate.md
```diff
@@ -0,0 +1,193 @@
...
+BEGIN
+  select openai_moderate(
+    'text-moderation-stable',
+    NEW.body
+  )->'results'->0 into out;
+  NEW.status = get_moderation_status(out);
+
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
```
jgpruitt, line 80: This trigger only works if the `ai.openai_api_key` session parameter is set. If not, the insert will fail and roll back. Maybe that's a good thing? We should probably note that. It might be worth setting the API key explicitly to make it more obvious, like:

```sql
  select openai_moderate(
    'text-moderation-stable',
    NEW.body,
    _api_key=>current_setting('ai.openai_api_key', false) -- fail if setting not available
  )->'results'->0 into out;
  NEW.status = get_moderation_status(out);
```

docs/moderate.md
```diff
@@ -0,0 +1,193 @@
...
+  comment record;
+  api_key text;
+BEGIN
+  RAISE NOTICE 'Executing action % with config %', job_id, config;
+  -- iterate over comments and moderate them
+  api_key := config->>'api_key';
+  for comment in select * from comments where status = 'pending' loop
+    update comments set status = get_moderation_status(comment.body, api_key)
+    where id = comment.id;
+  end loop;
```
jgpruitt, line 163: same comment about the "select for update skip locked" technique

jgpruitt (changes requested):
Do you think we should put both the markdown and SQL files in the same directory? I kinda think just having an "examples" directory with everything might be better.

docs/delayed_embed.md
```diff
@@ -0,0 +1,116 @@
...
+```bash
+echo "ai.openai_api_key=${API_KEY}" >> /path/to/postgresql.conf
+```
+
+If you're running as admin, you can also use `ALTER SYSTEM` to persist the configuration
+directly from your session:
+
+```sql
+SET ai.openai_api_key TO :'OPENAI_API_KEY' ;
+ALTER SYSTEM SET ai.openai_api_key TO :'OPENAI_API_KEY' ;
```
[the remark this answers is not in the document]
jonatas: Good point! let me change it.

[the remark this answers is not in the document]
jonatas: Not possible to change as `ALTER SYSTEM` seems not support the syntax.

docs/delayed_embed.md
jonatas: Thanks! That's great! Just learning about this options now!

docs/delayed_embed.md
jonatas: That's a good point, I started with timescale jobs but it maybe also a good tutorial we can build and make a lighter dependency.

docs/moderate.md
jonatas: was updating based on the learnings! very good!

docs/delayed_embed.md
jonatas: I'd go with it in another PR to just simplify the merge for now and then we introduce the new example.

docs/delayed_embed.md
```diff
@@ -81,9 +90,10 @@ DECLARE
   api_key text;
 BEGIN
   api_key := config->>'api_key';
-  FOR r IN select id, contents from document_embedding where embedding IS NULL LOOP
+  FOR r IN select id, contents from document_embedding where embedding IS NULL FOR UPDATE SKIP LOCKED LOOP
```
jgpruitt, line 93: I think you'll want to use `LIMIT` and lock a batch of N at a time. Loop until there are none left to do. This limits the number of rows that get locked at the same time.

jgpruitt (changes requested):
do the "for update skip locked" in batches

docs/delayed_embed.md
jonatas: Should we limit 1 by 1?

docs/moderate.md
jonatas: Added.

docs/moderate.md
```diff
@@ -0,0 +1,199 @@
+# Moderate
+
+Let's say you want to moderate comments using OpenAI. You can do it in two ways:
+
+1. Using a trigger that will moderate the comment before it's inserted or updated in the database.
+2. Using a background action that will moderate the comments every [N configurable] seconds.
+
+## Via trigger
+
+You can get the full example in the [trigger_moderate.sql](examples/trigger_moderate.sql) file.
```
jgpruitt, line 10:
```suggestion
```

That file doesn't exist anymore.

jgpruitt (changes requested):
Found 1 broken link.

docs/moderate.md
jonatas: Good point, I added it back just to avoid folks to suffer copying all fragments one by  one.

jonatas: Weird that it's not showing here but I added on f22a417d7caf4c894b37332bb5c706c6c150ba60

jgpruitt (approved):

PR #38, opened 2024-06-24
Support for Anthropic

jgpruitt:


avthars (approved):
Excited to see Claude models in pgai, let's go!

PR #39, opened 2024-06-24
Use PIP_BREAK_SYSTEM_PACKAGES env var instead of pip arg

### LINKED ISSUE #29: Installation failed with no such option: --break-system-packages
Setting it up on EC2 - Amazon Linux 

During installation from source it was failing with below issue.

_no such option: --break-system-packages
make: *** [Makefile:27: install_python_packages] Error 2_

WIth some research found below workaround that worked, i.e. setting up PIP_BREAK_SYSTEM_PACKAGES and remove --break-system-packages from MakeFile.

```
export PIP_BREAK_SYSTEM_PACKAGES=1

# Install the required Python packages
install_python_packages:
	@pip3 install -r requirements.txt

```

Sharing it as issue to retrospect if we can use PIP_BREAK_SYSTEM_PACKAGES as default way in MakeFile.

jgpruitt:
Fixes #29

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,9 +22,11 @@ install_extension:
 	@cp $(SQL_FILES) $(EXTENSION_DIR)
 	@cp $(CONTROL_FILE) $(EXTENSION_DIR)
 
+export PIP_BREAK_SYSTEM_PACKAGES=1
+
 # Install the required Python packages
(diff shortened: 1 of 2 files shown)


avthars (approved):
lgtm

PR #41, opened 2024-06-27
Add support for Cohere

jgpruitt:


docs/cohere.md
```diff
@@ -0,0 +1,445 @@
...
+- [cohere_detokenize](#cohere_detokenize)
+- [cohere_embed](#cohere_embed)
+- [cohere_classify](#cohere_classify)
+- [cohere_classify_simple](#cohere_classify_simple)
+- [cohere_rerank](#cohere_rerank)
+- [cohere_rerank_simple](#cohere_rerank_simple)
+- [cohere_chat_complete](#cohere_chat_complete)
+
+### cohere_list_models
+
```
billy-the-fish, line 146: There are multiple options here, can you update to match the openAI page and be a list of options please. Something along the lines of:

You can list the AI models supported by Cohere in the the following ways:

* List the models supported by the Cohere platform:
   ```
    indented code example
    ```
    Result:
    ```
    indented result
    ```

* Next item

billy-the-fish (changes requested):
Really nice work on the docs, great. Just a couple of small comments to align with the other pages. 

Awesome.

docs/anthropic.md
```diff
@@ -0,0 +1,187 @@
...
+    pip install psycopg2-binary
+    ```
+
+2. Set your Anthropic API key in a .env file or as an environment variable:
+    ```bash
+    ANTHROPIC_API_KEY="this-is-my-super-secret-api-key-dont-tell"
+    DB_URL="your connection string"
+    ```
+
+   1. Pass your API key as a parameter to your queries:
```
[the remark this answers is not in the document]
jgpruitt: done

docs/cohere.md
jgpruitt: done

docs/cohere.md
```diff
@@ -0,0 +1,445 @@
...
+   rerank-english-v3.0           | {rerank}                  | f         |           4096 | https://storage.googleapis.com/cohere-public/tokenizers/rerank-english-v3.0.json           | {}
+   embed-english-light-v3.0      | {embed,classify}          | f         |            512 | https://storage.googleapis.com/cohere-public/tokenizers/embed-english-light-v3.0.json      | {}
+  (18 rows)
+  ```
+
+* List the models on the Cohere platform that support a particular endpoint.
+  
+  ```sql
+  select *
+  from cohere_list_models(_endpoint=>'embed')
```
jonatas, line 185: That would be awesome if all models implement such convenience for filtering. I failed several times using the wrong model because I was testing and it does not implement it 😅

docs/cohere.md
```diff
@@ -0,0 +1,445 @@
...
+
+### cohere_embed
+
+Embed content.
+
+```sql
+select cohere_embed
+( 'embed-english-light-v3.0'
+, 'if a woodchuck could chuck wood, a woodchuck would chuck as much wood as he could'
+, _input_type=>'search_document'
```
jonatas, line 268: Just thinking loud, we may add the `_input_type` with a default just to allow us to have a common method signature between the models.

```
Caused by PG::ExternalRoutineException: ERROR:  cohere.errors.bad_request_error.BadRequestError: status_code: 400, body: {'message': 'invalid request: valid input_type must be provided with the provided model'}
```

jonatas (approved):
I tested locally the embeds ✅ 

Great work John! I think now we can also add more examples as delayed embed for the classifications!

billy-the-fish (approved):
LGTM: good doc.

ai--0.2.0--0.3.0.sql
```diff
@@ -69,3 +68,350 @@ $func$
...
+, default_endpoints text[]
+)
+as $func$
+_api_key_1 = _api_key
+if _api_key_1 is None:
+    r = plpy.execute("select pg_catalog.current_setting('ai.cohere_api_key', true) as api_key")
+    if len(r) >= 0:
+        _api_key_1 = r[0]["api_key"]
+if _api_key_1 is None:
+    plpy.error("missing api key")
```
alejandrodnm, line 95: The `cohere` client also reads the api key from the environment:

```
os.getenv("CO_API_KEY", os.getenv("COHERE_API_KEY"))
```

Should we also support that?, It would be a matter of nor failing here and letting the client fail if the api key is not passed and it cannot load it from the environment.

Also, the code for handling the api key is repeated across all the other functions, could we extracted away? 

https://github.com/cohere-ai/cohere-python/blob/f7c268f08634071dbf5c85d2de749a7c57b625c9/src/cohere/client.py#L430-L435

jgpruitt: >Should we also support that?, It would be a matter of nor failing here and letting the client fail if the api key is not passed and it cannot load it from the environment.

I would like to support that, but the API key would need to be in the environment of the backend postgres process on the server (not the client) since the python is being executed on the server in the backend process. I'm not sure whether arbitrary environment variables that exist when you start the postgres server are passed to background processes. Is that something you'd like to explore?

jgpruitt: >Also, the code for handling the api key is repeated across all the other functions, could we extracted away to a helper function?

Yeah, I've been thinking about this a bit.

- I don't particularly want a helper function exposed as SQL function that the users can call. It's an internal thing
- I'm not sure of the performance implications of a plpython3 function calling another plpython3 function. We should probably explore that.
- I think we could have a Python module with common helper code, but it would need to be installed in the system python environment where we could `import` it. I'm not sure it's worth the trouble, but it's probably worth trying.

I don't really want to hold up a release on this, unless you think we should.

adolsalamanca (approved):

ai--0.2.0--0.3.0.sql
alejandrodnm: No, 🚢

PR #45, opened 2024-07-09
Improve testing

jgpruitt:
- `make test` now works whether you are inside or outside the docker container
- tests now better handle errors that cause the `actual` psql variable to be unset
- fail if no tests were run

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -8,35 +8,12 @@ if [ -f .env ]; then
   set +a
 fi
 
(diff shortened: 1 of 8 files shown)


test.sh
adolsalamanca, line 1: nit, missing of file line

adolsalamanca (approved):
Left a tiny nit

avthars (approved):
LGTM

test.sh
jgpruitt: fixed

PR #44, opened 2024-07-08
Use hoverfly to record tests

alejandrodnm:

PR #47, opened 2024-07-16
Support versioned python dependencies and shared internal python code

jgpruitt:
Adapt build/install to deal with versioning

* "Internal" Python code is now supported
* New versions do not install Python dependencies
  system-wide
* Each version's Python code and dependencies are
  maintained and installed to separate target
  locations
* SQL source code is either idempotent or incremental
* Add migration table to manage incremental SQL
* Revised make to facilitate all of the above

See the changes to DEVELOPMENT.md for further explanation.


make.py
```diff
@@ -0,0 +1,341 @@
...
+
+def build_incremental_sql_file(input_file: Path) -> str:
+    template = sql_dir().joinpath("migration.sql").read_text()
+    migration_name = input_file.name
+    migration_body = input_file.read_text()
+    version = this_version()
+    return template.format(migration_name=migration_name, migration_body=migration_body, version=version)
+
+
+def build_idempotent_sql_file(input_file: Path) -> str:
```
alejandrodnm, line 84: If `this_version` is always the current version, couldn't we extract this code into its own extension function? All the other functions would just call this new other function instead of having to do a search and replace?

make.py
```diff
@@ -0,0 +1,341 @@
...
+    # this is necessary for versions prior to 0.4.0
+    # we will deprecate these versions and then get rid of this function
+    old_reqs_file = src_dir().joinpath("old_requirements.txt")
+    if old_reqs_file.exists():
+        env = {k: v for k, v in os.environ.items()}
+        env["PIP_BREAK_SYSTEM_PACKAGES"] = "1"
+        subprocess.run(f"pip install --compile -r {old_reqs_file}", shell=True, check=True, env=env)
+
+
+def install_prior_py() -> None:
```
alejandrodnm, line 187: What's the use case of this? Is it to downgrade? Wouldn't people just checkout the git tag they want to install and run make install from there?

make.py
```diff
@@ -0,0 +1,341 @@
...
+    for version in prior_versions():
+        if version in {"0.3.0", "0.2.0", "0.1.0"}:  # these are handled by install_old_py_deps()
+            continue
+        if os.sep in version:
+            print(f"'{os.sep}' in version {version}. this is not supported", file=sys.stderr)
+            exit(1)
+        version_target_dir = python_install_dir().joinpath(version)
+        if version_target_dir.exists():
+            continue
+        tmp_dir = Path("/tmp/pgai").joinpath(version)
```
alejandrodnm, line 198: You can use:

```python
import tempfile
with tempfile.TemporaryDirectory() as dir:
    ...
```

sql/ai.control
```diff
@@ -1,6 +1,6 @@
 # pgai extension
 comment='helper functions for ai workflows'
-default_version='0.3.0'
+default_version='0.4.0'
```
alejandrodnm, line 3: If you're worried about having the version in multiple files, we can set up something like release-please to take care of it. We use it in other repos and it works really well.

sql/head.sql
```diff
@@ -0,0 +1,40 @@
+
+set local search_path = pg_catalog, pg_temp;
+
+/*
+if this is the first installation of the extension, create the migration table.
+if the extension has already been installed (i.e. we are upgrading), make sure
+that the user doing the upgrade is the same user that owns the ai_migration
+table. abort the upgrade if different.
```
alejandrodnm, line 8: Is this a common practice? Can't we rely on regular table permissions?

tests/test.sql
alejandrodnm, line 1: I think at some point we'd want a way to run these tests without needing a valid API key. That'll make it easier for people to contribute, and we can run it as part of our CI in the repo.

This is an option with docker-compose and hoverfly https://github.com/timescale/pgai/pull/44/files

alejandrodnm (approved):

make.py
jgpruitt: We *could make it another SQL function, but I was worried about the overhead of making another extra SQL call for EVERY function call. Maybe that's an unnecessary concern. I didn't test the impact of the overhead.

make.py
jgpruitt: That works for NEW installs, but we have to consider upgrade paths.

Each docker (new) image needs to support both the current version AND prior versions so that you can run the image against an old database and it work and then do an upgrade. So, each version's Python dependencies have to be available in the image too.

make.py
jgpruitt: I like it! Lemme change that. That's more portable.

jgpruitt: Any ideas on an appropriate place to put the Python stuff on a Windows machine? `C:\pgai`? I haven't touched Windows in so long...

tests/test.sql
jgpruitt: I agree about wanting to run tests in CI, and you're probably right about not wanting to require a valid API key.

Having to run an HTTP server to run tests gives me pause. Think you could demo that PR for me?

sql/head.sql
jgpruitt: We had this check in Promscale. I know it was security-related. We don't want one database user to be able to take over the extension from another. I don't remember the specifics of the vulnerability. Maybe @cevian does?

sql/ai.control
jgpruitt: yeah, i want to at least make sure the the version is correct in all the places, but having it automatically filled out would be nice

make.py
jgpruitt: fixed.

alejandrodnm: I think something like

C:\Program Files\pgai
C:\Program Files (x86)\pgai

sql/ai.control
jgpruitt: fixed

build.py
```diff
@@ -0,0 +1,368 @@
...
+    template = sql_dir().joinpath("migration.sql").read_text()
+    migration_name = input_file.name
+    migration_body = input_file.read_text()
+    version = this_version()
+    return template.format(migration_name=migration_name, migration_body=migration_body, version=version)
+
+
+def build_idempotent_sql_file(input_file: Path) -> str:
+    inject = f"""
+    if "ai.python_lib_dir" not in GD:
```
cevian, line 99: how does this work when an extension is updated. Do we clear this in the update script? Is it possible to remove the sitedir?

build.py
```diff
@@ -0,0 +1,368 @@
...
+        dest = ext_dir.joinpath(src.name)
+        shutil.copyfile(src, dest)
+
+
+def uninstall_sql() -> None:
+    ext_dir = extension_dir()
+    if not ext_dir.exists():
+        return
+    ext_dir.joinpath("ai.control").unlink(missing_ok=True)
+    for f in ext_dir.glob("ai--*.sql"):
```
cevian, line 179: should this uninstall all ai files or only for this version? I am actually not sure what the right thing to do is.

build.py
```diff
@@ -0,0 +1,368 @@
...
+        tmp_dir.mkdir(parents=True, exist_ok=True)
+        subprocess.run(f"git clone https://github.com/timescale/pgai.git --branch {version} {tmp_dir}", shell=True,
+                       check=True, env=os.environ)
+        tmp_src_dir = tmp_dir.joinpath("src")
+        subprocess.run(f'pip install --compile -t "{version_target_dir}" "{tmp_src_dir}"', check=True,
+                       shell=True, env=os.environ)
+        shutil.rmtree(tmp_dir)
+
+
+def build_pyproject_toml() -> None:
```
cevian, line 222: why is this only needed for the current version and not the previous versions? Shouldn't I need to do it in install_prior_py?

sql/ai.control
```diff
@@ -1,6 +1,7 @@
...
 requires='vector,plpython3u'
 relocatable=false
+schema=ai
```
cevian, line 6: Yeah I think this makes sense. Let's double check with @avthars as well since this changes all the function names.

sql/head.sql
```diff
@@ -0,0 +1,54 @@
+
```
cevian, line 1: Does pgspot pass on this? Or maybe run it on the final generated SQL.

sql/idempotent/01-openai.sql
```diff
@@ -0,0 +1,215 @@
+
```
cevian, line 1: Not reviewing this as I am assuming no change, just moving code around? (everything in dir)

src/ai/anthropic.py
```diff
@@ -0,0 +1,22 @@
+from typing import Optional
```
cevian, line 1: Not reviewing this as I am assuming no change, just moving code around? (everything in dir)

cevian (changes requested):
Looks good, but have some questions

build.py
jgpruitt: When we release version 0.4.0, we will check in all the `ai--*0.4.0.sql` scripts which will have the path to the 0.4.0 python directory hardcoded in them (courtesy of this function).
Then, we put 0.5.0 at the front of the versions list, and the next time we build we get `ai--*0.5.0.sql` scripts with the path hardcoded to the new 0.5.0 python directory.
When you do an `alter extension update` all the 0.4.0 functions will get replaced with 0.5.0 functions with the new path. You'll need to start a new database session after the upgrade.

jgpruitt: maybe we could improve it to remove the need for starting a new session though

jgpruitt: i don't see a "remove site dir" function at quick glance

build.py
jgpruitt: I don't know what the "right" thing is. Both options would be valuable. I intended it to be a "fully remove the pgai extension from my system" thing. For uninstall-py, I delete the directory with all the versions' python libs.

Maybe I should make `clean` also uninstall THIS version?

build.py
jgpruitt: `pyproject.toml` is checked in to version control. So, all the previous versions will have the file with the correct version already in it. This function just ensures that you can't screw up the current version. The only place you have to update the version when starting a new release is in the `versions()` function.

src/ai/anthropic.py
jgpruitt: yep!

sql/idempotent/01-openai.sql
jgpruitt: I did move some code from the bodies of the SQL functions to the Python module

sql/head.sql
jgpruitt: good call. let me try that.

sql/ai.control
jgpruitt: I think this needs a migration to move existing function from prior versions to the new schema.

sql/head.sql
jgpruitt: head.sql passes.

pgspot doesn't know what to do with plpython3u.
`Unknown function language: plpython3u`

build.py
jgpruitt: https://github.com/timescale/pgai/pull/47/commits/c0eef3aa452e054ed0d84160daf938fe0954419d

Will now bail if the version has changed.

sql/head.sql
jgpruitt: migration.sql [now passes](https://github.com/timescale/pgai/pull/47/commits/1361f5e88def739050d7b5df461774584cfe09df)

sql/head.sql
jgpruitt: Okay, I remember the vulnerability. If a user "bob" creates the schema "ai" and the migration table "ai.migration" and THEN someone does `create extension ai cascade;` then "bob" can do nasty, hateful things.

sql/ai.control
jgpruitt: okay. we're good here.

cevian (approved):
LGTM. Some minor nits.

Also we MUST remember to tell people to drop and recreate the extension in the release notes.

build.py
```diff
@@ -0,0 +1,382 @@
...
+        python_lib_dir = Path(python_lib_dir).joinpath("{this_version()}")
+        import site
+        site.addsitedir(str(python_lib_dir))
+        from ai import __version__ as ai_version
+        assert("{this_version()}" == ai_version)
+        GD["ai.python_lib_dir"] = python_lib_dir
+    else:
+        python_lib_dir = GD["ai.python_lib_dir"]
+        from pathlib import Path
+        if Path(python_lib_dir).name != "{this_version()}":
```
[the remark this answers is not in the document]
jgpruitt: AH! Good point. Duh. I will change.

tests/test.sql
jgpruitt: will do in another commit

PR #53, opened 2024-07-25
Add linting tools

jgpruitt:
Adds 4 new make targets:

1. lint-sql - runs pgspot linter against the output SQL for the current version
2. lint-py - runs ruff linter against the python source files
3. lint - runs both lint-sql and lint-py
4. format-py - runs ruff format check against the python source files

Fixes issues found by linters and format check


avthars (approved):
LGTM

PR #54, opened 2024-07-26
Fixed silly logic bug

jgpruitt:


jonatas (approved):

src/ai/anthropic.py
```diff
@@ -6,10 +6,9 @@ def find_api_key(plpy) -> str:
     r = plpy.execute(
         "select pg_catalog.current_setting('ai.anthropic_api_key', true) as api_key"
     )
-    if len(r) >= 0:
-        return r[0]["api_key"]
-    else:
+    if len(r) == 0:
```
alejandrodnm, line 9: NIT: the pythonic way to check if a string is empty is with `if not r:`

jgpruitt: It's not a string, but I get the point. Such a weird convention. I'll change it if I get a chance today.

alejandrodnm: don't change it, it works.

alejandrodnm:
```
🔸 Falsy Values
Sequences and Collections:

Empty lists []
Empty tuples ()
Empty dictionaries {}
Empty sets set()
Empty strings ""
Empty ranges range(0)
Numbers

Zero of any numeric type.
Integer: 0
Float: 0.0
Complex: 0j
Constants

None
False
```

PR #56, opened 2024-07-26
OpenAI base_url parameter and Ollama OpenAI support

### LINKED ISSUE #33: Add support for openrouter
I noticed that the current pgai is only support chatgpt. But there are gemini, claude and llama and even more large language model existed. which provide same level service to GPT4 and cost efficient. 

Is is possible to provide openrouter service as well? 
Here is the document link, https://openrouter.ai/docs/quick-start
From the quick-start note, we can find it can use openai python library and compatible with openai chat.completion. The main changes will be 
== connection
1. support base_url parameter , by set it to openrouter address
2. Add another api_keys
By support base_url and additonal api_keys, the pgai will support connection to openrouter even ollama model
==model name
use the model name from openrouter return api list
I think let's all.

I will try to do it in my local computer first.

jgpruitt:
@avthars not sure whether we want to support this or not.

It would allow someone to use the `openai` functions but point it to [openrouter](https://openrouter.ai/docs/quick-start) or [ollama](https://github.com/ollama/ollama/blob/main/docs/openai.md) instead of openai.

There are actually a number of parameters to creating an openai client that we don't currently support that we may want to consider:

```python
    def __init__(
        self,
        *,
        api_key: str | None = None,
        organization: str | None = None,
        project: str | None = None,
        base_url: str | httpx.URL | None = None,
        timeout: Union[float, Timeout, None, NotGiven] = NOT_GIVEN,
        max_retries: int = DEFAULT_MAX_RETRIES,
        default_headers: Mapping[str, str] | None = None,
        default_query: Mapping[str, object] | None = None,
        # Configure a custom httpx client.
        # We provide a `DefaultHttpxClient` class that you can pass to retain the default values we use for `limits`, `timeout` & `follow_redirects`.
        # See the [httpx documentation](https://www.python-httpx.org/api/#client) for more details.
        http_client: httpx.Client | None = None,
        # Enable or disable schema validation for data returned by the API.
        # When enabled an error APIResponseValidationError is raised
        # if the API responds with invalid data for the expected schema.
        #
        # This parameter may be removed or changed in the future.
        # If you rely on this feature, please open a GitHub issue
        # outlining your use-case to help us decide if it should be
        # part of our public interface in the future.
        _strict_response_validation: bool = False,
    ) -> None:
```

Fixes #33

### DIFF
--- a/sql/idempotent/01-openai.sql
+++ b/sql/idempotent/01-openai.sql
@@ -35,7 +35,7 @@ set search_path to pg_catalog, pg_temp
 -- openai_list_models
 -- list models supported on the openai platform
 -- https://platform.openai.com/docs/api-reference/models/list
-create or replace function ai.openai_list_models(_api_key text default null)
+create or replace function ai.openai_list_models(_api_key text default null, _base_url text default null)
 returns table
 ( id text
 , created timestamptz
@@ -44,7 +44,7 @@ returns table
 as $python$
     #ADD-PYTHON-LIB-DIR
     import ai.openai
-    for tup in ai.openai.list_models(plpy, _api_key):
+    for tup in ai.openai.list_models(plpy, _api_key, _base_url):
         yield tup
 $python$
 language plpython3u volatile parallel safe security invoker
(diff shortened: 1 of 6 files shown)


adolsalamanca (approved):

avthars (approved):
Let's do it

PR #60, opened 2024-08-05
Modify func volatility for performance

jgpruitt:
Database functions were all volatile. This is not as performant as it could be. This PR changes the volatility of many functions to address this.

- The text generation/chat complete functions were left volatile
- ollama_ps was left volatile since it describes the running state of ollama
- the "list models" were made stable since models can be added/removed at any time
- embed, classify, moderate, etc. were made immutable

### DIFF
--- a/sql/idempotent/01-openai.sql
+++ b/sql/idempotent/01-openai.sql
@@ -11,7 +11,7 @@ as $python$
     tokens = encoding.encode(_text)
     return tokens
 $python$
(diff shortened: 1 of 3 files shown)


cevian:
I would leave "list models" as volatile because the changes to models don't depend on DB state.

jgpruitt:
@cevian fixed!

cevian (approved):

PR #61, opened 2024-08-07
Convert tests to pytest

jgpruitt:
Removes the old psql-based tests and implements the same tests using pytest. This sets us up well for testing both SQL as we have been and the internal Python code directly.


tests/test_ollama.py
```diff
@@ -0,0 +1,249 @@
...
+        , _options=> jsonb_build_object
+          ( 'seed', 42
+          , 'temperature', 0.6
+          )
+        )
+    """,
+        (ollama_host,),
+    )
+    actual = cur.fetchone()[0]
+    assert "response" in actual and "done" in actual and actual["done"] is True
```
alejandrodnm, line 124: Small thing. If one of these conditions fails, would it be easy to know which one? I really don't know, but I suspect you'd get something like assertion fail . Maybe it's better to have one assert statement per check instead of using `and`.

Don't change them, but maybe for future tests it's better to split them.

alejandrodnm: Well I tested this. You get something like

```
======================================================= FAILURES ======================
________________________________________________________ test_a _______________________

    def test_a():
        a = 42
        b = "asd"
>       assert a == 42 and "afs" in b
E       AssertionError: assert (42 == 42 and 'afs' in 'asd')

tests/test_lambda.py:15: AssertionError
=============================================== short test summary info ===============
FAILED tests/test_lambda.py::test_a - AssertionError: assert (42 == 42 and 'afs' in 'as
================================================== 1 failed in 0.06s ==================
```

alejandrodnm (approved):

tests/test_ollama.py
```diff
@@ -0,0 +1,249 @@
...
+@pytest.fixture()
+def con(db_url) -> psycopg.Connection:
+    return psycopg.connect(db_url)
+
+
+@pytest.fixture()
+def cur(con) -> psycopg.Cursor:
+    with con:
+        with con.cursor() as cursor:
+            yield cursor
```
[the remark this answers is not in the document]
jgpruitt: fixed

PR #69, opened 2024-09-12
pass secrets on event payload

alejandrodnm:


src/ai/vectorizer.py
```diff
@@ -31,7 +31,14 @@ def execute_vectorizer(plpy, vectorizer_id: int) -> None:
     result = plan.execute([vectorizer_id], 1)
     if not result:
         plpy.error(f"vectorizer {vectorizer_id} not found")
-    vectorizer = json.loads(result[0]["vectorizer"])
+
+    vectorizer = result[0]["vectorizer"]
+    try:
+        embedding_api_key = vectorizer["config"]["embedding"]["api_key_name"]
+        vectorizer["secrets"] = [embedding_api_key]
```
jgpruitt, line 38: Why is this necessary? I don't understand.

alejandrodnm: We need a way to signal to the savannah-function that it needs to fetch that secret.

PR #70, opened 2024-09-13
Remove pgcron scheduling implementation

jgpruitt:
We decided that support for pgcron is not necessary. Backing out the implementation.


build.py
```diff
@@ -486,17 +486,16 @@ def docker_build() -> None:
 
 
 def docker_run() -> None:
-    if does_container_exist():
-        print("pgai container already exists", file=sys.stderr)
-        sys.exit(1)
+    #if does_container_exist():
```
cevian, line 489: why is this commented out?

cevian (approved):
One thing to check, otherwise lgtm

build.py
jgpruitt: should have gone in this one https://github.com/timescale/pgai/pull/71
I'll back it out

PR #71, opened 2024-09-13
Fix issue with docker ps hanging

jgpruitt:
I tried to make certain make targets "smarter" (e.g. don't try to remove a docker container if it doesn't exist), but the calls to `docker ps` were hanging in some situations, and it became more annoying than valuable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ docker-rm:
 
 .PHONY: run
 run:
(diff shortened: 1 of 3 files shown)


cevian (approved):
I wonder why docker ps hanged

jgpruitt:
seems to happen more often when there is no pgai container

PR #75, opened 2024-09-16
Add vectorizer processing config

jgpruitt:
New config type to allow for runtime processing configuration.


sql/idempotent/010-processing.sql
```diff
@@ -0,0 +1,85 @@
...
+    )
+$func$ language sql immutable security invoker
+set search_path to pg_catalog, pg_temp
+;
+
+-------------------------------------------------------------------------------
+-- processing_cloud_functions
+create or replace function ai.processing_cloud_functions
+( batch_size int default 50
+, concurrency int default 1
```
cevian, line 18: I would actually default these to NULL. The reason? NULL would mean "default". So if we ever wanted to change the default we could just change in in the lambda. Otherwise no way to know if batch_size of 50 was explicitly set or the default value and thus lambda wouldn't know if it's "free" to change it.

sql/idempotent/010-processing.sql
```diff
@@ -0,0 +1,85 @@
...
+            end if;
+            if cast(_val as int) < 1 then
+                raise exception 'batch_size must be greater than 0';
+            end if;
+
+            _val = pg_catalog.jsonb_extract_path(config, 'concurrency');
+            if pg_catalog.jsonb_typeof(_val) operator(pg_catalog.!=) 'number' then
+                raise exception 'concurrency must be a number';
+            end if;
+            if cast(_val as int) > 10 then
```
cevian, line 69: I would set the limit here to 50?

cevian (changes requested):

sql/idempotent/010-processing.sql
jgpruitt: good idea. i like that

sql/idempotent/010-processing.sql
jgpruitt: okay

alejandrodnm (approved):

PR #82, opened 2024-09-24
Add openai_chat_complete_simple (and tests)

akulkarni:
Current openai_chat_complete function requires choosing the model, submitting the message in a json object, and parsing the json object output. This can feel cumbersome to do.

This openai_chat_complete_simple chooses a default model, only requires the message to be passed as text, and returns the response as text. This should be easier to use, albeit maybe not as powerful.

Future work:

- Consider making the default model an env variable

- Repeat for the other LLMs (Anthropic, Cohere, Ollama)

### DIFF
--- a/sql/ai--0.4.0.sql
+++ b/sql/ai--0.4.0.sql
@@ -400,6 +400,26 @@ language plpython3u volatile parallel safe security invoker
 set search_path to pg_catalog, pg_temp
 ;
 
+------------------------------------------------------------------------------------
+-- openai_chat_complete_simple
+-- simple chat completion that only requires a message and only returns the response
+create or replace function ai.openai_chat_complete_simple
+( _message text
+, _api_key text default null
+) returns text
+as $$
+declare
+    model text := 'gpt-4o';
+    messages jsonb;
+begin
+    messages := jsonb_build_array(
+        jsonb_build_object('role', 'system', 'content', 'you are a helpful assistant'),
+        jsonb_build_object('role', 'user', 'content', _message)
+    );
+    return ai.openai_chat_complete(model, messages, _api_key)->'choices'->0->'message'->>'content';
+end;
+$$ language plpgsql;
+
 -------------------------------------------------------------------------------
 -- openai_moderate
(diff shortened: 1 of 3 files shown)


akulkarni:
I have been using pgai a lot lately. Adding something that I would love to have as a user. 

@jgpruitt Don't go easy on me! Need to keep code quality high :-)

akulkarni:
(Force pushed some code cleanup I missed on the initial commit. No real changes.)

sql/idempotent/001-openai.sql
```diff
@@ -198,6 +198,25 @@ language plpython3u volatile parallel safe security invoker
 set search_path to pg_catalog, pg_temp
 ;
 
+------------------------------------------------------------------------------------
+-- simple chat completion that only requires a message and only returns the response
```
jgpruitt, line 202:
```suggestion
------------------------------------------------------------------------------------
-- ai.openai_chat_complete_simple
-- simple chat completion that only requires a message and only returns the response
```

sql/ai--0.4.0.sql
```diff
@@ -400,6 +400,25 @@ language plpython3u volatile parallel safe security invoker
 set search_path to pg_catalog, pg_temp
 ;
 
+------------------------------------------------------------------------------------
+-- simple chat completion that only requires a message and only returns the response
```
jgpruitt, line 404:
```suggestion
------------------------------------------------------------------------------------
-- ai.openai_chat_complete_simple
-- simple chat completion that only requires a message and only returns the response
```

jgpruitt (approved):
This is great! One little asthetic suggestion. I like to put the name of the function in a comment header.

akulkarni:
@jgpruitt Thanks! I made the change, but following the format in the other function comments (i.e., no "ai." header). I amended the previous commit with the changes, and then force pushed the amended commit. Please let me know if you'd rather I do it another way. Thanks!

jgpruitt (approved):
LGTM! Thanks for contributing!

PR #91, opened 2024-10-01
Add vectorizer worker

jgpruitt:
Creating a foundation for implementing the vectorizer worker. The vectorizer code doesn't "DO" anything yet. This PR reorganizes the codebase to "make room" for it and adds make targets to install, uninstall, and clean the vectorizer. Other make targets were adjusted as necessary.


cevian (approved):
Looks good. Are we sure we want root-user-action=ignore tho?

PR #95, opened 2024-10-04
feat: add config option to only build vec idx when q empty

jgpruitt:
adds a `create_when_queue_empty` flag to the diskann and hnsw index creation configs. If set and the queue table is NOT empty, we skip vector index creation.


sql/idempotent/011-vectorizer-int.sql
```diff
@@ -613,11 +615,82 @@ begin
...
     from ai.vectorizer v
     where v.id operator(pg_catalog.=) _vectorizer_id
     ;
+
+    commit;
+    set local search_path = pg_catalog, pg_temp;
+
+    -- create a vector index if the conditions are right. otherwise proceed to vectorizer execution
+    -- hacky. a loop that can only iterate once. we can exit the loop early if it's not time to create the index
+    -- this logic *could be moved to _vectorizer_create_vector_index, but it would be nice to have function that
```
cevian, line 630: suggestion (but not a must) I would move this code into a _vectorizer_should_create_index(vectorizer_row) returns boolean function. That avoids some of the loop weirdness and is a bit cleaner.

cevian (approved):

sql/idempotent/011-vectorizer-int.sql
jgpruitt: yeah. i like that

jgpruitt: done.

PR #109, opened 2024-10-10
feat: add destination param to create_vectorizer

jgpruitt:


projects/extension/sql/idempotent/012-vectorizer-api.sql
```diff
@@ -22,6 +22,7 @@ create or replace function ai.create_vectorizer
 , formatting jsonb default ai.formatting_python_template()
 , scheduling jsonb default ai.scheduling_timescaledb()
 , processing jsonb default ai.processing_default()
+, destination name default null
```
cevian, line 25: should this be the second param

cevian (approved):

alejandrodnm:
Should we update the collision error message to instead instruct users to pass `destination` instead of `target_table` and `view_name`?

alejandrodnm (approved):

PR #110, opened 2024-10-11
chore: prepare pypi build

alejandrodnm:


thedodd (approved):
LGTM

projects/pgai/pgai/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "0.4.0"
+__version__ = "0.1.0rc1"
```
jgpruitt, line 1: If we are going to change this (which we should), then please remove the `build_pgai_init_py` function from `build.py` too

Makefile
```diff
@@ -154,3 +154,8 @@ docker-shell:
 .PHONY: psql-shell
 psql-shell:
 	@docker exec -it -u postgres pgai /bin/bash -c "set -e; if [ -f .env ]; then set -a; source .env; set +a; fi; psql"
+
+.PHONY: build-pgai
+build-pgai:
+	@cd projects/pgai && python -m build --sdist --wheel
+	@twine check projects/pgai/dist/*
```
jgpruitt, line 161: why is this here and not in `build.py`? should we replace or remove `install_vectorizer()` in `build.py`?

projects/pgai/pyproject.toml
```diff
@@ -4,8 +4,33 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "pgai"
+description = "AI workflows in your PostgreSQL database"
 dynamic = ["version", "dependencies"]
 requires-python = ">=3.10"
+readme = "README.md"
+keywords = ["ai", "postgres"]
```
jgpruitt, line 11:
```suggestion
keywords = ["ai", "postgres", "postgresql", "llm", "gpt"]
```

jgpruitt (changes requested):

Makefile
alejandrodnm: This just builds the package and runs a check on the artifact. 

I find `build.py` a little confusing if we already have a `Makefile`. I understand that there are some complicated tasks like the dependencies management of the extension. There are some targets that are simple and could just live in the Makefile.

In this particular case, `install_vectorizer` installs the package directly from the source code into the current environment. While this produces the distributable and just checks it for issues.

jgpruitt: Initially, I tried to do everything solely in a Makefile, but many tasks were way too complex to deal with in make and stay sane. Then, I tried implementing those in a shell script called by make, but again, some were too complex, and I was running into issues with portability. Then, I realized that anyone working on pgai would have Python installed, and anywhere it would be installed would have Python installed. Implementing these tasks in Python is mostly trivial and portable with just the standard library. The main reason I kept the Makefile around was that people expect there to be one and look for it by default.

jgpruitt: Should `install_vectorizer` use the distributable instead? I want to make sure the developer experience is as easy and understandable as possible so it's not a blocker for contributors.

alejandrodnm: The end result of both approaches is the same. I have no strong opinion. I can change `install_vectorizer` to create the wheel first and then install it.

alejandrodnm: I changed `build-pgai` to `build-vec-dist` and moved the logic to `build.py`.

jgpruitt (approved):

PR #113, opened 2024-10-11
chore: add docker-compose

alejandrodnm:
Adds a Docker compose example.

Installs the pgai extension when building the Docker container.

Enables auto prefix for the env vars of the vectorizer cli options.


thedodd (approved):

jgpruitt (commented):
Why are we adding docker compose? If this is just an example, maybe it should be put in the `examples` directory?

Dockerfile
```diff
@@ -64,6 +64,8 @@ COPY requirements-test.txt /build/requirements-test.txt
 RUN pip install -r /build/requirements-test.txt
 COPY projects/pgai/requirements.txt /build/requirements-pgai.txt
 RUN pip install -r /build/requirements-pgai.txt
-RUN rm -r /build
 
 WORKDIR /pgai
+
+COPY . .
+RUN make install
```
jgpruitt, line 71: this may cause issues with the existing developer experience. Also, if you're going to do this, you have to do `RUN make build-install`.

jgpruitt (changes requested):

Dockerfile
jgpruitt: If we want to go with docker compose, that's fine, but we need all the tooling to work with it. We should pick ONE way and support it fully. I don't want multiple ways that work slightly differently and incompletely. For example, `make psql-shell` does not work with this docker compose setup.

PR #114, opened 2024-10-14
feat: add polling mode to vectorizer cli

alejandrodnm:
- Introduced `--poll-interval` option to specify the wait time before checking for new work after processing all available tasks.
- Added `--once` flag to exit the program after processing all available work in the queue.

Closes [AI-58](https://linear.app/timescale/issue/AI-58/modify-vectorizer-container-to-keep-running-and-log-status)

### DIFF
--- a/projects/pgai/pgai/cli.py
+++ b/projects/pgai/pgai/cli.py
@@ -1,16 +1,18 @@
 import asyncio
+import datetime
 import logging
 import os
 import random
 import sys
+import time
 from collections.abc import Sequence
-from pathlib import Path
 
 import click
 import psycopg
 import structlog
 from dotenv import load_dotenv
 from psycopg.rows import dict_row, namedtuple_row
+from pytimeparse import parse  # type: ignore
 
 from .__init__ import __version__
 from .vectorizer.secrets import Secrets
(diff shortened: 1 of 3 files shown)


projects/pgai/pgai/cli.py
```diff
@@ -119,28 +151,31 @@ async def run_workers(
...
+    show_default=True,
+    help="The interval, in duration string or integer (seconds), to wait before checking for new work after processing all available work in the queue.",  # noqa
+)
+@click.option(
+    "--once",
+    type=click.BOOL,
+    default=False,
+    show_default=True,
+    help="Exit after processing all available work.",
+)
```
jgpruitt, line 167: If I want `--once`, it looks like I still have to provide a `--poll-interval` value.
Might be nice to make `--poll-interval` optional, OR make `--poll-interval=0` mean `--once`

jgpruitt (changes requested):

projects/pgai/pgai/cli.py
alejandrodnm: `--poll-interval` has a default , so you don't need to pass it always.

I thought about `--poll-interval=0` mean `--once`. My only issue is that it's not as explicit and maybe people don't realize that. But I'm willing to change it, I'm not 100% sold on either option really.

thedodd (approved):

projects/pgai/pgai/cli.py
jgpruitt: okay, I'm fine with it as-is

jgpruitt (approved):

PR #115, opened 2024-10-14
chore: remove datadog tracer

alejandrodnm:


smoya (approved):

cevian (changes requested):
I am skeptical we want to do this. lets discuss before merging.

PR #120, opened 2024-10-15
chore: add docs to classes and functions

alejandrodnm:


projects/pgai/pgai/vectorizer/chunking.py
```diff
@@ -11,12 +11,44 @@
...
+
+    Defines the interface for any chunker that breaks a dictionary item into
+    chunks of text.
+    """
+
     @abstractmethod
     def into_chunks(self, item: dict[str, Any]) -> list[str]:
-        pass
+        """
+        Breaks the provided dictionary item into chunks of text.
```
jgpruitt, line 24: nit: this is a little vague and confusing. It might help to explain that the dict represents a database row, and that the Chunker will chunk one specified item within the dict.

jgpruitt (approved):

projects/pgai/pgai/vectorizer/chunking.py
alejandrodnm: Makes sense, I'll update it

PR #123, opened 2024-10-16
chore: doc review

billy-the-fish:
This is an awesome doc, a few tw suggestions for you to emphasize the workflow in the doc rather than the feature name. 

What I suggest is that we change the name of this guide to vectorizer.md, and the current vectorizer.md to vectorizer-api-reference.md.

### DIFF
--- a/docs/vectorizer_guide.md
+++ b/docs/vectorizer_guide.md
@@ -1,4 +1,4 @@
-# Vectorizer Guide
+# Automate AI embedding with Vectorizer
 
 Vector embeddings have emerged as a powerful tool for transforming text into
 compact, semantically rich representations. This approach unlocks the potential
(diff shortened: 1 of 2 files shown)


docs/vectorizer_guide.md
```diff
@@ -62,11 +68,11 @@ SELECT ai.create_vectorizer(
 );
 ```
 
-In this example, the `contents` field is split into multiple chunks if it's
+In this example, if the `contents` field is lengthy, it is split into multiple chunks
```
JamesGuthrie, line 71: Lengthy is duplicated (it's on the next line, but I can't highlight that line in github's review).

docs/vectorizer_guide.md
```diff
@@ -167,12 +171,12 @@ SELECT ai.create_vectorizer(
...
-The default format string is simply '$chunk'.
+The default format string is simply `$chunk`.
 
-### Indexing
+## Improve query performance on your Vectorizer
 
-The vectorizer can create a vector index on the embedding column to improve
-query performance. By default, a vectorscale index is created after 100,000
+To improve query performance, you create a vector index on the embedding column 
+in a Vectorizer. By default, a vectorscale index is created after 100,000
```
JamesGuthrie, line 179: I'm not convinced about this change, IMO it doesn't clarify anything.

There are three concepts at play here:
1. Creating an index on vector data is good (and sometimes necessary).
2. The Vectorizer will automatically create an index when 100k rows of vector data are present.
3. This behaviour is configurable.

This edit feels like it mixes the concepts.

JamesGuthrie (changes requested):

docs/vectorizer_guide.md
```diff
@@ -19,25 +19,31 @@ representing only a part of a row's data -- we've simplified the entire workflow
...
+- Keep embeddings continuously synchronized with source data (asynchronously).
+- Utilize a convenient view that seamlessly joins base tables with their embeddings.
 
-This guide offers a comprehensive overview of the Vectorizer feature,
+This page offers a comprehensive overview of Vectorizer features,
 demonstrating how it streamlines the process of working with vector embeddings
-in your database. A more detailed technical specification is available
-[here](./vectorizer.md).
+in your database. For a more detailed technical specification, see the
+[Vectorizer API reference](./vectorizer.md).
```
smoya, line 30: Actually, just by reading `Vectorizer API reference`, I expected to find somewhere in `vectorizer.md` a compact list of all possible extension functions a user can call. 
Instead, all of them are separated into paragraphs exposing user actions, which I find it right for some use cases. However, for the use case of "I don't know what I don't know" I find it useful to have a visual of all the available functions, perhaps as a link pointing to the right paragraph where they usage are documented.

WDYT? Does it make sense? Discard if you don't see the value.

billy-the-fish: In https://github.com/timescale/pgai/blob/vectorizer/docs/vectorizer.md, i have a list showing where to find what you need. However, it is not sorted by API name, still by action group. Are you ok with that or would you prefer by api name?

docs/vectorizer_guide.md
billy-the-fish: :-), thanks. Fixed.

docs/vectorizer_guide.md
billy-the-fish: Updated the description to match what your comment.

cevian (approved):
Thanks

JamesGuthrie (approved):

PR #128, opened 2024-10-16
chore: add vectorizer load test scripts

jgpruitt:


cevian (approved):
👍

.github/workflows/ci.yml
```diff
@@ -51,13 +48,9 @@ jobs:
...
-        run: docker exec pgai make test
-
-      - name: Run vectorizer tests
-        run: make test-vectorizer
-
-      - name: Run type check
-        run: make type-check-py
+        run: |
+          docker exec pgai make test
+          docker exec pgai make test-vec
```
smoya, line 53: Wondering if we can parallelize those two jobs rather than running them sequentially

jgpruitt: I'm backing these changes out. I think we need a more wholistic reworking of the build

PR #132, opened 2024-10-17
fix: support text-embedding-ada-002

jgpruitt:


cevian (changes requested):
Wait how do we know the dimensions of the vector column in the storage table then?

jgpruitt:
Crap. How did this not trip a test?

jgpruitt:
Okay I guess we have to handle this in the vectorizer code

cevian (approved):

projects/extension/sql/ai--0.4.0.sql
```diff
@@ -2652,14 +2652,6 @@ begin
             raise warning 'one or more grant_to roles do not exist: %', _missing_roles;
```
cevian, line 2652: I think you need to regenerate the sql before merging

PR #137, opened 2024-10-18
fix: do not send dimensions for ada-002 model

jgpruitt:


projects/pgai/pgai/vectorizer/embeddings.py
```diff
@@ -201,6 +201,8 @@ class OpenAI(ApiKeyMixin, BaseModel, Embedder):
 
     @cached_property
     def _openai_dimensions(self) -> int | openai.NotGiven:
+        if self.model == "text-embedding-ada-002":
```
alejandrodnm, line 204: Are we guarding against this in the extension as well?

jgpruitt: No. We have to have the dimensions specified at the SQL level so we know how to create the target table

cevian (approved):

PR #150, opened 2024-10-21
feat: use commands for scoping the cli

alejandrodnm:
Instead of the cli running directly the vectorizer worker, the worker has been moved to its own command. The entry point has been rename from vectorizer to pgai.

This will allow us to add other commands seamlessly in the future, without having to introduce breaking changes or multiple scripts.

The workflow will look like this:

```
$ pip install pgai
$ pgai --help
Usage: pgai [OPTIONS] COMMAND [ARGS]...

Options:
  --version  Show the version and exit.
  --help  Show this message and exit.

Commands:
  vectorizer

$ python -m pgai vectorizer --help
Usage: python -m pgai vectorizer [OPTIONS] COMMAND [ARGS]...

Options:
  --version  Show the version and exit.
  --help     Show this message and exit.

Commands:
  worker

$ pgai vectorizer worker --help
Usage: pgai vectorizer worker [OPTIONS]

Options:
  --version                       Show the version and exit.
  -d, --db-url TEXT               The database URL to connect to  [default:
                                  postgres://postgres@localhost:5432/postgres]
  -i, --vectorizer-id INTEGER     Only fetch work from the given vectorizer
                                  ids. If not provided, all vectorizers will
                                  be fetched.
  -c, --concurrency INTEGER RANGE
                                  [default: 1; x>=1]
  --log-level [DEBUG|INFO|WARN|ERROR|FATAL|CRITICAL]
  --poll-interval TIME DURATION   The interval, in duration string or integer
                                  (seconds), to wait before checking for new
                                  work after processing all available work in
                                  the queue.  [default: 5m]
  --once BOOLEAN                  Exit after processing all available work.
                                  [default: False]
  --help                          Show this message and exit.
```

### DIFF
--- a/build.py
+++ b/build.py
@@ -543,10 +543,10 @@ def test_server() -> None:
 
 def vectorizer() -> None:
     if where_am_i() == "host":
-        cmd = "docker exec -it pgai vectorizer --version"
+        cmd = "docker exec -it pgai pgai vectorizer-worker --version"
         subprocess.run(cmd, shell=True, check=True, env=os.environ, cwd=root_dir())
     else:
-        cmd = "vectorizer --version"
+        cmd = "pgai vectorizer-worker --version"
         subprocess.run(
             cmd,
             shell=True,
--- a/projects/pgai/Dockerfile
+++ b/projects/pgai/Dockerfile
@@ -11,5 +11,5 @@ RUN pip3 install --no-cache-dir --disable-pip-version-check --compile --root-use
 USER pgaiuser
 COPY pyproject.toml /app
 COPY pgai /app/pgai
-ENTRYPOINT [ "python", "-m", "pgai" ]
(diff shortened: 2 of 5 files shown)


smoya:
Nit: From a UX point of view, I find `pgai vectorizer-worker` command sequence not pretty descriptive nor natural. Additionally, from the technical point of view, if you want to add more vectorizer subcommand , you are tied to use `vectorizer-<subcommand>` name pattern.

Can we add a proper subcommand level such as `pgai vectorizer run` or `pgai vectorizer worker`? Still `pgai vectorizer` could run a default behaviour. That let's user understand there can be more subcommand for `vectorizer`.

alejandrodnm:
@smoya I went with `pgai vectorizer worker`.

billy-the-fish:
@alejandrodnm can you ping me when this is live and I will update the docs to match. I'd like to do it in a separate PR as there are a couple of other doc moving parts at the moment.

PR #151, opened 2024-10-21
chore: improve docs for self hosting

Askir:
I ran into a bunch of little issues trying out the self hosted vectorizer. These changes should make it a little easier for the next person to go through this.


docs/vectorizer.md
```diff
@@ -140,13 +152,15 @@ The view includes all columns from the blog table, plus the following additional
 To find the closest embeddings to a query, use this canonical SQL query:
 
 ```sql
-SELECT 
-   chunk,
-   embedding <=> <query embedding> as distance
+SELECT chunk,
+   embedding <=> ai.openai_embed('text-embedding-3-small', 'my query string', _dimensions=>768)  as distance
```
Askir, line 156: The `ai.openai_embed` function allows users to directly do semantic search in SQL, should we replace all occurences of `<query embedding>` with it?

Askir, line 156: Is the `_dimensions=>768` parameter supposed to be private or why does it lead with underscore? without it, the dimensions are too big and I got an error.

billy-the-fish:
Pushed a couple of doc things into your branch.

docs/vectorizer.md
```diff
@@ -140,13 +151,15 @@ The view includes all columns from the blog table, plus the following additional
...
 FROM blog_contents_embeddings
-ORDER BY distance 
-LIMIT 10;
+ORDER BY distance
+   LIMIT 10;
```
cevian, line 158: nit:
```suggestion
LIMIT 10;
```

docs/vectorizer.md
cevian: Good point we may need to remove the underscore @jgpruitt

docs/vectorizer.md
```diff
@@ -69,6 +69,13 @@ Timescale Cloud or on a self-hosted Postgres server.
   vectorizer
   ```
 
+To validate that your API key is set correctly, run the following SQL query:
```
cevian, line 72: This is true on self-hosted but not on cloud yet. Unfortunately. Eventually that should be true. Don't know  if we want to move this to self-hosted section by now?

docs/vectorizer.md
cevian: @jgpruitt what do we need to do to make this work on cloud @jgpruitt I am wondering about secrets.

cevian (changes requested):
Mostly looks good. The only change I'd make is moving the `SELECT *
FROM ai.openai_list_models();` test into the self-hosted section.

docs/vectorizer.md
```diff
@@ -69,6 +69,13 @@ Timescale Cloud or on a self-hosted Postgres server.
...
+```sql
+SELECT *
+FROM ai.openai_list_models();
+```
+If you get a list of available models you are all set.
```
JamesGuthrie, line 77: `ai.openai_list_models()` won't work as described on Timescale Cloud.

JamesGuthrie (changes requested):

docs/vectorizer.md
jgpruitt: it works on cloud. The secret handling is awkward to deal with securely, but that's an issue across the board -- not specific to this. I think this is a good change.

jgpruitt: all the parameters in versions 0.1.0, 0.2.0, and 0.3.0 use leading underscores. it was just a habit I have, because leading underscores prevent name clashes with keywords, reserved words, and object/column names.

docs/vectorizer.md
jgpruitt: why not?

jgpruitt: well, you have to run `select set_config('ai.openai_api_key', '<your-key-here>', false);` first

docs/vectorizer.md
jgpruitt: we need the secrets/env-var extension. some way to query cloud secrets

jgpruitt: you can manually do `select set_config('ai.openai_api_key', '<your-key-here>', false);` to set it in your db session, but that's no fun

Askir: How do you use this function on cloud right now? Would you have to provide they api key as a parameter?

billy-the-fish:
@Askir I've moved the updates in this PR, less the technical implementations under discussion to https://github.com/timescale/pgai/pull/158. Are you ok to close this PR?

PR #152, opened 2024-10-21
fix: make better defaults for indexing, scheduling, grants

jgpruitt:


docs/vectorizer-api-reference.md
```diff
@@ -404,10 +398,35 @@ Key points about indexing:
...
+- [ai.indexing_default](#aiindexing_default): when you do not want indexes created automatically.
 - [ai.indexing_none](#aiindexing_none): when you do not want indexes created automatically.
 - [ai.indexing_diskann](#aiindexing_diskann): configure indexing using the [DiskANN algorithm](https://github.com/timescale/pgvectorscale).
 - [ai.indexing_hnsw](#aiindexing_hnsw): configure indexing using the [Hierarchical Navigable Small World (HNSW) algorithm](https://en.wikipedia.org/wiki/Hierarchical_navigable_small_world).
 
+### ai.indexing_default
+
+You use `ai.indexing_default` to use the platform-specific default value for indexing.
+
+On Timescale Cloud, the default is `ai.indexing_diskann()`. On self-hosted, the default is `ai.indexing_none()`.
```
alejandrodnm, line 410: Should we explain why?

docs/vectorizer-api-reference.md
```diff
@@ -518,10 +537,35 @@ considerations.
 
 The available functions are:
 
+- [ai.scheduling_default](#aischeduling_default): to use the plwhen you want manual control over when the vectorizer runs. Use this when you're using an external scheduling system, as is the case with self-hosted deployments.
```
alejandrodnm, line 540:
```suggestion
- [ai.scheduling_default](#aischeduling_default): to use when you want manual control over when the vectorizer runs. Use this when you're using an external scheduling system, as is the case with self-hosted deployments that use the pgai vectorizer worker.
```

I added the last part to tie it with the vectorizer worker. Feel free to remove that part, the first part it's a typo

docs/vectorizer-api-reference.md
```diff
@@ -518,10 +537,35 @@ considerations.
...
 - [ai.scheduling_none](#aischeduling_none): when you want manual control over when the vectorizer runs. Use this when you're using an external scheduling system, as is the case with self-hosted deployments.
 - [ai.scheduling_timescaledb](#aischeduling_timescaledb): leverages TimescaleDB's robust job scheduling system, which is designed for reliability and scalability. Use this when you're using Timescale Cloud.
 
 
+### ai.scheduling_default
+
+You use `ai.scheduling_default` to use the platform-specific default scheduling configuration.
+
+On Timescale Cloud, the default is `ai.scheduling_timescaledb()`. On self-hosted, the default is `ai.scheduling_none()`.
```
alejandrodnm, line 549: Should we explain why it's different in one vs the other?

projects/extension/sql/idempotent/011-grant-to.sql
```diff
@@ -0,0 +1,13 @@
+-------------------------------------------------------------------------------
+-- grant_to
+create or replace function ai.grant_to(variadic grantees name[]) returns name[]
+as $func$
+    select pg_catalog.array_agg(cast(x as name))
+    from (
+        select pg_catalog.unnest(grantees) x
+        union
+        select pg_catalog.unnest(pg_catalog.string_to_array(pg_catalog.current_setting('ai.grant_to_default', true), ',')) x
```
jgpruitt, line 9: image
Having a username with a comma in it is unlikely but possible.

docs/vectorizer-api-reference.md
JamesGuthrie: This was actually a copy-paste error, thanks for the catch.

projects/extension/sql/idempotent/011-grant-to.sql
jgpruitt: not worth the effort

docs/vectorizer-api-reference.md
jgpruitt: done

PR #153, opened 2024-10-22
chore: add self hosted quick start

Askir:


docs/self-hosted-getting-started.md
```diff
@@ -0,0 +1,99 @@
...
+       destination => 'blog_contents_embeddings',
+       embedding => ai.embedding_openai('text-embedding-3-small', 768),
+       chunking => ai.chunking_recursive_character_text_splitter('contents'),
+       scheduling => ai.scheduling_none()
+    );
+    ```
+    If you check the logs of the vectorizer worker, you should see that it has picked up the table and is processing it.
+
+
+5. You'll need to restart the vectorizer now, since it fails to run if pgai is not installed or no vectorizer is defined yet. A simple `docker-compose up -d` should do the trick.
```
Askir, line 74: This is quite ugly, can we change the CLI so that it works even if pgai is not installed and there is no vectorizer yet? logging a warning/error is fine but the container crashing is at least for this tutorial annoying (I think also in normal use?)

docs/self-hosted-getting-started.md
```diff
@@ -0,0 +1,99 @@
...
+
+
+6. Now we can run a simple semantic search query to see the embeddings in action:
+    ```sql
+    SELECT
+        chunk,
+        embedding <=>  ai.openai_embed('text-embedding-3-small', 'pgai', _dimensions=>768) as distance
+    FROM blog_contents_embeddings
+    ORDER BY distance;
+    ```
```
Askir, line 84: There is a bit of a problem where this query will return empty results if the worker hasn't run yet. I think I will add to the compose that the cli runs every 5 seconds for this tutorial.
Right now this is not an issue because of the restart above which instantly triggers another embedding run.

docs/self-hosted-getting-started.md
```diff
@@ -0,0 +1,99 @@
...
+
+4. Now we need to create a vectorizer for the table. The vectorizer worker will pick this one up and create the corresponding embeddings for us:
+    ```sql
+    SELECT ai.create_vectorizer(
+       'blog'::regclass,
+       destination => 'blog_contents_embeddings',
+       embedding => ai.embedding_openai('text-embedding-3-small', 768),
+       chunking => ai.chunking_recursive_character_text_splitter('contents'),
+       scheduling => ai.scheduling_none()
```
Askir, line 68: This is a bug in the new version I assume.

docs/self-hosted-getting-started.md
```diff
@@ -0,0 +1,99 @@
...
+        environment:
+          POSTGRES_PASSWORD: postgres
+          OPENAI_API_KEY: your-api-key
+        ports:
+          - "5432:5432"
+        volumes:
+          - ./data:/var/lib/postgresql/data
+    
+      vectorizer:
+        image: timescale/pgai-vectorizer-worker:rc
```
alejandrodnm, line 26: We only have available `timescale/pgai-vectorizer-worker:0.1.0rc4`

docs/self-hosted-getting-started.md
```diff
@@ -0,0 +1,99 @@
...
+        image: timescale/timescaledb-ha:cicd-024349a-arm64
+        environment:
+          POSTGRES_PASSWORD: postgres
+          OPENAI_API_KEY: your-api-key
+        ports:
+          - "5432:5432"
+        volumes:
+          - ./data:/var/lib/postgresql/data
+    
+      vectorizer:
```
alejandrodnm, line 25: let call this `pgai-vectorizer-worker`

docs/self-hosted-getting-started.md
```diff
@@ -0,0 +1,99 @@
...
+## Pre-requisites
+- Docker installed on your machine
+- An OpenAI API key. You can create one [here](https://platform.openai.com/api-keys).
+
+## Setup
+
+1. Run the following docker-compose file to start a TimescaleDB instance and the Vectorizer worker: 
+    ```yaml
+    services:
+      db:
```
alejandrodnm, line 15: Add a `healthcheck` like in https://github.com/timescale/pgai/pull/113/files#diff-facaded51b7d1c9656ae43b449b69aa398fee9129321a0001d97048c00c8cc1cR9-R15 so that the other container starts when this one is ready to accept connections. Otherwise, you might get error while postgres is starting.

docs/self-hosted-getting-started.md
```diff
@@ -0,0 +1,99 @@
...
+        volumes:
+          - ./data:/var/lib/postgresql/data
+    
+      vectorizer:
+        image: timescale/pgai-vectorizer-worker:rc
+        environment:
+          VECTORIZER_DB_URL: postgres://postgres:postgres@db:5432/postgres
+          OPENAI_API_KEY: your-api-key
+        depends_on:
+          - db
```
alejandrodnm, line 31: If you implement the healthcheck, then change this to:

```
    depends_on:
      db:
        condition: service_healthy
```

docs/self-hosted-getting-started.md
```diff
@@ -0,0 +1,99 @@
...
+If you want to see how it works, follow these steps to setup 
+a self-hosted database instance with automatic embedding vectorization:
+
+## Pre-requisites
+- Docker installed on your machine
+- An OpenAI API key. You can create one [here](https://platform.openai.com/api-keys).
+
+## Setup
+
+1. Run the following docker-compose file to start a TimescaleDB instance and the Vectorizer worker:
```
alejandrodnm, line 12: We should add the command that starts the docker-compose. 

We can also start just the db with `docker compose up -d db`, let the user run through al the setup. Then tell them to run `docker compose up -d pgai-vectorizer-worker` to start the vectorizer worker. 

If they start everything together, the pgai-vectorizer-worker container will fail because the pgai is not installed. 

Another option is to do setup in a sql init file, and have that be loaded when the service starts.

docs/self-hosted-getting-started.md
```diff
@@ -0,0 +1,99 @@
...
+      vectorizer:
+        image: timescale/pgai-vectorizer-worker:rc
+        environment:
+          VECTORIZER_DB_URL: postgres://postgres:postgres@db:5432/postgres
+          OPENAI_API_KEY: your-api-key
+        depends_on:
+          - db
+    ```
+
+2. Connect to your db instance with your DB client of choice. We need to enable the pgai extension and create a simple blog table with the following schema:
```
alejandrodnm, line 34: We should add a command to connect directly, for those that don't have a preference or any program installed. If we set `name: pgai`, then we can connect to the db with:

```
docker exec -it pgai-db-1 psql -U postgres
```

docs/self-hosted-getting-started.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Pre-requisites
+- Docker installed on your machine
+- An OpenAI API key. You can create one [here](https://platform.openai.com/api-keys).
+
+## Setup
+
+1. Run the following docker-compose file to start a TimescaleDB instance and the Vectorizer worker: 
+    ```yaml
+    services:
```
alejandrodnm, line 14: I'd add `name: pgai` so that we can assume `pgai-` will be the prefix for all the containers.

docs/self-hosted-getting-started.md
alejandrodnm: This can be solve by starting first the db with `docker compose up -d db`, and at this point we run `docker compose up -d pgai-vectorizer-worker`.

docs/self-hosted-getting-started.md
alejandrodnm: We can tell them to run:

```
select ai.vectorizer_queue_pending(1) \watch
```

To monitor how many pending items are left. When the query returns `0` then they can proceed.

docs/self-hosted-getting-started.md
alejandrodnm: If we go this route we need to tell them that if they want to see the logs use

```
docker compose logs -f
```

Just in case something goes wrong they can have the logs.

docs/self-hosted-getting-started.md
Askir: oh yes somehow I must have messed this up while copy pasting back and forth 😅

docs/self-hosted-getting-started.md
Askir: I like the separate compose up commands I think. Keeps things simple, and I can remove the `depends_on` then as well.

I want to really get to that successful search-query in the least complex way possible.

Askir:
@alejandrodnm Are you fine with adding:
```
select ai.vectorizer_queue_pending(1) \watch
```
and
```
docker compose logs -f
```
In a troubleshooting section at the bottom?  In a successful run you wont need it I hope. I'll then also add a CTA to join the discord if more help is needed.

docs/self-hosted-getting-started.md
smoya: Indeed `ai.scheduling_none()` should be the default for self hosted -> https://github.com/timescale/pgai/pull/89/commits/2e9046f4ebd3d79004bc5f45c5dcca84908efd4b#diff-5f65bc4a75b352eedf28c26480e8451d5f605d3fac22159361fd4f4ac7388f74R542

Askir: Fixing it here: https://github.com/timescale/pgai/pull/154

docs/self-hosted-getting-started.md
alejandrodnm: You can leave the depends_on , that way if you stop and restart everything, then the worker will wait for the DB to be ready to accept connections. Users will be fine, since data will be stored in the volume, so they don't have to go through all the set up again.

docs/self-hosted-getting-started.md
Askir: I like the split up `docker-compose up` so far, makes it simple and removes the need for depends_on etc. 
For logs I added this at the bottom in the troubleshooting section.

docs/self-hosted-getting-started.md
Askir: Also added this in the troubleshooting section 👍

docs/self-hosted-getting-started.md
Askir: Since we now start the containers 1 by 1 I don't think we need this. Just bloats the example and makes a user have to understand a more complex compose. I also removed the `depends_on`.

docs/self-hosted-getting-started.md
Askir: Done 👍

docs/self-hosted-getting-started.md
Askir: I've switched this again to just vectorizer-worker since the namespace of the compose is now pgai and the container is otherwise called `pgai-pgai-vectorizer-worker`. Is that fine with you?

Askir: I've switched this again to simply `vectorizer-worker` to avoid the container being called `pgai-pgai-vectorizer-worker` because I've also added the compose namespace (pgai). Is that fine with you?

docs/self-hosted-getting-started.md
Askir: Not sure about this, I prefer to keep it as simple as possible. The more i add to the compose the more complex it makes it look to a potential user. And we want the setup to look and be as trivial as possible imo.

I could make the same argument about removing the volume actually. Maybe somewhere we have to draw a line between usability and ease of understanding.

billy-the-fish:
@Askir I've moved the updates in this PR, less the technical implementations under discussion to https://github.com/timescale/pgai/pull/158. Are you ok to close this PR?

PR #154, opened 2024-10-22
fix: missing else in default scheduling

Askir:


jgpruitt (changes requested):
Run 'make build' so that the ai--0.4.0.sql is updated with the change too

jgpruitt (approved):

JamesGuthrie (approved):
LGTM.

There's a redundant return in the equivalent indexing function (which had the same problem as this, but a test caught and exposed it, so I fixed it - badly). Would be great if you could fix it while you're in there, but not necessary.

```diff
diff --git a/projects/extension/sql/idempotent/009-indexing.sql b/projects/extension/sql/idempotent/009-indexing.sql
index d9c901c..68506f6 100644
--- a/projects/extension/sql/idempotent/009-indexing.sql
+++ b/projects/extension/sql/idempotent/009-indexing.sql
@@ -69,7 +69,6 @@ begin
         else
             return ai.indexing_none();
     end case;
-    return ai.indexing_none();
 end;
 $func$ language plpgsql volatile security invoker
 set search_path to pg_catalog, pg_temp
```

PR #158, opened 2024-10-23
Vectorizer naming update

billy-the-fish:

PR #160, opened 2024-10-23
Adn/env vars

alejandrodnm:


projects/pgai/pgai/cli.py
```diff
@@ -117,13 +117,22 @@ def convert(self, value, param, ctx) -> int:  # type: ignore
...
+    return logging.getLevelName("INFO")  # type: ignore
+
+
 @click.command(name="worker")
 @click.version_option(version=__version__)
 @click.option(
     "-d",
     "--db-url",
     type=click.STRING,
-    envvar="VECTORIZER_DB_URL",
```
jgpruitt, line 126: The docs need to be updated to reflect the change in env var name.
https://github.com/timescale/pgai/blob/vectorizer/docs/self-hosting-vectorizer.md?plain=1#L42

jgpruitt (changes requested):
one small thing

thedodd (approved):

projects/pgai/pgai/cli.py
alejandrodnm: That's part of @billy-the-fish PR

Askir (approved):
Ayy thanks for fixing this, we can revert it once we drop support for 3.10 I suppose.

projects/pgai/pgai/cli.py
alejandrodnm: https://github.com/timescale/pgai/pull/158#discussion_r1813163360

jgpruitt: ahh. ok

jgpruitt (approved):

PR #176, opened 2024-10-28
chore: add logo to pgai pypi

alejandrodnm:


jgpruitt (approved):

projects/pgai/README.md
```diff
@@ -1,3 +1,7 @@
+<p align="center">
+    <img height="200" src="https://github.com/timescale/pgai/blob/main/docs/images/pgai_logo.png" alt="pgai"/>
```
smoya, line 2: Why fixing a heigh here?
```suggestion
    <img src="https://github.com/timescale/pgai/blob/main/docs/images/pgai_logo.png" alt="pgai"/>
```

alejandrodnm: I copy pasted from the original README.

smoya: Looks way better without that restriction in all resolutions

PR #183, opened 2024-10-30
chore: update reference inline with url changes

RickVM:
The old url returns a 404, I found the working page at the updated url


alejandrodnm (approved):
Thanks for this 🙏🏼

PR #188, opened 2024-10-31
chore: migrate to uv and hatch

Askir:
Using `uv.lock` for a reproducible dev environment and more stable/easy builds.
I created a minimal project with a super close publish action here: https://github.com/Askir/minimal_math and it works so I think this is also fine to be merged.

Todo:
- ~Add development docs on how to use it to install and manage deps~
- ~Test the publish process somehow?~


.github/workflows/ci.yml
```diff
@@ -60,15 +60,21 @@ jobs:
...
+          version: "0.4.29"
+          enable-cache: true
+          cache-dependency-glob: "./projects/pgai/uv.lock"
+
+      - name: "Set up Python"
         uses: actions/setup-python@v5
         with:
-          python-version: "3.12"
-          cache: "pip" # caching pip dependencies
+          python-version-file: "./projects/pgai/.python-version"
```
Askir, line 73: We have the minimum version of python set to 3.10 but are using 3.12 in the docker image. I feel like we might want to use a test matrix here to ensure it works with 3.11, 3.12, 3.13?

For development I think it makes sense to use the minimum python version so that we don't accidentally make use of newer features than we wanted to support.

projects/pgai/.python-version
```diff
@@ -0,0 +1 @@
+3.10
```
Askir, line 1: If you don't use a `.python-version` file, uv will take the newest python version that's compatible with your pyproject.toml. I don't think this is desirable for development as it makes us potentially unaware that we are using newer python features than we want. So explicitly setting this here.

projects/pgai/Makefile
```diff
@@ -1,4 +1,3 @@
-
 VERSION := $(shell awk "/^__version__ = .*/ {gsub(/__version__ = |\"/, \"\"); print}" ./pgai/__init__.py)
```
Askir, line 2: Not sure we need the indirection of a Makefile? Most steps simply do `uv run step` so instead of typing that directly you have to write `make step` 🤷‍♂️ .

But happy to keep it if that's what people are used to.

projects/pgai/pyproject.toml
```diff
@@ -1,29 +1,44 @@
 [build-system]
-requires = ["setuptools"]
-build-backend = "setuptools.build_meta"
+requires = ["hatchling"]
+build-backend = "hatchling.build"
```
Askir, line 3: Hatchling allows building from pyproject.toml directly. Setuptools is regarded as legacy tooling. The folks at uv say eventually they will release their own build tool but for now this is their recommendation.

.github/workflows/ci.yml
JamesGuthrie: I think the test matrix is a good idea. Agree on using the python version file with 3.10 for development.

projects/pgai/.python-version
JamesGuthrie: `.python-version` files support comments, can we add one here explaining why we use 3.10?

```suggestion
# the 'timescaledb-ha' docker image uses 3.10, so do we
3.10
```

projects/pgai/pyproject.toml
```diff
@@ -70,29 +79,32 @@ target-version = "py310"
 docstring-code-format = true
```
JamesGuthrie, line 79: (This pertains to line 76, which is a few lines above this comment). Apparently we can drop the `target-version` from the ruff config if we have `project.requires-python` set (see https://docs.astral.sh/ruff/settings/#target-version)

.github/workflows/ci.yml
jgpruitt: The timescaledb ha image uses Python 3.10 which is therefore what we have to use for the extension. I believe the pgai library can use whatever we wish since it is separate.

projects/pgai/Makefile
jgpruitt: I'm fine with not having a Makefile if we instead have all the commands that a person would need to know documented somewhere. New people shouldn't have to first learn uv and guess at how we expect things to be done.

jgpruitt: IMO the Makefile is almost more valuable as documentation than functionality

projects/pgai/.python-version
jgpruitt: The pgai library doesn't run in the timescaledb-ha image. So, we may not need this constraint.

.github/workflows/ci.yml
Askir: yes, but if as a user your project is e.g. not yet python 3.12 you will run into compatibility issues when you try to install the package.

Tbf right now there is little need to install pgai into an existing project but if we add more python helpers, this might become important and I would suggest that we roughly follow the [release and support cadence of python](https://devguide.python.org/versions/) for this.

projects/pgai/.python-version
Askir: added a comment for the actual reason :)

projects/pgai/Makefile
Askir: sure, happy to keep it!

projects/pgai/pyproject.toml
Askir: removed

projects/pgai/.python-version
Askir: Hmm the comments do not actually seem to be allowed:
```
  Error: The version '3.10
  # We try to somewhat follow the python release schedule for supported versions to allow more users to use our library.
  # Therefore we are about a year behind the latest python release.' with architecture 'x64' was not found for Ubuntu 22.04.
  ```
  
  Adding this to the dev docs instead.

.github/workflows/ci.yml
Askir: I'll do test matrix in a separate PR.

projects/pgai/.python-version
JamesGuthrie: Hmmm. interesting. I did try out comments with uv locally before making this comment - maybe it was something that was added relatively recently.

jgpruitt (approved):

PR #192, opened 2024-11-01
feat: allow SQL to be gated behind feature flags

jgpruitt:
We would like to avoid long-lived feature branches and the git headaches that entails.  Yet some features take a while to mature to a level that we are happy to ship and support in production environments. Avoiding feature branches means we may be in a position where we need to ship but have SQL code that is not fully ready. Making changes to SQL objects once they ship is tricky and painful and should be minimized.

We want to:

- avoid long-lived feature branches. 
- merge progress to main
- minimize database migrations that alter existing objects
- avoid installing prerelease code in production databases
- allow prerelease code to be installed/tested/trialed in an opt-in manner
- warn when prerelease code is installed
- avoid supporting extension upgrades after prerelease code has been installed

To this end, we allow prerelease functionality to be gated behind feature flags. We only build and ship prerelease code if the version being built includes a prerelease tag. If the version is a production version, we omit the prerelease code altogether. Furthermore, if the feature flag is not enabled, the prerelease code is NOT executed/installed.

Incremental or idempotent files numbered greater than 899 must have a comment at the top of the file with a prefix of `--FEATURE-FLAG: ` followed by the name of a feature flag. These files are not executed AT ALL unless a session-level GUC 
like `ai.enable_feature_flag_<feature-flag>` is set to `true` when the `create extension` or `alter extension ... update` statement is executed. 

Incremental/idempotent files numbered less than 900 must NOT have a feature flag 
comment.

Zero or more feature flag GUCs may be enabled at once. 

Flags that are enabled when creating/updating the extension are recorded in the `ai.feature_flag` table. Since feature-flag-gated code is only built and shipped for prerelease versions in the first place, it can only be installed in an environment not intended for production in the first place. If one or more feature flag GUCs is enabled, all bets are off. The code is pre-production, may not work, is not supported, **and upgrades from this state are not supported**. This is a dead-end state. We do not generate upgrade paths from prerelease versions to production versions.

When working on pre-release features, tests can be written for these features. The tests must create a database, enable the correct feature flag, create the extension, and then proceed with testing.

**Example:**

```sql
select set_config('ai.enable_feature_flag_my_new_feature', 'true', false);
create extension ai cascade;
NOTICE:  installing required extension "vector"
NOTICE:  installing required extension "plpython3u"
WARNING:  Feature flag "ai.enable_feature_flag_my_new_feature" has been enabled. Pre-release software will be installed. This code is not production-grade, is not guaranteed to work, and is not supported in any way. Extension upgrades are not supported once pre-release software has been installed.
CREATE EXTENSION
```

Once a feature that was gated is finally blessed and "finished", we need a final PR that moves the incremental and idempotent SQL files to their final (less than 900) places and removes the feature flag comment. Changes to existing 
code/structures that were previously in a gated file could be folded into the original file. Tests must be updated to remove the feature flag references. Frequently, working on new features requires changes to existing SQL code/structures. 
In the case of gated features, we are not able to make changes inline where the original code is defined. These changes have to be included in the gated files. While this may be somewhat awkward, it works, and it clearly delineates what 
changes to existing stuff are required for a new feature.

#### Incremental SQL

Incremental SQL files are migrations. They are only safe to be executed exactly once and in a deterministic order. You cannot modify an Incremental SQL file once it has shipped.

The ai.migration table is used to enforce these guarantees.

These files usually consist of things like:

- Create table
- Create index
- Create constraint
- Create (event) trigger
- Config data

Once these kinds of things are shipped, they are more challenging to make changes to. You cannot just modify the original code in-place; you have to introduce a new migration script that uses ALTER commands to change what is already there carefully. Some things, like changing a column’s data type, are not straightforward. Some changes you might want to make, like reordering the columns on a table, cannot be done practically.

These types of objects should be expected to change quite a bit and frequently during the course of developing a feature.

Migrations can be dangerous and difficult to make robust:

- [common db schema change mistakes](https://postgres.ai/blog/20220525-common-db-schema-change-mistakes)
- [zero downtime postgres migrations](https://postgres.ai/blog/20210923-zero-downtime-postgres-schema-migrations-lock-timeout-and-retries)
- [postgres schema change gotchas](https://medium.com/preply-engineering/postgresql-schema-change-gotchas-bf904e2d5bb7)

#### Idempotent SQL

Idempotent SQL files are safe to run repeatedly (obviously). 

These files consist of things like:

- Create or replace function/procedure
- Create or replace view

Care has to be taken when modifying these kinds of objects. Changing the signature of a function/procedure will not replace a function/procedure but create a new one with an overridden name. The same applies to views. If you make these changes, you must introduce a drop function/procedure/view statement into the script. Otherwise, databases can end up in different states depending on the upgrade path, leading to a support nightmare. Changing signatures and renaming functions happens very frequently during the course of developing a feature.

We run these files in a deterministic order, but it is safe to reorder them between releases. Views and functions/procedures defined using the SQL language require that any dependent objects already exist for the creation to succeed. Functions/procedures defined in other languages, like plpgsql, only require dependent objects to exist at execution time.

#### Python

Some of the database functions in pgai are backed by Python installed “outside the database” as a library (/usr/local/lib/pgai/<version>). Every released version of pgai gets a new complete copy of the Python code and its dependencies. The SQL functions “know” what version of the extension is running and look up the appropriate copy of the library and dependencies using the right path.

Since each extension version has a bespoke copy of the Python library, the Python library only has to ensure that it works with the latest version.


Askir:
Wow, love the commitment behind trying this and it does read quite well thought out to me already! Some thoughts on the questions:
> How do we handle tests for existing features when those features are changed by new-feature-flag-gated features? We somehow have to test both versions.

We will have to write tests both for disabled and enabled feature flags. So when a new feature changes an existing one the old tests still need to pass and the new tests only need to pass after enabling the feature flag. After the test we need to reset the database state, so all incremental SQL is reset.

> What if multiple gated features need to modify the same existing features? That could get messy and complicated quickly.
I think this makes visible what would otherwise be hidden in merge conflicts. I think on a case-by-case basis you have 3 options:
1. Use the same feature-flag for two features -> this is trivial but makes it impossible to enable one without the other so they have to ship together
2. Separate into two flags -> Make only one of the features depend on the other so you have to enable both feature-flags for the "later one", I think this will usually be the best idea.
3. Separate into two flags but write migrations independent from each other -> This seems quite hard, we might have to write tests for all 4 cases:
- No feature flag on
- Feature A enabled
- Feature B enabled
- Feature A and B enabled

jgpruitt:
>Should we record what feature flags were enabled if any so that we can refer to it after-the-fact at runtime? I think yes.

I think yes. It may be necessary for writing tests. The tests will need to know what feature flags are enabled and may need to be skipped accordingly.

Also, if we want to enforce that no upgrades are allowed from an installation that has pre-release software, we'll need a way to detect that that has happened.

jgpruitt:
>Do we need to keep gated idempotent files above 899, or should we allow them to be mixed in with production-grade files? Reordering idempotent files is not a big deal. Keeping them separate would help reduce confusion though.

I think we start with this restriction for the sake of clarity and only remove it if it becomes an issue.

jgpruitt:
>Is a whole file the right unit of size? I think so.

Yes, I think so. Mixing production and pre-production code in the same file will start looking like #ifdef pre-processor macros from C, and I think we should avoid that level of complexity until we prove we need it.

jgpruitt:
>Should we also consider the extension version? If the current version is a pre-production version (e.g. 0.4.1-rc1) then we include the feature-flag gated code. If the current version does not have a pre-production tag (e.g. 0.4.1), we exclude all the feature-flag gated stuff -- the only code shipped is the blessed stuff. ??

I'm leaning towards yes on this too. Omitting the pre-production code when building a production version feels clean.

JamesGuthrie:
As discussed, in general I like this idea. For now (until we gather some more experience) I would treat this only to be used for internal testing.

As for your open questions, my opinions:

> Should we record what feature flags were enabled if any so that we can refer to it after-the-fact at runtime? I think yes.

Can't hurt.

> How do we handle tests for existing features when those features are changed by new-feature-flag-gated features? We somehow have to test both versions.

Yes, we will have to test both. This also means that the "release feature" cleanup PR has a bit more work to do.

> What if multiple gated features need to modify the same existing features? That could get messy and complicated quickly.

I suggest that we try to avoid this 🙈 

> Do we need to keep gated idempotent files above 899, or should we allow them to be mixed in with production-grade files? Reordering idempotent files is not a big deal. Keeping them separate would help reduce confusion though.

Keep them above 899 to reduce confusion.

> Is a whole file the right unit of size? I think so.

Ya.

> Should we also consider the extension version? If the current version is a pre-production version (e.g. 0.4.1-rc1) then we include the feature-flag gated code. If the current version does not have a pre-production tag (e.g. 0.4.1), we exclude all the feature-flag gated stuff -- the only code shipped is the blessed stuff. ??

I like this idea: it ensures that we only have feature-flag stuff in non-rc releases, which is in line with my initial suggestion that we (for now) treat this as an internal-only tool.

jgpruitt:
Updated the description to reflect what was decided on and coded.

projects/extension/build.py
```diff
@@ -116,13 +186,33 @@ def incremental_sql_files() -> list[Path]:
...
+                f"incremental sql files under 900 must be NOT gated by a feature flag: {path.name}",
+                file=sys.stderr,
+            )
+            sys.exit(1)
+        if this >= 900 and not ff:
+            print(
+                f"incremental sql files over 899 must be gated by a feature flag: {path.name}",
+                file=sys.stderr,
+            )
+            sys.exit(1)
```
adolsalamanca, line 215: Nit, would it make sense to extract this block of code extracted to a common func to be used in both `check_incremental_sql_files` and `check_idempotent_sql_files`? Seems to be checking the same conditions.

projects/extension/sql/head.sql
```diff
@@ -51,3 +51,10 @@ begin
 end;
 $bootstrap_extension$;
 
+-- records any feature flags that were enabled when installing
+-- a prerelease version of the extension
+create table if not exists ai.feature_flag
+( "name" text not null primary key
+, applied_at_version text not null
+, applied_at timestamptz not null default pg_catalog.clock_timestamp()
```
adolsalamanca, line 59: neat 👌🏽

adolsalamanca (approved):
Looks good, left a nit

projects/extension/build.py
```diff
@@ -90,16 +108,68 @@ def idempotent_sql_files() -> list[Path]:
     return paths
 
 
+def parse_feature_flag(path: Path) -> str | None:
+    with path.open(mode="rt", encoding="utf-8") as f:
+        line = f.readline()
```
smoya, line 113: Nit: Would it make sense to skip possible empty lines in the top of the file here?

I believe this should do the job but please double check.


```suggestion
        line = f.readline()
        while line.strip() == '':
            line = f.readline()
```

projects/extension/build.py
```diff
@@ -90,16 +108,68 @@ def idempotent_sql_files() -> list[Path]:
...
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
+
+def file_number(path: Path) -> int:
+    return int(path.name[0:3])
+
+
 def check_idempotent_sql_files(paths: list[Path]) -> None:
```
smoya, line 141: I suggest we add a comment mentioning the paths are sorted.
```suggestion
def check_idempotent_sql_files(paths: list[Path]) -> None:
```

projects/extension/build.py
```diff
@@ -90,16 +108,68 @@ def idempotent_sql_files() -> list[Path]:
...
+        if path.name == "999-privileges.sql":
+            break
+        this = file_number(path)
+        if this < 900 and this != prev + 1:
             print(
                 f"idempotent sql files must be strictly ordered. this: {this} prev: {prev}",
                 file=sys.stderr,
             )
             sys.exit(1)
+        if this >= 900 and this == prev:  # allow gaps in pre-production scripts
```
smoya, line 154:
```suggestion
        # avoiding file number duplication
        if this >= 900 and this == prev:  # allow gaps in pre-production scripts
```

projects/extension/build.py
```diff
@@ -90,16 +108,68 @@ def idempotent_sql_files() -> list[Path]:
...
+            print(
+                f"idempotent sql files must not have duplicate numbers. this: {this} prev: {prev}",
+                file=sys.stderr,
+            )
+            sys.exit(1)
+        ff = parse_feature_flag(path)
+        if this < 900 and ff:
```
smoya, line 161:
```suggestion
        # feature flagged file numbers should be between 900 and 999
        if this < 900 and ff:
```

projects/extension/build.py
```diff
@@ -90,16 +108,68 @@ def idempotent_sql_files() -> list[Path]:
...
+            print(
+                f"idempotent sql files under 900 must be NOT gated by a feature flag: {path.name}",
+                file=sys.stderr,
+            )
+            sys.exit(1)
+        if this >= 900 and not ff:
```
smoya, line 167:
```suggestion
        # only feature flagged files go over 899
        if this >= 900 and not ff:
```

DEVELOPMENT.md
```diff
@@ -250,6 +250,78 @@ targets in favor of the Python-specific make targets:
...
+- warn when prerelease code is installed
+- avoid supporting extension upgrades after prerelease code has been installed
+
+To this end, we allow prerelease functionality to be gated behind feature flags.
+We only build and ship prerelease code if the version being built includes a 
+prerelease tag. If the version is a production version, we omit the prerelease 
+code altogether. Furthermore, if the feature flag is not enabled, the prerelease 
+code is NOT executed/installed. 
+
+Incremental or idempotent files numbered greater than 899 must have a comment at
```
smoya, line 277: Your implementation only takes 3 digits. I know we won't be potentially having such amount of files but I believe better to mention it here.
```suggestion
Incremental or idempotent files numbered greater than 899 and lower than 999 must have a comment at 
```

projects/extension/build.py
```diff
@@ -116,13 +186,33 @@ def incremental_sql_files() -> list[Path]:
 def check_incremental_sql_files(paths: list[Path]) -> None:
     prev = 0
     for path in paths:
-        this = int(path.name[0:3])
-        if this != prev + 1:
+        check_sql_file_name(path)
+        this = file_number(path)
```
smoya, line 190: WDYT about simplifying this to have only one function that uses a little modification on your regex to extract the both match and extract the file number?

Something like 
```python
def extract_sql_file_number(path: Path) -> int:
    pattern = r"^(\d{3})-[a-z][a-z_-]*\.sql$"
    matches = re.match(pattern, path.name, re.IGNORECASE) # I guess we want to be non case sensitive

    if not matches:
        print(
            f"{path} file name does not match the pattern {pattern}",
            file=sys.stderr,
        )
        sys.exit(1)
    return matches.group(1)
```

smoya (commented):
Left some nits on the python code. Seems to look good otherwise. 🚀 

Want to take a moment to say the description of this PR is 💯  -- It is explaining the issue, giving context, and proposing a solution to a non-trivial problem.
Clear, concise, and i believe understandable even to individuals with no experience in such a codebase.

projects/extension/build.py
jgpruitt: I want to enforce that it be the first line.

projects/extension/build.py
jgpruitt: done

projects/extension/build.py
jgpruitt: done

projects/extension/build.py
jgpruitt: done

PR #195, opened 2024-11-04
Fix interpolation call: target_schema to target_name

jonatas:
I just found an error showing the following message:

```
 ERROR:  an object named public.public already exists. specify an alternate destination or target_table explicitly 
```
 
 Digging the code I see the table name was missing and the schema was repeated.
 
 
 ## How to test
 
 1. create a table
 2. create a vectorizer based on the table
 3. drop the table
 4. `truncate ai.vectorizer cascade`;
 5. repeat step 1 and then step 2 will fail.


jgpruitt (approved):
thanks again!

PR #197, opened 2024-11-05
feat: Add example python project

Askir:


examples/code-llm-sync/alembic/versions/2024_11_04_1147-497e69a2bca9_add_code_files_table.py
```diff
@@ -0,0 +1,38 @@
+"""empty message
+
+Revision ID: 497e69a2bca9
+Revises:
+Create Date: 2024-11-04 11:47:57.345379
+
+"""
```
Askir, line 7: This file was generated with `alembic revision --autogenerate` which compares the database to the sqlalchemy models and generates a naive version of a diff. Works quite well for adding new tables. This of course doesn't work for our views.

examples/code-llm-sync/alembic/versions/2024_11_04_1148-bb79790304f7_add_embeddings_to_table.py
```diff
@@ -0,0 +1,54 @@
...
+        chunking => ai.chunking_recursive_character_text_splitter(
+            'contents',
+            chunk_size => 1000,
+            chunk_overlap => 200
+        ),
+        formatting => ai.formatting_python_template(
+            'File: $file_name\n\nContents:\n$chunk'
+        )
+    );
+    """)
```
Askir, line 41: This is how I am creating a vectorizer via migrations. I guess it works but it's not as pretty as the native alembic functions. It's probably possible to have nice wrappers around those somehow.

examples/code-llm-sync/db/models.py
```diff
@@ -0,0 +1,54 @@
...
+        Vector(768), nullable=False
+    )  # 768 dimensions for text-embedding-3-small
+    chunk_seq = Column(Integer, nullable=False)
+
+    # Relationship back to original CodeFile
+    code_file = relationship("CodeFile", foreign_keys=[id])
+
+    @override
+    def __repr__(self) -> str:
+        return f"<CodeFileEmbedding(file_name='{self.file_name}', chunk_seq={self.chunk_seq})>"
```
Askir, line 54: It is nice that this works at all, but having to define all fields twice is a bit ugly I feel like. We can maybe provide a helper annotation or similar here. Like `@embedded` on the original model and then it automatically injects the embedding and chunk field, etc.

examples/code-llm-sync/main.py
```diff
@@ -0,0 +1,143 @@
...
+    async with AsyncSession(engine) as session:
+        yield session
+
+
+def parse_embedding_string(embedding_str: str) -> np.ndarray:
+    """Convert a pgai embedding string to a numpy array"""
+    # Remove brackets and split on commas
+    values = embedding_str.strip("[]").split(",")
+    # Convert to float array
+    return np.array([float(x) for x in values], dtype=">f4")
```
Askir, line 79: This function is currently not in use. But when I originally called `openai_embed()` it simply returned a string representation of the vector (float array). Which is extremely unfortunate if you want to use this embedding in any way afterwards. Maybe I setup `pgvector-python` in the wrong way though.

examples/code-llm-sync/main.py
```diff
@@ -0,0 +1,143 @@
...
+        super().__init__()
+
+
+@compiles(PGAIFunction)
+def _compile_pgai_embed(element, compiler, **kw):
+    return "ai.openai_embed('%s', %s, dimensions => %d)" % (
+        element.model,
+        compiler.process(element.text),
+        element.dimensions,
+    )
```
Askir, line 100: This is already part of a proposed solution, it was quite hard to pass named arguments to sqlalchemy. It does allow [basically arbitrary function calling](https://docs.sqlalchemy.org/en/20/tutorial/data_select.html#tutorial-functions) but I couldn't define the dimensions parameter somehow (maybe this is some SQL magic I don't understand).

This implementation also has another benefit though, where you can define return value and parameter types. I think this is quite the low hanging fruit, and mostly a bit of busy work to build. But has some nice benefits.

examples/code-llm-sync/main.py
```diff
@@ -0,0 +1,143 @@
...
+    results = await session.execute(
+        select(
+            CodeFileEmbedding.file_name,
+            CodeFileEmbedding.chunk,
+            CodeFileEmbedding.chunk_seq,
+            similarity_score,
+        )
+        .order_by(similarity_score.desc())
+        .limit(limit)
+    )
```
Askir, line 131: btw does putting the function twice in the code like here make it execute twice? Or is the Optimizer smart enough to understand that the result is the same since it's an idempotent function and only executes it once?

examples/code-llm-sync/tests/conftest.py
```diff
@@ -0,0 +1,104 @@
...
+
+from sqlalchemy import NullPool
+from sqlalchemy.ext.asyncio import create_async_engine, AsyncSession, async_sessionmaker
+
+from main import app
+
+# Test database configuration
+TEST_DB_URL = "postgresql+asyncpg://postgres:postgres@localhost/postgres"
+project_root = Path(__file__).parent.parent
+
```
Askir, line 21: Testing in general was a bit of a pain. But mostly because I struggled with setting up migrations and a database setup. However I am also not trying to mock anything here, which makes it a little easier.

examples/code-llm-sync/tests/conftest.py
```diff
@@ -0,0 +1,104 @@
...
+    # Wait for container to be running
+    container.reload()
+    assert container.status == "running"
+
+    yield container
+
+    # Cleanup
+    container.stop()
+    container.remove()
```
Askir, line 77: This is probably not how you'd actually want to test this, instead you can depend on pgai and call the vectorizer-worker from there. But Claude came up with this (as most of the code in this PR) and it worked, so I'm keeping it for now.

smoya:
Closing due to inactivity. Please reopen if applies

PR #198, opened 2024-11-05
perf: vector columns use storage main

jgpruitt:
Based on prior benchmarking, using storage plain (rather than extended/external) for vector columns performs much better. Unfortunately, it is difficult to determine whether a target table row will definitively fit on a single postgres page. Therefore, we will assign storage main to the vector columns, which will keep them inline unless they won't fit and toast otherwise.

NOTE: We will set storage to main for existing vectorizers (unless they have been already manually set to main or plain), but this will not take effect until a table rewrite happens. This can be done with `CLUSTER` or `VACUUM FULL` and may take a while depending on the size of the table.

>The TOAST management code recognizes four different strategies for storing TOAST-able columns on disk:
>
>PLAIN prevents either compression or out-of-line storage. This is the only possible strategy for columns of non-TOAST-able data types.
>
>EXTENDED allows both compression and out-of-line storage. This is the default for most TOAST-able data types. Compression will be attempted first, then out-of-line storage if the row is still too big.
>
>EXTERNAL allows out-of-line storage but not compression. Use of EXTERNAL will make substring operations on wide text and bytea columns faster (at the penalty of increased storage space) because these operations are optimized to fetch only the required parts of the out-of-line value when it is not compressed.
>
>MAIN allows compression but not out-of-line storage. (Actually, out-of-line storage will still be performed for such columns, but only as a last resort when there is no other way to make the row small enough to fit on a page.)

https://www.postgresql.org/docs/17/storage-toast.html#STORAGE-TOAST-ONDISK

### DIFF
--- a/projects/extension/sql/incremental/003-vec-storage.sql
+++ b/projects/extension/sql/incremental/003-vec-storage.sql
@@ -1,0 +1,31 @@
+
+-- switch the vector columns from storage external to storage main
+do language plpgsql $block$
+declare
+    _sql pg_catalog.text;
+begin
+    set local search_path = pg_catalog, pg_temp;
+
+    for _sql in
+    (
+        select pg_catalog.format
+        ( $sql$alter table %I.%I alter column embedding set storage main$sql$
+        , v.target_schema
+        , v.target_table
+        )
+        from ai.vectorizer v
+        inner join pg_catalog.pg_class k on (k.relname operator(pg_catalog.=) v.target_table)
+        inner join pg_catalog.pg_namespace n
+            on (k.relnamespace operator(pg_catalog.=) n.oid and n.nspname operator(pg_catalog.=) v.target_schema)
+        inner join pg_catalog.pg_attribute a on (k.oid operator(pg_catalog.=) a.attrelid)
+        where a.attname operator(pg_catalog.=) 'embedding'
+        and a.attstorage not in ('m', 'p') -- not main or plain
+    )
+    loop
+        raise info '%', _sql;
+        execute _sql;
+        commit;
+        set local search_path = pg_catalog, pg_temp;
+    end loop;
+end;
+$block$;
--- a/projects/extension/sql/idempotent/012-vectorizer-int.sql
+++ b/projects/extension/sql/idempotent/012-vectorizer-int.sql
@@ -116,6 +116,7 @@ begin
     into strict _pk_cols
     from pg_catalog.jsonb_to_recordset(source_pk) x(pknum int, attname name)
     ;
+
     select pg_catalog.format
     ( $sql$
     create table %I.%I
@@ -123,7 +124,7 @@ begin
     , %s
     , chunk_seq int not null
     , chunk text not null
-    , embedding @extschema:vector@.vector(%L) not null
+    , embedding @extschema:vector@.vector(%L) storage main not null
     , unique (%s, chunk_seq)
     , foreign key (%s) references %I.%I (%s) on delete cascade
     )
--- a/projects/extension/tests/vectorizer/test_vectorizer.py
+++ b/projects/extension/tests/vectorizer/test_vectorizer.py
@@ -117,7 +117,7 @@ TARGET_TABLE = """
  published      | timestamp with time zone |           | not null |                   | plain    |             |              | 
  chunk_seq      | integer                  |           | not null |                   | plain    |             |              | 
  chunk          | text                     |           | not null |                   | extended |             |              | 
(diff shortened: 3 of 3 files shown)


projects/extension/sql/incremental/003-vec-storage.sql
JamesGuthrie, line 1: I have two concerns:
1. Table rewrite
2. Will we be able to fit rows?

1. Is addressed by https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-DESC-SET-STORAGE, namely:
> Note that ALTER TABLE ... SET STORAGE doesn't itself change anything in the table; it just sets the strategy to be pursued during future table updates.

2. By setting the storage type to `plain`, we constrain the row size to a max of 8160 bytes. This means that users will be unable to store vectors of dimension >= ~2000. Are we sure that we want this limitation?

JamesGuthrie: I see that the documentation for `main` says:

> MAIN allows compression but not out-of-line storage. (Actually, out-of-line storage will still be performed for such columns, but only as a last resort when there is no other way to make the row small enough to fit on a page.)

We don't really want compression (I don't think), but it would be nice to get opt-in out-of-line storage.

JamesGuthrie: 1. Yes, my concern was: "would this trigger a table rewrite (and cause an outage)". If we do keep this change, and we do expect users to run a `VACUUM FULL` to get benefits from it, I would suggest that we explicitly mention that in the commit message.

jgpruitt:
```sql
test=# create table foo (id int, vec vector(9000) storage plain);
CREATE TABLE
test=# insert into foo (id, vec) select 1, (select array_agg(random())::vector(9000) from generate_series(1, 9000) x);
ERROR:  row is too big: size 36040, maximum size 8160
test=# create table boo (id int, vec vector(2000) storage plain);
CREATE TABLE
test=# insert into boo (id, vec) select 1, (select array_agg(random())::vector(2000) from generate_series(1, 2000) x);
INSERT 0 1
test=# create table roo (id int, vec vector(9000));
CREATE TABLE
test=# insert into roo (id, vec) select 1, (select array_agg(random())::vector(9000) from generate_series(1, 9000) x);
INSERT 0 1
```

jgpruitt: Maybe we should only set storage to plain if the number of dimensions is 2000 or less?

projects/extension/sql/incremental/003-vec-storage.sql
```diff
@@ -0,0 +1,30 @@
...
+        , v.target_schema
+        , v.target_table
+        )
+        from ai.vectorizer v
+        inner join pg_catalog.pg_class k on (k.relname = v.target_table)
+        inner join pg_catalog.pg_namespace n on (k.relnamespace = n.oid and n.nspname = v.target_schema)
+        inner join pg_catalog.pg_attribute a on (k.oid = a.attrelid)
+        where a.attname = 'embedding'
+        and a.attstorage != 'p' -- not plain
+        and a.atttypmod <= 2000
```
jgpruitt, line 21: this migration only happens if the vector dimensions are less than 2001 and storage isn't already plain

JamesGuthrie (approved):
This seems like a sane compromise.

For completeness: all of the other fields in the table must be < 160 bytes for the row to fit. Currently we have:
uuid (16 bytes), jsonb (unknown length, for the foreign key data), int (4 bytes), text (unknown length, for the chunk data), vector

Both the jsonb and the chunk will be TOASTed if necessary, so as long as we don't add a large number of columns to this table it should be fine.

JamesGuthrie:
I forgot to mention before, I would like a second opinion from @cevian.

jgpruitt:
Primary keys from the source table are duplicated 1 for 1 into the target table so we can have a proper foreign key constraint. We don't use jsonb for this; they are independent columns.

So, if the source table has multiple columns of large enough size, we could still have an issue with dimensions <= 2000.

jgpruitt:
I could attempt to sum the positive `typlen`s from `pg_type` for the primary key columns. If they are below a constant, then we are still okay to assign `plain` storage, otherwise don't.

jgpruitt:
Okay, I've wrestled with this for days. Trying to determine when it's safe to set the vector storage to `plain` is fragile. We decided to set it to `main` instead.

cevian (approved):

PR #200, opened 2024-11-06
docs: update docker-compose command

bytrangle:
`docker-compose` only works on Compose v1.
For Compose v2, which was released 4 years ago, the command should be `docker compose`.


JamesGuthrie:
@bytrangle thanks for this - we need you to sign our CLA in order to accept the PR though.

bytrangle:
Already signed. Thanks for the prompt reply.

JamesGuthrie (approved):

PR #202, opened 2024-11-07
fix: make vectorizer worker poll for new vectorizers

JamesGuthrie:
The vectorizer worker (running in continuous mode) now checks for the presence of new vectorizers on each pass through its work loop. This ensures that it processes vectorizers which are added after the worker starts.

### DIFF
--- a/projects/pgai/pgai/cli.py
+++ b/projects/pgai/pgai/cli.py
@@ -52,22 +52,26 @@ def get_pgai_version(cur: psycopg.Cursor) -> str | None:
 
 
 def get_vectorizer_ids(
-    cur: psycopg.Cursor, vectorizer_ids: Sequence[int] | None = None
+    db_url: str, vectorizer_ids: Sequence[int] | None = None
 ) -> list[int]:
-    valid_vectorizer_ids: list[int] = []
-    if vectorizer_ids is None or len(vectorizer_ids) == 0:
-        cur.execute("select id from ai.vectorizer")
-    else:
-        cur.execute(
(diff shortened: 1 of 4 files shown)


projects/pgai/pgai/cli.py
```diff
@@ -219,13 +223,23 @@ def vectorizer_worker(
...
+                f"invalid vectorizers, wanted: {list(vectorizer_ids)}, got: {valid_vectorizer_ids}"  # noqa: E501 (line too long)
+            )
+            sys.exit(1)
 
     while True:
-        for vectorizer_id in vectorizer_ids:
+        if dynamic_mode:
+            valid_vectorizer_ids = get_vectorizer_ids(db_url, vectorizer_ids)
+            if len(valid_vectorizer_ids) == 0:
+                log.warning("no vectorizers found")
```
Askir, line 240: This is nice, because it also stops it from crashing when there is no vectorizer defined yet.

projects/pgai/tests/vectorizer/conftest.py
```diff
@@ -46,7 +46,7 @@ def vcr_():
     )
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="class")
```
Askir, line 49: Took me a bit to understand this change, but I suppose the existing CLI tests contained a DB reset which is why we don't need a new container for each one.

I'm not so sure about the class grouping. I feel like we should either make the fixture sessions scoped and reset the DB every time. Or always create a new container.
Now it's in a bit of a weird mixed state, but well it works.

PR #204, opened 2024-11-07
chore: test cli not lambda handler

Askir:
https://linear.app/timescale/issue/AI-83/vectorizer-tests-target-the-lambda-handler-and-not-the-cli

Also removing the lambda handler from the code base in the process.
I honestly think the fixtures are trying to be a bit too generic and it ends up being a bit complicated to read that way but I didn't really change that model for now.


projects/pgai/tests/vectorizer/test_vectorizer_cli.py
```diff
@@ -0,0 +1,420 @@
...
+
+
+@pytest.fixture
+def cli_db(
+    cli_postgres_container: PostgresContainer,
+) -> Generator[dict[str, Any], None, None]:
+    """Creates a test database with pgai installed"""
+    role = "tsdbquerier"
+    password = "my-password"
+    db_host = cli_postgres_container._docker.host()  # type: ignore
```
JamesGuthrie, line 31: Do you know if this is a docker for mac workaround, and if it's necessary? I'm not sure it is.

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
```diff
@@ -0,0 +1,420 @@
...
+        yield postgres
+
+
+@pytest.fixture
+def cli_db(
+    cli_postgres_container: PostgresContainer,
+) -> Generator[dict[str, Any], None, None]:
+    """Creates a test database with pgai installed"""
+    role = "tsdbquerier"
+    password = "my-password"
```
JamesGuthrie, line 30: These could be removed with the change above

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
```diff
@@ -0,0 +1,420 @@
...
+            "event_db_config": {
+                "host": db_host,
+                "port": int(
+                    cli_postgres_container.get_exposed_port(cli_postgres_container.port)
+                ),
+                "db_name": cli_postgres_container.dbname,
+                "ssl_mode": "disable",
+                "role": role,
+                "password": password,
+            },
```
JamesGuthrie, line 61: The whole `event_db_config` field could be dropped. It's only used in `cli_db_url`, which could use `cli_db["container"].get_connection_url()`.

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
```diff
@@ -0,0 +1,420 @@
...
+        cli_postgres_container.get_connection_url(host=db_host),
+        autocommit=True,
+    ) as conn:
+        # Setup test user
+        conn.execute(sql.SQL("DROP USER IF EXISTS {}").format(sql.Identifier(role)))
+        conn.execute(
+            sql.SQL("CREATE USER {} WITH SUPERUSER PASSWORD {}").format(
+                sql.Identifier(role), sql.Literal(password)
+            )
+        )
```
JamesGuthrie, line 44: Not necessary if we create the user with the container.

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
```diff
@@ -0,0 +1,420 @@
...
+                "password": password,
+            },
+        }
+
+
+@pytest.fixture
+def cli_db_url(cli_db: dict[str, Any]) -> str:
+    """Constructs database URL from the cli_db fixture"""
+    config = cli_db["event_db_config"]
+    return f"postgres://{config['role']}:{config['password']}@{config['host']}:{config['port']}/{config['db_name']}?sslmode={config['ssl_mode']}"
```
JamesGuthrie, line 69: As mentioned above, this could become:
```suggestion
    return cli_db["container"].get_connection_url()
```

Or (if we want to keep the custom "host" thing which I don't understand):

```suggestion
    container = cli_db["container"]
    return container.get_connection_url(container.get_docker_client().host())
```

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
```diff
@@ -0,0 +1,420 @@
...
+    with cli_db["conn"].cursor(row_factory=dict_row) as cur:
+        # Cleanup from previous runs
+        cur.execute("SELECT id FROM ai.vectorizer")
+        for row in cur.fetchall():
+            cur.execute("SELECT ai.drop_vectorizer(%s)", (row["id"],))
+
+        # Drop tables if they exist
+        cur.execute("DROP VIEW IF EXISTS blog_embedding")
+        cur.execute("DROP TABLE IF EXISTS blog_embedding_store")
+        cur.execute("DROP TABLE IF EXISTS blog")
```
JamesGuthrie, line 98: I wish we didn't have to manually clean up. One idea would be to create a new database for each test, so we know that we're dealing with a fresh db. We can do it as a small follow-up PR though.

JamesGuthrie (approved):

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
alejandrodnm: We need that for running the tests from the the github action runners containers.

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
```diff
@@ -0,0 +1,420 @@
...
+           INSERT INTO
+           blog_embedding_store(embedding_uuid, id, chunk_seq, chunk, embedding)
+           VALUES (gen_random_uuid(), 1, 1, 'post_1',
+            array_fill(0, ARRAY[1536])::vector)
+        """)
+    monkeypatch.setenv("OPENAI_API_KEY", "test-key")
+
+    # When running the worker with cassette matching original test params
+    cassette = (
+        f"openai-character_text_splitter-chunk_value-"
```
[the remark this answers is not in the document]
Askir: Yes I also noticed the original fixture seemed to be configured to basically be the same as the character text splitter right now. But agree, we should add one for actual recursive text splitting.

PR #206, opened 2024-11-08
chore: run pgai tests against extension from source

alejandrodnm:
Change pgai testcontainer to build the extension and copy the extension code to a timescaledb-ha container.

This will make CI run the pgai python library against the latest changes made to the extension, and surface breaking changes sooner.

### DIFF
--- a/projects/extension/Dockerfile
+++ b/projects/extension/Dockerfile
@@ -1,11 +1,31 @@
 # syntax=docker/dockerfile:1.3-labs
-ARG PG_MAJOR
+ARG PG_MAJOR=16
 ARG TIMESCALEDB_IMAGE=timescale/timescaledb-ha:pg${PG_MAJOR}
 FROM ${TIMESCALEDB_IMAGE} as timescaledb
 
-ARG PG_MAJOR
-FROM postgres:${PG_MAJOR}
+FROM postgres:${PG_MAJOR} AS build-latest
+ENV PG_MAJOR=${PG_MAJOR}
+ENV PIP_BREAK_SYSTEM_PACKAGES=1
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends make python3-pip git && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+WORKDIR /pgai
+COPY requirements.txt .
+COPY ai/__init__.py ai/
+RUN version=$(grep -oP '(?<=__version__ = ")[^"]*' ai/__init__.py) && \
+    mkdir -p /usr/local/lib/pgai/$version && \
+    pip3 install -v --no-deps --compile -t /usr/local/lib/pgai/$version -r requirements.txt
+COPY . .
+RUN make install-py build-sql install-sql
 
+FROM timescaledb as pgai-test-db
+ARG PG_MAJOR=16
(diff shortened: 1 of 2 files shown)


projects/extension/Dockerfile
jgpruitt, line 1: This doesn't work for versions 0.1.0, 0.2.0, and 0.3.0 of the extension. Maybe that doesn't matter though for the pgai library.

projects/extension/Dockerfile
```diff
@@ -1,11 +1,31 @@
...
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends make python3-pip git && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+WORKDIR /pgai
+COPY requirements.txt .
+COPY ai/__init__.py ai/
+RUN version=$(grep -oP '(?<=__version__ = ")[^"]*' ai/__init__.py) && \
+    mkdir -p /usr/local/lib/pgai/$version && \
+    pip3 install -v --no-deps --compile -t /usr/local/lib/pgai/$version -r requirements.txt
```
jgpruitt, line 18: why are we doing this? `make install` does this already

projects/extension/Dockerfile
```diff
@@ -1,11 +1,31 @@
...
+COPY . .
+RUN make install-py build-sql install-sql
```
jgpruitt, line 20: why not just `make build-install`??

projects/extension/Dockerfile
```diff
@@ -1,11 +1,31 @@
...
 
+FROM timescaledb as pgai-test-db
+ARG PG_MAJOR=16
+COPY --from=build-latest /usr/share/postgresql/${PG_MAJOR}/extension/ai--*.sql /usr/share/postgresql/${PG_MAJOR}/extension/
+COPY --from=build-latest /usr/share/postgresql/${PG_MAJOR}/extension/ai.control /usr/share/postgresql/${PG_MAJOR}/extension/ai.control
+COPY --from=build-latest /usr/local/lib/pgai/ /usr/local/lib/pgai/
```
jgpruitt, line 26: for versions 0.1.0, 0.2.0, and 0.3.0, the dependencies are installed in the system path

jgpruitt (changes requested):
very in favor of the pgai tests using the latest extension code! 🎉

projects/extension/Dockerfile
alejandrodnm: Because `build_install` also calls `install_prior_py`, and I don't need to install those deps.

projects/extension/Dockerfile
alejandrodnm: To cache de dependencies in a Docker layer. If we change the extension code we don't need to re-install the dependencies.

projects/extension/Dockerfile
alejandrodnm: The pgai tests are not parametrize on the extension version, so it'll only use the default version. We'd need those version if we parametrize the extension versions. 

I think if we do that, instead of doing it in the code, we can do it in the ci matrix.

projects/extension/Dockerfile
jgpruitt: ah. yeah. bummer

jgpruitt (approved):

PR #208, opened 2024-11-08
feat: SQLAlchemy and alembic integration

Askir:
Adds Python integration for pgai that allows defining and managing vectorizers through SQLAlchemy models and Alembic migrations. The integration provides a declarative interface for vector embeddings and automated migration support.

```python
class BlogPost(DeclarativeBase):
    content_embeddings = VectorizerField(
        source_column="content",
        embedding=EmbeddingConfig(
            model="text-embedding-3-small",
            dimensions=768
        )
    )

# Semantic search query
results = (
    session.query(BlogPost.content_embeddings)
    .order_by(
        BlogPost.content_embeddings.embedding.cosine_distance(
            func.ai.openai_embed('text-embedding-3-small', 'search query')
        )
    )
    .limit(5)
)
```

Features:
- Declarative SQLAlchemy field for vectorizer configuration
- Alembic operations for creating and dropping vectorizer
- Support for migration generation for vectorizer changes

Known missing features:
- Downgrading of migrations
- Custom [sqlalchemy functions](https://docs.sqlalchemy.org/en/20/core/functions.html#function-api) for `ai.xxx` functions
- Comparison of existing vectorizer does not work yet! (Working on it)
- ~~Chunking configuration is not separated for character and recursive splitter (will fix)~~
- ~~Scheduling config only supports default and specific scheduling not no scheduling right now (will fix)~~

How to review:
1. Have a look at some tests and the docs to understand how it is supposed to work. If you don't like this or have questions, stop here and leave a comment of what to improve/change
2. Implementation wise:
- `sqlalchemy/__init__.py`: Contains the VectorizerField that defines sqlalchemy embedding models and registers the configuration for alembic.
- `alembic/autogenerate.py`: Contains the comparison functions as well as the python code generation.
- `alemic/operations.py`: Contains the SQL to be executed (and some alembic boilerplate).
- `configuration.py`: Contains dataclasses that map to all the sql parameters for create_vectorizer which act as the input for both the VectorizerField as well as the alembic create_vectorizer and drop_vectorizer operations.


projects/pgai/pgai/extensions/sqlalchemy/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __get__(self, obj, objtype=None):
+        """
+        When accessed on class (BlogPost.content_embeddings) -> return Embedding class
+        When accessed on instance (post.content_embeddings) -> return relationship data
+        """
+        if obj is None:  # Class access
+            return self._embedding_class
+        # Instance access - return relationship data through the relationship property
+        return getattr(obj, f"_{self.name}_relation")
```
Askir, line 50: This is a bit of a nasty hack.
I wanted to enable querying all Embeddings under the same key as you'd retrieve the subobjects if you already have a BlogPost.

This is not default sqlalchemy behaviour (because you'd usually have the embedding class defined separately). Alternatively I can just defined a property `Embeddings` on the parent model, that might end up working out better.

Or we disabled the relationship mapping by default then this attribute free to be used by the Embedding model I suppose?

Will think about it a bit more, happy about additional input.

projects/pgai/pgai/extensions/__init__.py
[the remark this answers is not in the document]
alejandrodnm: nit: what do you think about these class definitions to a non `__init__.py` file. I think putting them inside the `__init__.py` makes them less discoverable while browsing the codebase.

docs/python-integration.md
```diff
@@ -0,0 +1,302 @@
...
+        session.query(
+            BlogPost.content_embeddings,
+            BlogPost.title,
+            # Calculate and include distance in results
+            BlogPost.content_embeddings.embedding.cosine_distance(
+                func.ai.openai_embed(
+                    'text-embedding-3-small',
+                    query_text,
+                    text('dimensions => 768')
+                )
```
alejandrodnm, line 72: Can we reference this from the Vectorizer config? That way we can remove the cognitive load on the user. Maybe that's what @cevian was referring when talking about improving the dev experience.

docs/python-integration.md
```diff
@@ -0,0 +1,302 @@
...
+                text('dimensions => 768')
+            )
+        )
+    )
+)
+```
+
+### Model Relationships
+
+You can optionally create SQLAlchemy relationships between your model and its embeddings:
```
alejandrodnm, line 127: Would it be worth it to mention that we expose the relationship via the view that's created. I don't know much about sqlalchemy, so maybe this `relationship` way is better than querying the view directly.

alejandrodnm (commented):
Did a very brief pass over the files, left some comments. I'll try to do a more through review later.

projects/pgai/pyproject.toml
```diff
@@ -23,6 +23,9 @@ dependencies = [
     "typing_extensions>=4.0,<5.0",
     "datadog_lambda>=6.9,<7.0",
     "pytimeparse>=1.1,<2.0",
+    "sqlalchemy>=2.0.36",
+    "psycopg2>=2.9.10",
+    "alembic>=1.14.0",
```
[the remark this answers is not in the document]
Askir: yes for sure.
Question is a bit if the vectorizer itself should also live in its own extra? I feel like you wont need that in your core python e.g. fastapi application.
Almost an argument for a separate package but I think we want to make use of the pgai brand.

projects/pgai/pgai/extensions/__init__.py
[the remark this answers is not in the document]
Askir: I could also just have `pgai.sqlalchemy` and `pgai.alembic`? What do you think about that?
pgvector-python doesn't have an extension subpackage either.

docs/python-integration.md
Askir: This function is just plain sqlalchemy core right now, this works even without my extension code. We can add custom sql alchemy functions. e.g. so you don't have to wrap the dimension paramter in a `text`, I guess I could also run a subquery maybe in such a function. We can definitely provide a python helper that references the underlying vectorizer or loads it from the python config.

Mats idea was to do this on sql level though and then it's automatically also available in python so we safe some effort there.

docs/python-integration.md
Askir: I need to look up exactly how relationships work, but generally speaking it's just an eager join that enables the related objects to be loaded as a list on the parent.
Mat mentioned that he's, understandably a bit skeptical about automatic join, because you can run into N+1 problems depending on how it is configured, which is why I made it optional.

projects/pgai/pgai/alembic/operations.py
```diff
@@ -0,0 +1,248 @@
...
+        params.append(f"min_rows=>{config.min_rows}")
+    if config.opclass is not None:
+        params.append(f"opclass=>'{config.opclass}'")
+    if config.m is not None:
+        params.append(f"m=>{config.m}")
+    if config.ef_construction is not None:
+        params.append(f"ef_construction=>{config.ef_construction}")
+    if config.create_when_queue_empty is not None:
+        params.append(f"create_when_queue_empty=>{str(config.create_when_queue_empty).lower()}")
+    return f"ai.indexing_hnsw({', '.join(params)})"
```
Askir, line 162: One thing I am considering is to make the dataclass config objects smarter:
Currently there is multiple places where logic for each config is placed:
- rendering to SQL (here)
- rendering to python (in autogenerate.py)
- loading from the db state (in compare_vectorizers)
- Storing configuration in the metadata info dict on the vectorizer field

I think if I make these config objects "smarter" and add a few functions:
- `from_db_state`
- `to_sql()`
- `to_python` or just `__repr__`?

For each one. It should make keeping this in sync with our actual underlying db code a lot easier. Since you'll only have to touch one spot in the code and not understand how everything works?

Honestly I am a bit disappointed by alembic, i would have thought that at least the `to_python` part is covered by the library generically already.

Askir: Note: this actually kind of overlaps with the pydantic models we have in the vectorizer part of the library. Maybe there is also a way to re-use that config. But e.g. `IndexingConfig` does not exist there and needs to be loaded separately.

projects/pgai/pgai/alembic/autogenerate.py
```diff
@@ -0,0 +1,341 @@
...
+    for table_name, model_config in model_vectorizers.items():
+        if table_name not in existing_vectorizers:
+            # Create new vectorizer
+            upgrade_ops.ops.append(
+                _create_vectorizer_op(model_config)
+            )
+        else:
+            # Check for configuration changes
+            existing_config = existing_vectorizers[table_name]
+            if _config_has_changed(model_config, existing_config):
```
Askir, line 94: This change detection part doesn't work yet, no need to review.

projects/pgai/pgai/alembic/operations.py
cevian: I strongly encourage us to have one set of dataclass objects across all of our libraries. That way we have code reuse and only one place to add things. (and I expect we'll constantly be adding things). My mental model would be one set of dataclass objects (probably pydantic based) that's used (imported) by both vectorizer and these kind of integrations. So that could include things like `IndexingConfig` that's used by only some parts of the code and not others.

docs/python-integration.md
```diff
@@ -0,0 +1,299 @@
...
+from pgai.configuration import EmbeddingConfig
+
+class BlogPost(DeclarativeBase):
+    __tablename__ = "blog_posts"
+    
+    id = Column(Integer, primary_key=True)
+    title = Column(Text, nullable=False)
+    content = Column(Text, nullable=False)
+    
+    content_embeddings = VectorizerField(
```
cevian, line 22: I think this is a great interface. One nit `VectorizerField` or `Vectorizer`? In my mind this isn't really a field...

docs/python-integration.md
cevian: I agree we can improve this UX. But I also think we can do that in a separate PR

docs/python-integration.md
cevian: I wonder if we can set up a relationship but not have it do the eager join by default?

docs/python-integration.md
```diff
@@ -0,0 +1,299 @@
...
+def downgrade():
+    op.drop_vectorizer(1)
+```
+
+When changes are made to a model's vectorizer configuration, the autogeneration will create appropriate migration operations:
+
+```python
+def upgrade():
+    # Update vectorizer configuration
+    op.drop_vectorizer(1, drop_objects=True)
```
cevian, line 287: We should probably make it more clear this is a dangerous and destructive operations which requires re-embedding everything again and paying $$. Perhaps just a code comment explaining this is sufficient.

projects/pgai/pgai/alembic/autogenerate.py
```diff
@@ -0,0 +1,341 @@
...
+        view_schema=model_config.get("view_schema"),
+        view_name=model_config.get("view_name"),
+        queue_schema=model_config.get("queue_schema"),
+        queue_table=model_config.get("queue_table"),
+        grant_to=model_config.get("grant_to"),
+        enqueue_existing=model_config.get("enqueue_existing", True),
+    )
+    
+
+def _config_has_changed(
```
cevian, line 147: I wonder if this can be as easy as rendering the two configs as create_vectorizer calls and doing a string diff.

cevian (commented):
I think this is great. Straightforward and helpful approach. Left a few comments. The most important one is about trying to get common, reusable data classes between all of our components

projects/pgai/pgai/sqlalchemy/__init__.py
```diff
@@ -0,0 +1,175 @@
...
+
+# Type variable for the parent model
+T = TypeVar("T", bound=DeclarativeBase)
+
+
+class EmbeddingModel(DeclarativeBase, Generic[T]):
+    """Base type for embedding models with required attributes"""
+
+    embedding_uuid: Mapped[str]
+    id: Mapped[int]
```
jgpruitt, line 26: We aren't constrained to an integer `id` foreign key. How should we handle that? Source tables may have compound primary keys of various types

jgpruitt (changes requested):
Only got partway through. Will try to look at the rest after lunch.

projects/pgai/pgai/configuration.py
```diff
@@ -0,0 +1,61 @@
...
+@dataclass
+class EmbeddingConfig:
+    model: str
+    dimensions: int
+    chat_user: str | None = None
+    api_key_name: str | None = None
+
+    
+@dataclass
+class ChunkingConfig:
```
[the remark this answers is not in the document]
Askir: I'm only using one right now, because the recursive character text splitter is essentially the same as the character one if you only provide one splitting point right?
Maybe still good to support both for completeness.

projects/pgai/pgai/sqlalchemy/__init__.py
Askir: Oh you're right. I'll try make this a generic, but I'm a bit scared of breaking strict typing. I don't want the embedding class to have a possibly infinite amount of Any fields.

projects/pgai/pgai/alembic/operations.py
jgpruitt: It might actually be nice to have access to them in the extension too. I could use the pydantic models to validate the jsonb representations. But I don't know how cumbersome it would be to do this.

projects/pgai/pgai/configuration.py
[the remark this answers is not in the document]
jgpruitt: yeah, I'm also just wondering what will happen from a naming convention when we add a third chunking strategy

Askir:
I made the autogen work but have now separated this rather large PR in 3 smaller ones:
#265 #266 and #267

Closing this PR.

PR #214, opened 2024-11-12
fix: handle dropped columns when creating vectorizers

jgpruitt:
Currently, if the source table has had a column dropped, `create_vectorizer` will fail. This PR fixes this bug.


feikesteenbergen (approved):
🚀

adolsalamanca (approved):

PR #216, opened 2024-11-12
perf!: don't do an exact count by default in vectorizer_queue_pending

jgpruitt:
Previously, we counted every row in the queue in `ai.vectorizer_queue_pending`, which could be very slow for large queues. Now, we limit 10,001 by default and add an `exact_count` bool parameter, which can be used to get an exact count if desired.

### DIFF
--- a/docs/vectorizer-api-reference.md
+++ b/docs/vectorizer-api-reference.md
@@ -1014,6 +1014,19 @@ Return the number of pending items for the vectorizer with ID 1:
   SELECT ai.vectorizer_queue_pending(1);
   ```
 
+A queue with a very large number of items may be slow to count. The optional 
+`exact_count` parameter is defaulted to false. When false, the count is limited.
+An exact count is returned if the queue has 10,000 or fewer items, and returns
+9223372036854775807 (the max bigint value) if there are greater than 10,000 
+items.
+
+To get an exact count, regardless of queue size, set the optional parameter to
+`true` like this:
+
+  ```sql
+  SELECT ai.vectorizer_queue_pending(1, exact_count=>true);
+  ```
+
 #### Parameters
 
 `ai.vectorizer_queue_pending function` takes the following parameters:
@@ -1021,6 +1034,7 @@ Return the number of pending items for the vectorizer with ID 1:
 |Name| Type | Default | Required | Description |
 |-|------|-|-|-|
 |vectorizer_id| int  | -|✔|The identifier of the vectorizer you want to check|
+|exact_count| bool | false |✖|If true, return exact count. If false, capped at 10,000|
 
 
 ### Returns
--- a/projects/extension/sql/idempotent/013-vectorizer-api.sql
+++ b/projects/extension/sql/idempotent/013-vectorizer-api.sql
@@ -508,7 +508,10 @@ set search_path to pg_catalog, pg_temp
 
 -------------------------------------------------------------------------------
 -- vectorizer_queue_pending
-create or replace function ai.vectorizer_queue_pending(vectorizer_id int) returns bigint
+create or replace function ai.vectorizer_queue_pending
+( vectorizer_id int
+, exact_count boolean default false
+) returns bigint
 as $func$
 declare
     _queue_schema name;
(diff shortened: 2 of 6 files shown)


smoya:
What is the implication of such a change? Can this be considered as a breaking change for `vectorizer_queue_pending` in terms of functionality (not in function signature)?

jgpruitt:
I went back and forth on whether or not its a breaking change. I guess, technically, it is since you'd need to change your code to get the original behavior.

jgpruitt:
Added a ! to the commit message

smoya:
We would need to update the docs, for instance https://github.com/timescale/pgai/blob/main/docs/vectorizer-api-reference.md#aivectorizer_queue_pending-function

smoya:
> Added a ! to the commit message

You should change the PR title as it will be used as commit message in this case.

jgpruitt:
ew. ok. done

docs/vectorizer-api-reference.md
```diff
@@ -1014,13 +1014,21 @@ Return the number of pending items for the vectorizer with ID 1:
...
+  ```
+
 #### Parameters
 
 `ai.vectorizer_queue_pending function` takes the following parameters:
 
 |Name| Type | Default | Required | Description |
 |-|------|-|-|-|
 |vectorizer_id| int  | -|✔|The identifier of the vectorizer you want to check|
+|exact_count| bool | false |✖|If true, return exact count. If false, cap at 1000|
```
smoya, line 1031: Nit: not sure if `cap` is a short hand for `capped` but for me, a non-native speaker, `cap` can stand for other things.

```suggestion
|exact_count| bool | false |✖|If true, return exact count. If false, capped at 1000|
```

projects/extension/sql/idempotent/013-vectorizer-api.sql
```diff
@@ -523,11 +526,19 @@ begin
...
-    ;
+    if exact_count then
+        select format
+        ( $sql$select count(1) from %I.%I$sql$
+        , _queue_schema, _queue_table
+        ) into strict _sql
+        ;
+    else
+        select format
+        ( $sql$select count(*) from (select 1 from %I.%I limit 1000)$sql$
```
cevian, line 537: are we sure we want to return 1000 here and not say INT_MAX if the count is 1000? (btw I'd pbbl change it to 10000)

cevian: we could also change the return type to float and return +INF

cevian: we can also guestimate based on the relation size

jgpruitt: yep. or use -1 for "a lot"

jgpruitt: or `reltuples`

docs/vectorizer-api-reference.md
```diff
@@ -1014,13 +1014,27 @@ Return the number of pending items for the vectorizer with ID 1:
...
+  ```
+
 #### Parameters
 
 `ai.vectorizer_queue_pending function` takes the following parameters:
 
 |Name| Type | Default | Required | Description |
 |-|------|-|-|-|
 |vectorizer_id| int  | -|✔|The identifier of the vectorizer you want to check|
+|exact_count| bool | false |✖|If true, return exact count. If false, capped at 1000|
```
cevian, line 1037:
```suggestion
|exact_count| bool | false |✖|If true, return exact count. If false, capped at 10,000|
```

projects/extension/sql/incremental/004-vectorizer-queue-pending.sql
```diff
@@ -0,0 +1,6 @@
+
```
cevian, line 1: Think this file will need to be 005

cevian (approved):
LGTM except for 2 nits

projects/extension/sql/incremental/004-vectorizer-queue-pending.sql
jgpruitt: needed another rebase to see it

Issue #116, opened 2024-10-14
Missing OpenAI functionality


Tostino:
Just wanted to mention that I am looking at the `vectorizor` branch, as that seems to be the active development branch as far as I can tell.

The current OpenAI API is not really compliant with the python version of the API, restricting it's use. 

The extension is missing a bunch of arguments for the openai endpoints, and not all of the arguments that are there behave in-line with the API spec. Some of the missing arguments are needed for using proxy solutions or custom samplers on open source inference servers for example,
The return types for the embed endpoints also seems like it should just be json, because there are options to return the data in either base64 or floating point numbers for example.

I think the functions that provide a "simple" wrapper should be designed on top of the raw functions that return json rather than strict data types. Probably in a bit more structured a way than the current "simple" chat completion function. I haven't gotten to this yet. Having a full featured base set of functions was my priority first.

Either way, I put in a good bit of work into a branch, but cannot get the test suite to pass due to white-space issues with the `expected` files. I'm at my wits end with messing with the test suite considering I cannot run it locally because your build scripts aren't compatible with podman and that's what I use on my dev machine.

Comparison with branch: https://github.com/timescale/pgai/compare/vectorizer...Tostino:pgai:fix-openai-api?expand=1

The last major argument I haven't taken into consideration here is `timeout` which will need a little more thought because it needs to be supported at both the client level as well as the function level. 
Was planning on doing that after getting some feedback.

alejandrodnm:
Hey @Tostino, thanks for looking into the project.

> "simple" wrapper should be designed on top of the raw functions that return json rather than strict data types

This is something we might look post merge of the vectorizer branch. We've noticed some things from the openAI library as well, that might steer us in the direction of using http calls made by ourselves directly.

What tests are giving you issues? Can you share the logs. We want to improve devex regarding the repo, so any feedback is welcome.

We are merging this 2 projects together (vectorizer and extension) into one repository so things might be a little cumbersome while we iron out the wrinkles. Thanks.

Tostino:
> That's why there's been a little bit of radio silence on our part.

No problem at all.

> The library uses a lot of typing mechanism and casting that increase the CPU consumption under heavy load

The portion of the overhead time is spent creating the client for each and every request.
The actual `chat.completions.create` call (minus the inference server actually doing it's work) is relatively lightweight from what i've seen.

I had already planned on doing some work on caching the client between requests, because I know there is some serious overhead in setting that up for every call from a past project.

[...]

PR #219, opened 2024-11-13
OpenAI: Migrate to async client and enhance API support

Tostino:
Major changes:
- Migrate to async OpenAI client to support query cancellation and timeouts
- Add client caching using global dictionary (GD) to improve performance
- Migrate to using raw responses to minimize type conversions and improve performance
- Add comprehensive support for all OpenAI API parameters
- Add support for client create/destroy methods

Implementation details:
- Replace sync OpenAI client with AsyncOpenAI for better control flow
- Implement client caching in GD to reuse connections
- Add query cancellation support using asyncio
- Remove list_models and embed function implementations from openai.py to consolidate API handling
- Move functionality directly into the SQL functions for consistency
- Return raw API responses to minimize conversions
- Add complete OpenAI API parameter support across all functions
- Standardize parameter naming with leading underscore
- Update OpenAI and tiktoken package versions

Package updates:
- openai: 1.44.0 -> 1.51.2
- tiktoken: 0.7.0 -> 0.8.0

Breaking changes:
- Functions now return raw JSON responses instead of parsed objects
- Functions marked as parallel unsafe due to HTTP API constraints
- Parameter names now prefixed with underscore to reduce naming conflicts

Known issues:
- Inconsistent performance inside the plpython environment.
  - The first call to the endpoint is quick (3ms), and then every call after it is very delayed (40ms). Need to figure out what is happening here. I cannot reproduce outside of plpython.


projects/extension/sql/idempotent/001-openai.sql
```diff
@@ -3,12 +3,12 @@
 -- openai_tokenize
 -- encode text as tokens for a given model
 -- https://github.com/openai/tiktoken/blob/main/README.md
-create or replace function ai.openai_tokenize(model text, text_input text) returns int[]
+create or replace function ai.openai_tokenize(_model text, _text text) returns int[]
```
cevian, line 6: we do not use leading underscores in parameter names. They make it hard to call functions with `=>` syntax.  Even if it makes the developers job easier to use the underscore, it makes the user interface worse, and that's not a good tradeoff. This needs to be fixed all over.

projects/extension/ai/openai.py
```diff
@@ -17,58 +18,95 @@ def get_openai_api_key(plpy, api_key_name: Optional[str] = None) -> str:
...
+        "max_retries": max_retries,
+        "default_headers": default_headers,
+        "default_query": default_query,
+        "http_client": http_client,
+        "_strict_response_validation": _strict_response_validation
+    })
+
+    return openai.AsyncOpenAI(**client_kwargs)
+
+def get_or_create_client(plpy, GD: Dict[str, Any], api_key: str = None, api_key_name: str = None, base_url: str = None) -> Any:
```
cevian, line 60: Do you have any number showing that creating the client is expensive (and thus worth it to store in GD)?. Does this allow connection reuse or something? And if it's the latter then how/when do connections get closed? Is there a keepalive timeout.

Storing the client in GD seems like a good amount of complexity and I'd like to find out what we are gaining/loosing for it.

projects/extension/sql/idempotent/001-openai.sql
```diff
@@ -19,196 +19,391 @@ set search_path to pg_catalog, pg_temp
...
     return content
 $python$
 language plpython3u strict immutable parallel safe security invoker
 set search_path to pg_catalog, pg_temp
 ;
 
+-------------------------------------------------------------------------------
+-- openai_client_create
+-- create the client and store it in the global dictionary for the session
+CREATE OR REPLACE FUNCTION ai.openai_client_create(
```
cevian, line 37: What's the reason for exposing this functionality to the user? Seems like this should be an internal implementation detail.

projects/extension/sql/idempotent/001-openai.sql
```diff
@@ -19,196 +19,391 @@ set search_path to pg_catalog, pg_temp
...
+            'client': client,
+            'config': new_config
+        }
+
+$python$ LANGUAGE plpython3u VOLATILE SECURITY DEFINER PARALLEL UNSAFE;
+
+-------------------------------------------------------------------------------
+-- openai_client_destroy
+-- remove the client object stored in the global dictionary for the session
+CREATE OR REPLACE FUNCTION ai.openai_client_destroy() RETURNS void AS $python$
```
cevian, line 80: same question. This seems it should be internal and not exposed to the user.

projects/extension/sql/idempotent/001-openai.sql
```diff
@@ -19,196 +19,391 @@ set search_path to pg_catalog, pg_temp
...
+        return response.text
+
+    # Execute the API call with cancellation support
+    result = ai.openai.execute_with_cancellation(plpy, client, async_openai_call, **kwargs)
+
+    return result
 $python$
-language plpython3u volatile parallel safe security invoker
-set search_path to pg_catalog, pg_temp
+    language plpython3u volatile parallel unsafe security invoker
```
cevian, line 126: why are these functions parallel unsafe? My understanding is every parallel worker would get a different process and thus a different python interpreter. I don't think even GD is shared.

projects/extension/sql/idempotent/001-openai.sql
```diff
@@ -19,196 +19,391 @@ set search_path to pg_catalog, pg_temp
...
+        return response.text
+
+    # Execute the API call with cancellation support
+    result = ai.openai.execute_with_cancellation(plpy, client, async_openai_call, **kwargs)
+
+    return result
 $python$
-language plpython3u immutable parallel safe security invoker
-set search_path to pg_catalog, pg_temp
+    language plpython3u volatile parallel unsafe security invoker
```
cevian, line 182: why is this now volatile? That seems wrong. A call with the same parameters should always return the same embedding.

cevian (commented):
Thank you so much for this PR! I did a preliminary check and have a bunch of questions. I just want to understand the motivation/reasoning behind some decisions. Also we'll have to decide on the json vs vector return of some of these functions. I think you are right we'll need both sets of functions. Let me ask some of my colleagues about the naming conventions we want to use here.

projects/extension/sql/idempotent/001-openai.sql
Tostino: Got it, will make the change. This was an artifact from the fact that I was already building these functions prior to finding your project, and copy/paste.

projects/extension/ai/openai.py
Tostino: Yup, benchmarks here: https://github.com/timescale/pgai/issues/116#issuecomment-2421594920

Also note, that there is a known issue with the 2nd (and 3rd...etc) call to the client for the api has some extra 20ms delay that doesn't happen when I have this code running outside of a pl/python environment (noted in the thread above). I really should have mentioned that directly in the PR. will edit to mention that fact that it needs to be identified. Once that is fixed, the benchmark numbers should look much better.

Even with the above issue, this is still much faster, and lower CPU than the original implementation where we recreate the client.

Note specifically the CPU reduction. Recreating the client is heavy on CPU, I know this from past projects but the benchmarks also bare this out.

I believe the connection is closed after the request is complete, and the client becomes ready for the next call. If the request is cancelled early, we attempt to kill things gracefully.

projects/extension/sql/idempotent/001-openai.sql
Tostino: There are options when calling the client_create function that are unavailable as options to any of the specific endpoint calls. If you need to use them for your inference environment to work properly, there needs to be a way of configuring them.

projects/extension/sql/idempotent/001-openai.sql
Tostino: Ditto. Seemed wrong to have a create function without the corresponding cleanup. I'm sure in the majority of use cases this could be ignored.

projects/extension/sql/idempotent/001-openai.sql
Tostino: Sorry, another artifact from my prior implementation. I was thinking it was going to be unsafe even if each call created it's own client due to overloading the API unintentionally, but in retrospect, it's better to control that with the standard GUC for controlling parallelism.

projects/extension/sql/idempotent/001-openai.sql
Tostino: So, the embedding endpoint for OpenAI is not totally deterministic. You can get slightly different vectors back for the same input on the same endpoint with the same model.

I'd say that is more of a bug with their implementation though, and immutable would work and be _much_ cheaper in a lot of situations (with duplicated data).
Will change it back.

cevian:
@Tostino I believe we used `openai_user` before. Let's stick with that.

Tostino:
@cevian Alright, all changes made. Also noticed that when I rebased things I had accidentally committed changes to the ai--0.4.0.sql file, so I got that reverted.

This still has the performance problem we need to dig into before it's committed, but at least any of the other issues can be discussed and fixed in the meantime.

cevian:
@Tostino I am still not convinced we need `ai.openai_client_create(` as a public function. Can we just pass any client options we need as a jsonb to the other functions via a `client_extra_args` parameter?

Tostino:
I'm good with that solution.  Seems to solve the problem statement I was originally trying to solve. Will get it done tonight.

Tostino:
Well...some kid went and ripped out my neighborhoods internet interconnection wiring last night. Was slightly delayed.

Tested to make sure the client_extra_args were being passed through properly, and it seems to be with my initial "kick the tires" tests.

Tostino:
I should have a little time to try and figure out that 2nd run issue this week (or at least attempt, i'm not a Python dev so I am not used to the profiling tools in this space).

@cevian Is there anything else you see that needs attention at this point?

alejandrodnm:
Hey @Tostino is:

> The first call to the endpoint is quick (3ms), and then every call after it is very delayed (40ms). Need to figure out what is happening here. I cannot reproduce outside of plpython.

A problem with the current state as well? Or introduced by your PR?

If it's a current issue, could you open a github issue for it and we can tackle that on a separate PR.

adam-brusselback:
@alejandrodnm No, the current state is just a much slower overall call time every time (I believe it was roughly 25-30ms/call) and much higher CPU usage. Not a current issue, was introduced by the PR.

Sorry, holidays had me a bit busy. Will get back to this as soon as I can.

Tostino:
@alejandrodnm No, the current state is just a much slower overall call time every time (I believe it was roughly 25-30ms/call) and much higher CPU usage. Not a current issue, was introduced by the PR.

Sorry, holidays had me a bit busy. Will get back to this as soon as I can.

alejandrodnm:
@Tostino don't worry. Just wanted to see if we could support you better. You've put a lot of effort into this, and we really appreciate it.

Tostino:
Well, spent some time rebasing this branch this morning because I had a bit of time. It looks like recent changes have broken the build process again on my machine, so no progress made towards fixing whatever the issue is.

I've got a baby on the way in the next few days, so I likely won't get back to this now for a while. This PR is likely dead unless someone else wants to take it up.

alejandrodnm:
> Well, spent some time rebasing this branch this morning because I had a bit of time. It looks like recent changes have broken the build process again on my machine, so no progress made towards fixing whatever the issue is.
> 
> I've got a baby on the way in the next few days, so I likely won't get back to this now for a while. This PR is likely dead unless someone else wants to take it up.

Congrats on the baby 🙏🏼 sending all the good vibes to you so everything goes perfectly (I got 2 so I know how it's). 

I can try taking it up later, or maybe someone from the team will. Just one thing, could you leave a comment explaining what we'll need to do next?

Thanks, and congrats again.

Tostino:
So I spent more time today, and replaced the async implementation with an equivalent sync implementation. It made absolutely no difference in behavior between async and sync. The second call to the client (even when just called twice within the same function call) always has that extra ~45ms delay with both implementations.

When caching the client:
1. Initial client setup is slow (~20ms) and takes a lot of CPU
2. Using that client for a (local) call the first time is very quick (~1ms)
3. Further calls are slow (~45ms) but don't take much CPU

~6s/100 calls to openai_chat_completion 5% CPU per connection (e.g. can easily have 100+ connections to the server running inference and not kill performance)

When creating new client:
1. Client setup slow and takes a lot of CPU every call (~20ms)
2. Using that client for a (local) call the first time is very quick (~1ms)
3. Client setup slow and takes a lot of CPU every call (~20ms)
4. Using that client for a (local) call the first time is very quick (~1ms)
5. ...

~3s/100 calls to openai_chat_completion 100% CPU per connection (e.g. can only have NUM_CORES connections to the server before saturating CPU)

So when creating a new client even though each individual connection will finish the calls twice as fast, the server will be overloaded by just a handful of users calling these functions.

I hate that python performance is just so damn terrible to begin with that this is something we have to spend time optimizing.

With that said, I think the only real option is to stick with caching the client (just like it is now in this branch) because it doesn't knock over otherwise healthy servers when users decide to actually use the functions.
I'm sure someone can figure out what is going on with httpx and why it is introducing extra latency...but I sure failed to figure out a solution.

Tostino:
Uh, looks like I made a mistake in rebasing and pulled in some unexpected changes somehow. Gotta get that fixed right now.

alejandrodnm:
Hey @Tostino , hope everything went well with your child.

I was looking at the PR this week, not too deep. I closed and re-opened and that fixed the git diff stuff. 

I didn't get a good idea on how to tackle the client problem. Also, @cevian had some concerns and asked in the Postgres mailing list and didn't get a response.

https://www.postgresql.org/message-id/CADsUR0Aauim-HULPsks0PGAW6jJ%3D-VYfFOahvqGQxk6zBak7Vg%40mail.gmail.com

I think the problem might come from here:

https://github.com/openai/openai-python/issues/1596

What I was thinking on doing, is to split the PR into:

- Feature PR:
  - Migrate to using raw responses to minimize type conversions and improve performance
  - Add comprehensive support for all OpenAI API parameters
- Perf improvement PR:
  - Migrate to async OpenAI client to support query cancellation and timeouts
  - Add client caching using global dictionary (GD) to improve performance
  - Add support for client create/destroy methods
  
WDYT? Do you still want me to take over?

alejandrodnm:
The work on this PR is being superseded by smaller PRs that will be open against the repo. The work is going to be split into:

- Add support for extra request,header,body.
- Add function variants that return the raw response.
- Add support for client configuration arguments.
- Replace the sync client with the async client (cancellation support).

PR #220, opened 2024-11-13
feat: allow drop_vectorizer to optionally drop target and view

jgpruitt:
Adds an optional `drop_all` boolean parameter to `drop_vectorizer` to drop the target table and view.


projects/extension/tests/vectorizer/test_vectorizer.py
```diff
@@ -607,6 +607,135 @@ def test_drop_vectorizer():
...
+                and g.tgrelid = '{vectorizer.source_schema}.{vectorizer.source_table}'::regclass::oid
+                ;
+            """)
+            assert cur.rownumber is not None
+            pg_proc_oid = cur.fetchone()[0]
+
+            # drop the vectorizer with drop_all=>true
+            cur.execute(
+                "select ai.drop_vectorizer(%s, drop_all=>true)", (vectorizer_id,)
+            )
```
Askir, line 680: Is there also a test with `drop_all=>false` somewhere?

Askir (approved):
Looks good to me, I think the more common use-case would be to set `drop_all=>true` (I am doing this in migrations or when I messed up the config).

But I wouldn't make it the default as you might accidentally lose a lot of expensive embeddings. The only other option would be to check how many embeddings the store table has and if it's over a threshold default to false and raise a warning or something, but I think that's too much logic for a simple drop as well. So I'm happy with this.

projects/extension/tests/vectorizer/test_vectorizer.py
jgpruitt: Yes. The other drop vectorizer test uses the default of false

cevian (approved):
LGTM. Let's discuss dependency/event trigger stuff for a separate PR

PR #223, opened 2024-11-14
fix: make ai.secret_permissions dump/restore properly

jgpruitt:
Two rows are inserted into ai.secret_permissions on extension creation. If these rows are dumped, then on restore the extension creation inserts them AND the table restoration tries to insert them. This causes a constraint violation and prevents the table from restoring properly. This commit will add a filter to prevent the two rows from being dumped.


projects/extension/sql/incremental/007-secret-permissions-dump-filter.sql
```diff
@@ -0,0 +1,26 @@
+
```
cevian, line 1: lets add a comment in `002-secret_permissions.sql` that the config dump settings are changed in this file

cevian (approved):

projects/extension/sql/incremental/007-secret-permissions-dump-filter.sql
jgpruitt: it will generate a warning on upgrade since that file has already shipped and is now frozen. that okay with you?

cevian (approved):

PR #226, opened 2024-11-15
chore: a new database for each test

JamesGuthrie:
In 12c17809ec235d759e37eaa0898ea3274fea6319 we made tests less fragile by giving each test class its own fresh container, which also made the tests run a bit slower.

This change reuses the same container for all tests, but creates a new database for each test, achieving the same isolation, with low overhead.


alejandrodnm (approved):
Awesome.

PR #227, opened 2024-11-15
feat: add Ollama support to vectorizer

JamesGuthrie:


projects/pgai/pgai/vectorizer/embeddings.py
```diff
@@ -68,6 +89,92 @@ async def embed(
...
+        """
+
+    @abstractmethod
+    def _max_chunks_per_batch(self) -> int:
+        """
+        The maximum number of chunks that can be embedded per API call
+        :return: int: the max chunk count
+        """
+
+    async def _outer_embed(self, documents: Documents) -> list[EmbeddingVector]:
```
JamesGuthrie, line 107: I'm not really happy with this split between _outer_embed and _inner_embed, because it is IMHO hard to follow.

projects/pgai/pgai/vectorizer/embeddings.py
```diff
@@ -414,3 +464,85 @@ async def _encode(self, documents: list[str]) -> list[list[int]]:
     @cached_property
     def _encoder(self) -> tiktoken.Encoding:
         return tiktoken.encoding_for_model(self.model)
+
+
+@dataclass
+class OllamaOptions:
```
JamesGuthrie, line 470: I added this for completeness, these are the runtime parameters that ollama can take. We don't expose them in the SQL API yet, so I'm tempted to remove them.

projects/pgai/pgai/vectorizer/embeddings.py
```diff
@@ -414,3 +464,85 @@ async def _encode(self, documents: list[str]) -> list[list[int]]:
...
+        implementation (Literal["ollama"]): The literal identifier for this
+            implementation.
+        model (str): The name of the Ollama model used for embeddings.
+    """
+
+    implementation: Literal["ollama"]
+    model: str
+    truncate: bool = True
+    options: OllamaOptions | None = None
+    keep_alive: float | str | None = None
```
JamesGuthrie, line 504: We also don't expose this parameter in the SQL API yet (`ai.embedding_ollama()`), I'm not sure if we should. The `keep_alive` parameter basically tells ollama how long it should keep the model in memory after an API call. The default is 5 minutes.

projects/pgai/pgai/vectorizer/embeddings.py
jgpruitt: I *think these all apply to generation rather than embedding

projects/pgai/pgai/vectorizer/embeddings.py
jgpruitt: might be nice to have for performance reasons?

Doginal:
👍 Can't wait for Ollama embeddings! Any idea on a timeline to release? 

Thanks!

projects/pgai/pgai/cli.py
```diff
@@ -89,18 +89,26 @@ def get_vectorizer(db_url: str, vectorizer_id: int) -> Vectorizer | None:
...
+                    vectorizer_id=vectorizer_id,
+                )
+                return None
+            secrets: dict[str, str | None] = {api_key_name: api_key}
+            # The Ollama API doesn't need a key, so it doesn't support `set_api_key`
+            set_api_key = getattr(vectorizer.config.embedding, "set_api_key", None)
+            if callable(set_api_key):
+                set_api_key(secrets)
+            else:
+                log.error(
```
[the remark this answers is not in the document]
JamesGuthrie: We could remove the whole `else` branch here, it would only be hit by an incorrectly-set embedding configuration.

jgpruitt (approved):

JamesGuthrie:
@jgpruitt please take another look. I've added documentation, and in doing so noticed some changes that I wanted to make. I've pushed those changes in a separate commit on top of what you've already reviewed. I've purposefully left an incorrectly formatted commit message because I will squash that commit into its parent before merging.

projects/pgai/pgai/vectorizer/embeddings.py
```diff
@@ -414,3 +469,191 @@ async def _encode(self, documents: list[str]) -> list[list[int]]:
...
+    def _batcher(self) -> BatchApiCaller[StringDocument]:
+        return BatchApiCaller(self._max_chunks_per_batch(), self.call_embed_api)
+
+    @override
+    def _max_chunks_per_batch(self) -> int:
+        # Note: the chosen default is arbitrary - Ollama doesn't place a limit
+        return int(os.getenv("OLLAMA_MAX_CHUNKS_PER_BATCH", default="2048"))
+
+    async def call_embed_api(self, documents: str | list[str]) -> EmbeddingResponse:
+        response = await ollama.AsyncClient(host=self.base_url).embed(
```
smoya, line 580: Wondering if we should allow other client settings, with special emphasis in allowing custom headers. A use case could be for authenticating the request made to ollama's server (auth proxy in front of the server, etc).

projects/pgai/pgai/vectorizer/embeddings.py
```diff
@@ -414,3 +469,191 @@ async def _encode(self, documents: list[str]) -> list[list[int]]:
...
+            if "input length exceeds maximum context length" not in e.error:
+                raise e
+            # Note: We don't attempt tokenizing data sent to Ollama models
+            # because every model has its own tokenizer, and there's no way to
+            # programmatically determine the tokenizer. Without knowing the
+            # token length of a chunk, we can't filter the long ones out (like
+            # we do in the OpenAI integration). Instead, we retry each chunk
+            # individually. It's not ideal, but not too bad because it's
+            # (probably) a local API.
+            context_length = await self._context_length()
```
smoya, line 567: Just to understand, Is the `truncate=true` preventing this? Or we can still fall into this anyway?

projects/pgai/pgai/cli.py
```diff
@@ -89,18 +89,26 @@ def get_vectorizer(db_url: str, vectorizer_id: int) -> Vectorizer | None:
...
-            log.error(
-                "API key not found",
-                api_key_name=api_key_name,
-                vectorizer_id=vectorizer_id,
-            )
-            return None
-        secrets: dict[str, str | None] = {api_key_name: api_key}
         vectorizer = Vectorizer(**vectorizer)
-        vectorizer.config.embedding.set_api_key(secrets)
+        if "api_key_name" in embedding:
```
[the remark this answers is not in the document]
JamesGuthrie: I have a comment expressing the same on line 104. Should I move that one up here instead?

projects/pgai/pgai/vectorizer/embeddings.py
JamesGuthrie: Yes, `truncate=True` will prevent us getting the error "input length exceeds maximum context length".

docs/vectorizer-worker.md
```diff
@@ -105,15 +106,19 @@ On your local machine:
    name: pgai
    services:
      vectorizer-worker:
-       image: timescale/pgai-vectorizer-worker:0.1.0rc4
+       image: timescale/pgai-vectorizer-worker:0.1.0
        environment:
          PGAI_VECTORIZER_WORKER_DB_URL: postgres://<username>:<password>@<host>:<port>/<database-name>
          OPENAI_API_KEY: <your-api-key>
+         OLLAMA_HOST: <ollama-host-url>
+         OLLAMA_MAX_CHUNKS_PER_BATCH: <ollama-max-chunks-per-batch>
```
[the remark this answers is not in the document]
JamesGuthrie: done

docs/vectorizer-api-reference.md
```diff
@@ -287,6 +296,53 @@ The function takes several parameters to customize the OpenAI embedding configur
...
+|------------|---------|---------|----------|---------------------------------------------------------------------------------------------------------------------------------------------------------|
+| model      | text    | -       | ✔        | Specify the name of the Ollama model to use. For example, `nomic-embed-text`. Note: the model must already be available (pulled) in your Ollama server. |
+| dimensions | int     | -       | ✔        | Define the number of dimensions for the embedding vectors. This should match the output dimensions of the chosen model.                                 |
+| base_url   | text    | -       | ✖        | Set the base_url of the Ollama API. Note: no default configured here to allow configuration of the vectorizer worker through `OLLAMA_HOST` env var.     |
+| truncate   | boolean | true    | ✖        | Truncates the end of each input to fit within the chosen model's context length. Embedding fails if set to false and the context length is exceeded.    |
```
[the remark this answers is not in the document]
JamesGuthrie: done

docs/vectorizer-worker.md
```diff
@@ -151,6 +156,8 @@ On your local machine:
        ```bash
        export PGAI_VECTORIZER_WORKER_DB_URL="postgres://<user>:<password>@<host>:<port>/<dbname>"
        export OPENAI_API_KEY="Your OpenAI API key"
+       export OLLAMA_HOST="ollama host url"
+       export OLLAMA_MAX_CHUNKS_PER_BATCH=2048
```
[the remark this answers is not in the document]
JamesGuthrie: done

projects/extension/sql/idempotent/008-embedding.sql
```diff
@@ -41,6 +59,8 @@ begin
     case _implementation
         when 'openai' then
             -- ok
+        when 'ollama' then
+            -- ok
```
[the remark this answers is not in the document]
JamesGuthrie: What kind of error are we trying to catch here? The `ai.embedding_*` functions define mandatory arguments, so it will ensure that we have _some_ values. The configuration function may not turn those into a json which actually contains the right fields, but I'm not sure that we should be checking that here.

projects/extension/sql/idempotent/008-embedding.sql
```diff
@@ -21,6 +21,32 @@ $func$ language sql immutable security invoker
...
 
+-------------------------------------------------------------------------------
+-- embedding_ollama
+create or replace function ai.embedding_ollama
+( model text
+, dimensions int
+, base_url text default null
+, truncate boolean default true
+, options jsonb default null
+, keep_alive text default null
```
JamesGuthrie, line 32: I just saw that in the `ai.ollama_embed` function, we use `float8` for `keep_alive`: https://github.com/timescale/pgai/blob/eb697883c30634839d1fc4e4b3efb98d59d5283d/projects/extension/sql/idempotent/002-ollama.sql#L102. The python client supports both float8 and text. I chose text because I feel like duration specifiers like `10m` are easier to use than `600.0`. I wonder what we should do about aligning these. @jgpruitt WDYT?

projects/extension/sql/idempotent/008-embedding.sql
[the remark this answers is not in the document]
jgpruitt: It may be overkill, but since these values are jsonb, someone could try to hack together a config without using the appropriate configuration functions.

projects/pgai/pgai/cli.py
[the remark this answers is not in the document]
smoya: I would say yes; I missed it when reviewing this.

projects/pgai/pgai/vectorizer/embeddings.py
smoya: Worth to add a comment here 🙏 feel free to discard otherwise.

smoya (approved):
Great job @JamesGuthrie 👏 👏 . 

Giving my +1. Leftovers are only DX related nits.

projects/extension/sql/idempotent/008-embedding.sql
jgpruitt: I'm ambivalent. I'm okay with your aligning them if you want.

jgpruitt (approved):

projects/pgai/pgai/cli.py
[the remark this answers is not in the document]
JamesGuthrie: done

projects/extension/sql/idempotent/008-embedding.sql
JamesGuthrie: I considered adding a second overload with `keep_alive` having type `numeric`, but that doesn't work because then postgres can't determine which overload to use if you don't pass the optional parameter (e.g. `ai.embedding_ollama('test-model', 1234);`).

I'm going to stick to this API, as it's the more general one (you can also call it with `keep_alive => 600.0::text`).

PR #228, opened 2024-11-15
chore: use snapshot tests

JamesGuthrie:


jgpruitt (changes requested):
I'm not crazy about this change. I like that it reduces code, however it feels like a worse experience.

There's no typing or even an import statement to clue somebody into what `snapshot` is or how it's supposed to work. You have to know to go look at `syrupy` and learn how it works. Previously, there was more code, but it was straightforward.

The actual output isn't stored anywhere on disk where I can look at it easily with whatever tool I wish.

Correct me if I'm wrong, but `just test-update` is all-or-nothing. When dealing with changes that trigger these test failures, I prefer to run each in isolation and carefully fix each one at a time. There's less mental load that way.

PR #229, opened 2024-11-15
chore: only run CI on push

JamesGuthrie:
We were specifying that CI should be run on push, and on pull_request, this end up running jobs twice.

Push is the more useful of the two, because it means that you don't have to wait for a PR to be opened for the build to run against the changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request, workflow_dispatch]
+on: [pull_request, workflow_dispatch]
 permissions:
   contents: read
 


jgpruitt (changes requested):
There is one instance in which having the `pull_request` CI run is very helpful, and I've already run into it. If someone has already merged a new SQL file `005-bob.sql` and you are trying to add `005-fred.sql`, the `push` event won't catch the issue, but the `pull_request` event WILL catch the issue.

Thoughts?

JamesGuthrie:
I think we've been through this discussion before with Promscale. IIRC the only way to reliably make this work is to ensure that the type of situation you're talking about results in merge conflicts.

jgpruitt:
I can't think of a way to reliable cause a merge conflict for this situation. Can you?

If not, then I think we need to run CI on main on a daily schedule or something to catch these issues.

Askir:
I'd argue that the PR job is the more important one. I often push changes, that I know are broken but I want people to be able to see or reference (e.g. draft PRs), so running the on push job there is pretty redundant.

And as long as we require a branch to be up to date with main before merging this also ensures that main is never broken.

Askir:
btw if we disable the PR trigger I think PRs from forks simply don't run trough CI at all.

JamesGuthrie:
> I can't think of a way to reliable cause a merge conflict for this situation. Can you?

The only way to reliably cause merge conflicts is to serialize the migration script file names to a single file. In your example of `005-bob.sql` and `005-fred.sql`, if the "bob" change was merged first then the "conflict" file on main would contain:

```
// conflict file
001-sue.sql
002-me.sql
003-not.sql
004-dear.sql
005-bob.sql
```

While the "conflict" file in the "fred" branch would contain:
```
// conflict file
001-sue.sql
002-me.sql
003-not.sql
004-dear.sql
005-fred.sql
```

JamesGuthrie:
> I'd argue that the PR job is the more important one. I often push changes, that I know are broken but I want people to be able to see or reference (e.g. draft PRs), so running the on push job there is pretty redundant.
> 
> And as long as we require a branch to be up to date with main before merging this also ensures that main is never broken.


btw if we disable the PR trigger I think PRs from forks simply don't run trough CI at all.

alejandrodnm (approved):

jgpruitt (approved):

PR #238, opened 2024-11-20
build: use deb pkgs for timescaledb and pgvectorscale

jgpruitt:
This removes our dependency on the timescale/timescaledb-ha docker image. The debian packages are released first, and this allows us to test new versions sooner.

### DIFF
--- a/projects/extension/Dockerfile
+++ b/projects/extension/Dockerfile
@@ -1,33 +1,11 @@
 # syntax=docker/dockerfile:1.3-labs
 ARG PG_MAJOR=16
-ARG TIMESCALEDB_IMAGE=timescale/timescaledb-ha:pg${PG_MAJOR}
-FROM ${TIMESCALEDB_IMAGE} as timescaledb
+ARG PGVECTORSCALE_VERSION=0.5.0
 
-FROM postgres:${PG_MAJOR} AS build-latest
-ENV PG_MAJOR=${PG_MAJOR}
-ENV PIP_BREAK_SYSTEM_PACKAGES=1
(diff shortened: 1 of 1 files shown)


projects/extension/Dockerfile
```diff
@@ -45,14 +24,45 @@ RUN set -e; \
...
+    mkdir -p /usr/local/lib/pgai/$version && \
+    pip3 install -v --no-deps --compile -t /usr/local/lib/pgai/$version -r requirements.txt
+COPY . .
+RUN just install-py build-sql install-sql
+
+
+###############################################################################
+# image for use in extension development
+FROM base
+ENV WHERE_AM_I=docker
```
Askir, line 64: What is this Env var for?

Askir (approved):
Looks good to me. Wondering if there is an argument to be made that testing against the ha image is closer to how users will use pgai and could catch some issues? Maybe not relevant though, don't know how different the ha image actually is.

projects/extension/Dockerfile
jgpruitt: Many of the build/test commands (like `just test`) need to know whether they are being called from within the docker container or on the host machine. They use this environment variable to detect whether or not they are in the docker container and do "the right thing" depending.

jgpruitt:
@Askir we weren't really "using" the timescaledb-ha image. We were just copying the extensions' files out of it and into our image. The issue with using the timescaledb-ha image is that we will want to test new versions of postgres and timescaledb before the corresponding timescaledb-ha image is released. We end up with a chicken and egg issue.

PR #239, opened 2024-11-20
chore: sync the version in .control and python module

jgpruitt:


Askir (approved):
I had to google what a .control file is, so take this review with a grain of salt 😅

PR #240, opened 2024-11-20
test: support pg17 in test_contents

jgpruitt:
The \dx+ metacommand in v17 of psql returns different results than v16. Add different .expected files per major postgres version.


Askir (approved):
LGTM:
Do we plan to run test against both pg16 and pg17 in CI? Should the vectorizer also be tested against both? Or are we simply upgrading what we test against?

jgpruitt:
@Askir we need to test against both. It would be prudent to have vectorizer tested against both too.

PR #247, opened 2024-11-22
docs: document how to add a vectorizer integration

JamesGuthrie:
Compare this documentation with: https://github.com/timescale/pgai/pull/227


docs/adding-embedding-integration.md
```diff
@@ -0,0 +1,59 @@
...
+
+Add tests which perform end-to-end testing of the new integration. There are
+two options for handling API calls to the integration API:
+ 
+1. Use [vcr.py] to cache real requests to the API
+2. Run against the real API
+
+At minimum the integration should use option 1: vcr.py. Option 2 should be
+used conservatively. We will determine on a case-by-case basis what level of
+testing we would like.
```
jgpruitt, line 49: This may not match the latest discussion. Seems like we want to require both, but maybe if you implement vcr, you can do the real API's by putting it in a passthrough mode?

jgpruitt (approved):
This is awesome! Just one note.

docs/adding-embedding-integration.md
smoya: Idea: You can add a "Subject to change" in the meantime

smoya (approved):
This is awesome @JamesGuthrie 👏. We can iterate from here and add more detailed steps if needed, but with this base docs, i believe we are covering the critical points, so 💯.

alejandrodnm:
@billy-the-fish could you review this please?

PR #251, opened 2024-11-25
feat!: change type of keep_alive to text

JamesGuthrie:
The `keep_alive` parameter to the `ai.ollama_embed`, `ai.ollama_generate`, and `ai.ollama_chat_complete` functions now takes a `text` value instead of a `float8` value. This makes for a more flexible API, as `keep_alive` can now be specified using human-readable duration values like `'10m'`, in addition to numeric values like `'120'`.

This is a breaking change. The impact of this change is twofold:

1. Users who used `keep_alive => <some floating point value>` in the argument to one of `ai.ollama_embed`, `ai.ollama_generate`, `ai.ollama_chat_complete` must cast that value to text, i.e. `keep_alive => <some floating point value>::text`.

2. Users who have used one of the `ai.ollama_embed`, `ai.ollama_generate`, or `ai.ollama_chat_complete` functions in a SQL function using the `sql_body` form (see [CREATE FUNCTION]), or in a view will be forced to drop those functions and views, upgrade the pgai extension, and then re-create the functions and views.

[CREATE FUNCTION]: https://www.postgresql.org/docs/current/sql-createfunction.html


jgpruitt (approved):
looking good

projects/extension/sql/incremental/008-drop-ollama-functions-with-keep-alive.sql
JamesGuthrie, line 1: @jgpruitt what is the downstream impact of dropping these functions? How will it affect existing users? Is it possible that some users won't be able to upgrade the extension because of a dependency they have created on these functions?

cevian (approved):
I agree that getting the dx right is more important than back compat for now.

projects/extension/sql/incremental/008-drop-ollama-functions-with-keep-alive.sql
jgpruitt: Users would need to add an explicit cast to `text` where they are currently passing a float.
If they have existing functions in the SQL language or views that call these functions, the upgrade will fail. They'd need to drop these before the upgrade, and re-create after.

PR #252, opened 2024-11-25
fix: clarify env variable usage for OpenAI configuration.

billy-the-fish:


docs/vectorizer-api-reference.md
```diff
@@ -256,41 +256,50 @@ The embedding functions are:
...
-    embedding => ai.embedding_openai(
-      'text-embedding-3-small', 
-      768, 
-      chat_user => 'bob',
-      api_key_name => 'MY_OPENAI_API_KEY_NAME'
-    ),
-    -- other parameters...
-);
-```
+1. Set the value of your OpenAI API key [in an environmental variable][openai-set-key]. For example:
```
smoya, line 271: I believe the word `environmental` is not used in this context.
```suggestion
1. Set the value of your OpenAI API key [in an environment variable][openai-set-key]. For example:
```

docs/vectorizer-api-reference.md
```diff
@@ -256,41 +256,50 @@ The embedding functions are:
...
+
+   ```shell
+   echo "export OPENAI_API_KEY='yourkey'" >> ~/.zshrc
+   source ~/.zshrc
+   ```
```
smoya, line 276: This depends on the environment where the vectorizer worker runs. For example, if you run the worker through docker or not, etc. See https://github.com/timescale/pgai/blob/main/docs/vectorizer-worker.md#install-and-configure-vectorizer-worker

I don't think we need to be explicit here on how to set up an env var.

docs/vectorizer-api-reference.md
```diff
@@ -256,41 +256,50 @@ The embedding functions are:
...
+   
+2. In your vectorizer app, create your OpenAI configuration:
```
smoya, line 278: This is a database query. So you technically run it in your database. Additionally, you are not creating an OpenAI configuration here but a Vectorizer with OpenAI as embedding provider.

smoya (changes requested):
Thanks for the PR!! 🙌  

I left some requests for change. Feel free to reach me if any clarification is missing.

docs/vectorizer-api-reference.md
billy-the-fish: Updated.

docs/vectorizer-api-reference.md
billy-the-fish: Updated.

smoya (approved):
LGTM! 🚀🌔

I would suggest we have a final review by some of the other folks, such as @jgpruitt or @alejandrodnm (random)

PR #253, opened 2024-11-25
Add load dataset

cevian:
This function allows you to create a table from any dataset on
HuggingFace Hub. It's an easy way to load data for testing and
experimentation. The hub contains 250k+ datasets.

We use a streaming version of the API to allow for memory-efficient
processing and to optimize the case where you want less than
the entire dataset. On-disk caching should be minimal using streaming.


projects/extension/ai/load_dataset.py
```diff
@@ -0,0 +1,201 @@
...
+                elif type_str in ("int64", "int32", "int16", "int8"):
+                    batch_arrays[i] = [int(value) for value in array_values]
+                elif type_str in ("float64", "float32", "float16"):
+                    batch_arrays[i] = [float(value) for value in array_values]
+                else:
+                    batch_arrays[i] = array_values
+
+            num_rows += len(batch_arrays[0])
+            batch_count += 1
+            insert_plan.execute(batch_arrays)
```
jgpruitt, line 200: might be nice to log completion of each batch at a log or debug level

projects/extension/sql/idempotent/015-load_dataset.sql
```diff
@@ -0,0 +1,43 @@
+create or replace function ai.load_dataset
```
jgpruitt, line 1: we sure we want this to be a function rather than a procedure with commits after each batch?

jgpruitt (approved):
only minor stuff. this is super cool. will be my new favorite way to get datasets

projects/extension/ai/load_dataset.py
```diff
@@ -58,30 +66,57 @@ def create_table(
         # Replace any non-alphanumeric characters with underscore
         table_name = "".join(c if c.isalnum() else "_" for c in base_name.lower())
 
-    # Construct fully qualified table name
-    qualified_table = f'"{schema}"."{table_name}"'
+        # Check table name length - PostgreSQL has a 63 character limit for identifiers
+        if len(table_name) > 63:
```
jgpruitt, line 70: I think it's actually a 63 octet limit, which *should be the same unless someone starts using non-ascii characters, right?

projects/extension/ai/load_dataset.py
```diff
@@ -58,30 +66,57 @@ def create_table(
...
     plan = plpy.prepare(
         """
-        SELECT pg_catalog.to_regclass(pg_catalog.format('%I.%I', $1, $2)) is not null as exists
+        SELECT pg_catalog.format('%I.%I', $1, $2) as qualified_table_name
     """,
         ["text", "text"],
     )
     result = plan.execute([schema, table_name], 1)
-    table_exists = result[0]["exists"]
+    qualified_table = result[0]["qualified_table_name"]
```
jgpruitt, line 94: i like this way of constructing the fully-qualified table!

projects/extension/ai/load_dataset.py
```diff
@@ -58,30 +66,57 @@ def create_table(
...
+
+    # Check if table exists
+    result = plpy.execute(
+        f"""
+        SELECT pg_catalog.to_regclass('{qualified_table}')::text as friendly_table_name
```
jgpruitt, line 99: does this strip the schema if the search_path allows for it?

projects/extension/sql/idempotent/015-load_dataset.sql
```diff
@@ -1,9 +1,55 @@
...
+( name text
+, config_name text default null
+, split text default null
+, schema_name name default 'public'
+, table_name name default null
+, if_table_exists text default 'error'
+, field_types jsonb default null
+, batch_size int default 5000
+, max_batches int default null
+, commit_every_n_batches int default null
```
jgpruitt, line 11: are you sure you want commit_every_n_batches defaulted to null?

jgpruitt (approved):
left a few comments/questions, but LGTM

projects/extension/ai/load_dataset.py
cevian: yes it does. Is that a problem?

jgpruitt: No. Just curious

PR #254, opened 2024-11-26
fix: allow members of source table to create vectorizer

JamesGuthrie:
In f68e73ac we added support for cascading the source/target table drop to the queue and other dependent objects. In adding this, we added a security definer function which asserts that the owner of the source table is calling the security definer function.

Due to how security definer functions work with `current_user`, the ownership check opted to use `session_user` to check ownership. This is problematic because in connections which use `SET ROLE`, the `session_user` may not be the same as `current_user`.

This commit relaxes the ownership check to check whether the current
user is a member of the role that owns the source table.

### DIFF
--- a/projects/extension/sql/idempotent/012-vectorizer-int.sql
+++ b/projects/extension/sql/idempotent/012-vectorizer-int.sql
@@ -301,8 +301,8 @@ begin
     where v.id operator(pg_catalog.=) vectorizer_id
     ;
 
-    -- don't let anyone but the owner of the source table call this
-    select k.relowner operator(pg_catalog.=) pg_catalog.session_user()::pg_catalog.regrole
+    -- don't let anyone but the owner (or members of the owner's role) of the source table call this
+    select pg_catalog.pg_has_role(pg_catalog.session_user(), k.relowner, 'MEMBER')
     into strict _is_owner
     from pg_catalog.pg_class k
     inner join pg_catalog.pg_namespace n on (k.relnamespace operator(pg_catalog.=) n.oid)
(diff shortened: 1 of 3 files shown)


adolsalamanca (approved):

projects/extension/sql/idempotent/012-vectorizer-int.sql
```diff
@@ -301,8 +301,8 @@ begin
     where v.id operator(pg_catalog.=) vectorizer_id
     ;
 
-    -- don't let anyone but the owner of the source table call this
-    select k.relowner operator(pg_catalog.=) pg_catalog.session_user()::pg_catalog.regrole
+    -- don't let anyone but the owner (or members of the owner's role) of the source table call this
+    select pg_catalog.pg_has_role(_vectorizer_create_dependencies.caller, k.relowner, 'MEMBER')
```
alexeyklyukin, line 305: weird attribute qualification (_vectorizer_create_dependencies.caller instead of just caller), why?

alexeyklyukin (approved):
Looks good (a few nits)

projects/extension/sql/idempotent/012-vectorizer-int.sql
JamesGuthrie: done

projects/extension/sql/incremental/008-fix-broken-owner-check.sql
```diff
@@ -0,0 +1 @@
+drop function if exists ai._vectorizer_create_dependencies(integer);
```
cevian, line 1: Wait why do we need this

JamesGuthrie: doh - we don't. It was a stray leftover from before I removed some other bits.

JamesGuthrie: fixed.

cevian (approved):

PR #261, opened 2024-11-27
chore: start developing extension 0.5.1-dev

JamesGuthrie:


projects/extension/sql/ai.control
```diff
@@ -1,7 +1,7 @@
 # pgai extension
 comment='helper functions for ai workflows'
-default_version='0.5.0'
+default_version='0.5.1-dev'
 requires='vector,plpython3u'
 relocatable=false
 schema=ai
-no_relocate='vector'
```
jgpruitt, line 7: Was 0.5.0 released without this line?

jgpruitt (approved):

projects/extension/sql/ai.control
jgpruitt: The "no_relocate" line needs to be there. We don't want people to move the vector extension after the pgai extension is created

smoya (approved):

PR #263, opened 2024-11-28
fix: make vectorizer worker robust

JamesGuthrie:
These fixes were mostly done in the context of usability with docker compose. Includes the following changes:

- Handle being killed with SIGINT or SIGTERM
- Handle being unable to connect to the target DB
- Handle arbitrary failures in the core of vectorizer worker


jgpruitt (approved):
Lgtm!

projects/pgai/pgai/cli.py
```diff
@@ -218,42 +226,62 @@ def vectorizer_worker(
...
+            if can_connect and pgai_version is not None:
+                if not dynamic_mode and len(valid_vectorizer_ids) != len(
+                    vectorizer_ids
+                ):
+                    valid_vectorizer_ids = get_vectorizer_ids(db_url, vectorizer_ids)
+                    if len(valid_vectorizer_ids) != len(vectorizer_ids):
+                        log.critical(
+                            f"invalid vectorizers, wanted: {list(vectorizer_ids)}, got: {valid_vectorizer_ids}"  # noqa: E501 (line too long)
+                        )
+                        sys.exit(1)
```
jgpruitt, line 266: Do we want to fail if we request 5 vectorizers and only find 4 (for example)? Why not warn about the one and work the 4?

projects/pgai/pgai/cli.py
jgpruitt, line 1: Should we have a "die if unable to connect to db after x minutes" option?

projects/pgai/pgai/cli.py
JamesGuthrie: Note: I didn't change that behaviour in this PR.

The original reasoning behind the current behaviour was: "if you've told the vectorizer worker to process a specific set of vectorizers, then it should fail if that specific set isn't completely available". The idea being that it would help in catching misconfigurations.

jgpruitt (approved):

projects/pgai/pgai/cli.py
JamesGuthrie: I've changed this now. I've added an `--exit-on-error` parameter. When you specify `--once` it implies `--exit-on-error`. This means that if you run the vectorizer worker in `--once` mode, and tell it you want it to process a specific vectorizer that doesn't exist, it will fail and exit with exit code 1. If you run it in "normal" mode, then it will log a warning, but not exit, unless you tell it to (with `--exit-on-error`).

projects/pgai/pgai/cli.py
```diff
@@ -217,43 +232,82 @@ def vectorizer_worker(
...
+                ):
+                    valid_vectorizer_ids = get_vectorizer_ids(db_url, vectorizer_ids)
+                    if len(valid_vectorizer_ids) != len(vectorizer_ids):
+                        if exit_on_error:
+                            log.error(
+                                f"invalid vectorizers, wanted: {list(vectorizer_ids)}, got: {valid_vectorizer_ids}"  # noqa: E501 (line too long)
+                            )
+                            sys.exit(1)
+                        else:
+                            log.warn(
```
smoya, line 284: Why warn instead of err? 🤔

JamesGuthrie: good question. "upgraded" to error.

PR #266, opened 2024-12-02
feat: add alembic operations for vectorizer

Askir:
This PR adds native python operations to alembic so you don't have to write SQL to create vectorizers.


projects/pgai/pgai/configuration.py
```diff
@@ -0,0 +1,361 @@
...
+class SQLArgumentProvider(Protocol):
+    def to_sql_argument(self) -> str: ...
+
+
+@runtime_checkable
+class PythonArgProvider(Protocol):
+    def to_python_arg(self) -> str: ...
+
+
+def format_python_arg(config_type: str, instance: Any) -> str:
```
cevian, line 20: can't we right a similar function for sql to make all this simpler to write? I mean all parameters can be passed in with {name} => {value} and the way to render the value is type dependent, which we can switch on. We can also skip any None's etc

cevian (changes requested):
I gotta say I'm not convinced about the arguments for using a separate model than the models already in pgai/vectorizer or at least having both sets of models extend a base model. I think having 2 sets of models with similar params is really hard to maintain and quite a bit of code duplication. I'd like some more eyes on this tho. Can 
James and/or Alejandro chime in here. In particular I'd like us to consider three designs:
- simply extending the pydantic model we already have with optional fields that are present in either the stored json OR needed for the alembic stuff + having some kind of wrappers to create the config objects in alembic.
- Factoring common data fields into base classes and using those as mixins. (kinda like the ApiKeyMixin now).
- Maybe I'm just being stubborn and we should have separate models, like Jascha has them now.
leaving a few comments in but I think this is the big issue we need to resolve

projects/pgai/tests/vectorizer/extensions/fixtures/migrations/002_create_vectorizer.py.template
```diff
@@ -0,0 +1,49 @@
...
+            model='text-embedding-3-small',
+            dimensions=768
+        ),
+        chunking=CharacterTextSplitterConfig(
+            chunk_column='content',
+            chunk_size=800,
+            chunk_overlap=400,
+            separator='.',
+            is_separator_regex=False
+        ),
```
Askir, line 35: This is a bit of an ugly side effect of having merged the parent classes now.
The pydantic models that map the db config have these fields as non optional because the extension code provides the defaults.

But the **input function** should have these as optional for the same reason (if they are null the extension will provide e.g. `chunk_size=800`. 

So now in the python code you have to unnecessarily always provide all the arguments (true for some of the other configs as well).

Alternatively I could give these the same defaults as the extension code provides, but then we have to again maintain the defaults in 2 places.

This was previously not an issue as I could just make the input arguments nullable while leaving the load from db models required. I can override this of course in the child classes, but then again this breaks the purpose of sharing a parent class in the first place.

docs/python-integration.md
```diff
@@ -68,22 +68,50 @@ The model is only created at runtime, so depending on how your alembic migration
...
-```sql
-SELECT ai.create_vectorizer(
-    'blog_posts'::regclass,
-    target_table => 'blog_posts_content_embeddings_store',
-    embedding => ai.embedding_openai('text-embedding-3-small', 768),
-    chunking => ai.chunking_recursive_character_text_splitter(
-        'content',
-        50,  -- chunk_size
-        10   -- chunk_overlap
+pgai provides native Alembic operations for managing vectorizers. For them to work you need to run `setup_alembic` in your env.py file. Which registers the pgai operations under the global op context:
```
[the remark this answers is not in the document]
Askir: I renamed this to `register_operations` now and explained it here: https://github.com/timescale/pgai/pull/266#discussion_r1883753417
Hope that makes it a bit more clear what this is for.

projects/pgai/tests/vectorizer/extensions/fixtures/migrations/002_create_vectorizer.py.template
Askir: fixed it now by using a custom decorator that makes fields required for parsing internally but leaves them optional for the migrations.

projects/extension/Dockerfile
```diff
@@ -52,6 +52,10 @@ WORKDIR /pgai
 COPY . .
 RUN just build install
 
+RUN mkdir -p /docker-entrypoint-initdb.d && \
+    echo "#!/bin/bash" > /docker-entrypoint-initdb.d/configure-timescaledb.sh && \
+    echo "echo \"shared_preload_libraries = 'timescaledb'\" >> \${PGDATA}/postgresql.conf" >> /docker-entrypoint-initdb.d/configure-timescaledb.sh && \
+    chmod +x /docker-entrypoint-initdb.d/configure-timescaledb.sh
```
Askir, line 58: I had to add this to be able to run `create extension if not exists timescaledb` I'm not sure this is correct?

projects/pgai/pgai/vectorizer/chunking.py
```diff
@@ -52,19 +58,14 @@ class LangChainCharacterTextSplitter(BaseModel, Chunker):
...
-    chunk_size: int
-    chunk_column: str
-    chunk_overlap: int
-    is_separator_regex: bool
 
     @cached_property
     def _chunker(self) -> CharacterTextSplitter:
         return CharacterTextSplitter(
-            separator=self.separator,
+            separator=self.separator,  # type: ignore
```
Askir, line 65: This `type: ignore` is now necessary as pyright does not know that the decorator makes the field required and complains about possibly passing `None`.

projects/pgai/pgai/alembic/operations.py
```diff
@@ -0,0 +1,123 @@
...
+        drop_all: bool = True,
+    ):
+        """Issue a DROP VECTORIZER command."""
+        op = DropVectorizerOp(vectorizer_id, drop_all)
+        return operations.invoke(op)
+
+    @override
+    def reverse(self) -> MigrateOperation:
+        """Creates the upgrade operation"""
+        return CreateVectorizerOp(None)
```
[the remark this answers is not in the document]
Askir: actually the whole `reverse` function is not required when not using autogeneration so I removed it.

projects/pgai/pgai/vectorizer/base.py
```diff
@@ -0,0 +1,97 @@
...
+                # Create new required field with None removed from annotation
+                new_annotation = remove_none_from_annotation(field.annotation)
+                new_field = FieldInfo.from_annotated_attribute(
+                    new_annotation,  # type: ignore
+                    ...,  # Using ... as default makes it required
+                )
+                new_fields[name] = new_field
+            else:
+                new_fields[name] = field
+        _cls.model_fields = new_fields
```
Askir, line 94: Oh... I'm trying to build a sample application right now. And in this repo we use pydantic 2.9 where this works but in pydantic 2.10 this already breaks... 

I'll have to go back through this and find another way, but either way this seems brittle. Pydantic is not really designed to allow such overrides their idea is to use the typing system to infer the validation logic, overriding the types breaks with this declarative approach.

docs/python-integration.md
```diff
@@ -182,3 +182,49 @@ context.configure(
...
+            chunk_overlap=400,
+            separator='.',
+            is_separator_regex=False
+        ),
+        formatting=PythonTemplateConfig(template='$title - $chunk')
+    )
+
+
+def downgrade() -> None:
+    op.drop_vectorizer(vectorizer_id=1, drop_all=True)
```
cevian, line 227: i think it would be better for this to take the target_table name and not the vectorizer_id (which would probably not be known when writing the migration). The target table should be unique and so we should be able to look up the id from that

projects/extension/Dockerfile
cevian: why did you need timescaledb for this pr? This is a dev image so this is fine I'm just curious

projects/pgai/pgai/alembic/configuration.py
```diff
@@ -0,0 +1,227 @@
...
+
+class SQLArgumentMixin:
+    arg_type: ClassVar[str]
+    function_name: ClassVar[str | None] = None
+
+    def to_sql_argument(self) -> str:
+        # Get all fields including from parent classes
+        params = {}
+        for field_name, _field in self.model_fields.items():  # type: ignore
+            if field_name != "arg_type":
```
cevian, line 55: is the function_name field included then, how does that work?

projects/pgai/pgai/alembic/configuration.py
```diff
@@ -0,0 +1,227 @@
...
+            # Convert camelCase to snake_case
+            base_name = re.sub("([a-z0-9])([A-Z])", r"\1_\2", base_name).lower()
+            # Remove 'config' and clean up any double underscores
+            base_name = base_name.replace("config", "").strip("_")
+            fn_name = f"{self.arg_type}_{base_name}"
+
+        return f", {self.arg_type} => ai.{fn_name}({format_sql_params(params)})"  # type: ignore
+
+
+class OpenAIConfig(BaseOpenAIConfig, SQLArgumentMixin):
```
cevian, line 73: Naming (and I know naming discussions are always annoying) but why not stick to the sql convention we established and name this EmbeddingOpenAIConfig or EmbeddingConfigOpenAI? (and similar for others). The pro is that the name translation from sql->python is super easy and I think would be easier to understand. The con is that it's long. 

Otherwise the translation seems a bit ad-hoc. e.g. Indexing configs have "indexing" in the name but in a different spot than the sql. Let's think about this some more

projects/pgai/pgai/alembic/configuration.py
```diff
@@ -0,0 +1,227 @@
+import re
```
cevian, line 1: To my eye this is a huge improvement from before

projects/pgai/pgai/vectorizer/chunking.py
```diff
@@ -33,7 +38,8 @@ def into_chunks(self, item: dict[str, Any]) -> list[str]:
         """
 
 
-class LangChainCharacterTextSplitter(BaseModel, Chunker):
+@required
```
cevian, line 41: I only see these used in 2 places? don't we need it on more models?

docs/python-integration.md
```diff
@@ -164,7 +164,7 @@ for post, embedding in results:
```
cevian, line 164: I think we also need to add docs to adding-embedding-integration.md

cevian (changes requested):
This is looking a lot better. Some questions remaining but I think this is the right track.

projects/pgai/pgai/alembic/configuration.py
Askir: not sure how that worked, but now its excluded too :D

projects/pgai/pgai/alembic/configuration.py
Askir: I think thats fixed with the code gen!

docs/python-integration.md
Askir: Done ✅

docs/python-integration.md
Askir: Done ✅

cevian (changes requested):
This is soooo much better! Now seems like the right approach! Mostly just nits in the comments.

Please get one other person to review. But, again, looks great

projects/pgai/pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
     "typing_extensions>=4.0,<5.0",
     "datadog_lambda>=6.9,<7.0",
     "pytimeparse>=1.1,<2.0",
-    "voyageai>=0.3.1,<0.4.0",
+    "voyageai>=0.3.1,<0.3.2", # 0.3.2 fails the tests somehow
```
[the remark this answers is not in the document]
Askir: Nah, I just relocked the dependencies and uv tried to update to 0.3.2 which fails the tests. I think we just have to rerecord the requests, but I'll do this in a separate PR.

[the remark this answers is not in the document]
Askir: Nah it's not due to this change, I just ran `uv lock` to get a lock file with the new dependencies I added (alembic, sqlalchemy) and the ones that were added to main since i created this PR. And that tried to update voyageai to 0.3.2 (which doesn't work).
I think we just have to rerecord the API calls, but I'll do that in a separate PR. One of the disadvantages of mocking these 😇

docs/python-integration.md
Askir: good point now they are :D

projects/pgai/pgai/alembic/operations.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+
+    @classmethod
+    def create_vectorizer(cls, operations: Operations, **kw: Any):
+        op = CreateVectorizerOp(**kw)  # type: ignore
+        return operations.invoke(op)
+
+
+class DropVectorizerOp(MigrateOperation):
+    def __init__(self, table_name: str | None, drop_all: bool):
```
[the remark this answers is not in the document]
Askir: went with target_table since that's what create_vectorizer also calls it.

docs/python-integration.md
```diff
@@ -1,3 +1,38 @@
+# Creating vectorizers from python
```
cevian, line 1: If we are going to add this to the docs, I'd like at least some tests of using this functionality outside of Alembic

cevian (approved):
LGTM. Two comments that need addressing. Would like for @JamesGuthrie to review next and then we can merge

projects/pgai/pgai/vectorizer/generate/function_parser.py
```diff
@@ -0,0 +1,204 @@
...
+                    FROM pg_type t
+                    WHERE t.oid = %s
+                """
+                cur.execute(type_query, (type_oid,))
+                type_info = cur.fetchone()
+
+                type_name = type_info[1]  # type: ignore
+                is_array = type_info[2]  # type: ignore
+
+                default = None
```
JamesGuthrie, line 112: Is this correct?

projects/pgai/pgai/vectorizer/generate/function_parser.py
```diff
@@ -0,0 +1,204 @@
...
+        # Skip internal parameters that shouldn't be exposed
+        if param.name.startswith("_"):
+            continue
+
+        accepted_configs = None
+        python_type = param.python_type
+
+        # For each parameter, check if there is corresponding config functions
+        with conn.cursor() as cur:
+            cur.execute(config_functions_query, (f"{param.name}_%",))
```
JamesGuthrie, line 182: I'm sure you're aware, but this has the assumption that the name of the configuring function contains the parameter name. I suspect we might trip over this at some point in the future.

JamesGuthrie (approved):
LGTM, a few minor bits and pieces.

projects/pgai/pgai/vectorizer/generate/generate.py
```diff
@@ -0,0 +1,69 @@
...
+from pgai.vectorizer.generate.config_generator import (
+    generate_config_classes,
+    generate_vectorizer_params,
+)
+from pgai.vectorizer.generate.function_parser import (
+    get_function_metadata,
+    read_create_vectorizer_metadata,
+)
+
+VECTORIZER_FUNCTIONS = [
```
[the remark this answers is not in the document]
Askir: very little, I thought about somehow parsing all functions that are in the ai extension first. But there is a lot of functions that we probably don't want to have in python and I don't know how to distinguish the ones that can be used as vectorizer arguments from the ones that can't be.
Other option would be to try filter the function name out from the vectorizer-api-reference.md docs file but that also felt hacky.

projects/pgai/pgai/vectorizer/generate/function_parser.py
```diff
@@ -0,0 +1,204 @@
...
+) -> list[PostgresFunction]:
+    """Extract function metadata from Postgres catalog."""
+
+    query = """
+        SELECT
+            p.proname,
+            n.nspname,
+            p.proargnames,
+            p.pronargdefaults,
+            string_to_array(array_to_string(p.proargtypes, ' '), ' ') as argtypes,
```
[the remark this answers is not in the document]
Askir: I actually didn't understand this code in detail either, it's something that claude came up with but works so I didn't worry too much about it.

projects/pgai/pgai/vectorizer/generate/function_parser.py
Askir: yeh that was a happy coincidence with our current function naming, if that's no longer the case, see the discussion about figuring out the function names in the first place, either we have to maintain a manual mapping in the generator or we figure out a way to automatically find the association between parameter and possible function out somehow.

projects/pgai/pgai/vectorizer/generate/generate.py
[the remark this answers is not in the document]
JamesGuthrie: Ya, I understand that there's no obvious, simple, solution. I'm happy to punt on it.

projects/pgai/pgai/vectorizer/generate/function_parser.py
JamesGuthrie: ack.

projects/pgai/pgai/vectorizer/generate/function_parser.py
Askir: This is actually a bit misleading thanks for questioning it. I removed the default value from the code base now.

The generated classes don't actually have the correct default values the to_sql function just laves any None value out so that the ai.xyz() call uses the sqls default value. So the `default` is always None in this script (otherwise I'd have to correctly parse the sql default values into its python representation, which I am saving on this way.

docs/adding-embedding-integration.md
```diff
@@ -56,6 +56,11 @@ testing we would like.
 [embedders directory]:/projects/pgai/pgai/vectorizer/embedders
 [embedders \_\_init\_\_.py]:/projects/pgai/pgai/vectorizer/embedders/__init__.py
 
+## pgai library
+The pgai library exposes helpers to create a vectorizer via pure python.
+The classes for this are autogenerate via code generation. To update the classes
+with a new integration look into the code generator docs in [/project/pgai/pgai/alembic/generate](/pgai/vectorizer/generate/README.md).
```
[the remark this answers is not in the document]
Askir: fixed (I think)

PR #267, opened 2024-12-02
feat: add autogenerate support to sqlalchemy and  alembic

Askir:
Adds autogeneration support to the vectorizer field by allowing to configure (almost) all the create_vectorizer params on the field itself.
The user can run `alembic revision --autogenerate` and it will create the corresponding migration file with create_vectorizer.
Updates are handled by first dropping the original one and then creating a new one.

This works by loading the existing config from the database and then comparing it with whatever is configured by the user. The configuration classes implement `__eq__` for this purpose. Handling default parameters and mapping everything back to create_vectorizer params is quite complex, but it does work right now.


projects/extension/Dockerfile
```diff
@@ -51,6 +50,13 @@ WORKDIR /pgai
 COPY . .
 RUN just build install
 
+# Create a custom config file in docker-entrypoint-initdb.d
+RUN mkdir -p /docker-entrypoint-initdb.d && \
+    echo "#!/bin/bash" > /docker-entrypoint-initdb.d/configure-timescaledb.sh && \
+    echo "echo \"shared_preload_libraries = 'timescaledb'\" >> \${PGDATA}/postgresql.conf" >> /docker-entrypoint-initdb.d/configure-timescaledb.sh && \
+    chmod +x /docker-entrypoint-initdb.d/configure-timescaledb.sh
```
Askir, line 57: I ran into issues when running `Create extension timescaledb` and claude recommended this, i'm not quite sure what this does but it works.

projects/pgai/pgai/configuration.py
```diff
@@ -65,6 +126,12 @@ class OllamaEmbeddingConfig:
     truncate: bool | None = None
     keep_alive: str | None = None
 
+    _defaults = {"dimensions": 1536, "api_key_name": "OPENAI_API_KEY"}
```
Askir, line 129: these defaults are actually wrong 🙈 will fix

Askir:
By now this PR is very out of date and will have a bunch of merge conflicts, I also really doubt the value of maintaining all the defaults again and again. So I'm closing it for now. Maybe there is a better approach.

PR #271, opened 2024-12-02
docs: focus on Ollama

JamesGuthrie:


README.md
```diff
@@ -201,11 +201,11 @@ data in the source table and update the destination embedding table
...
-to implement embeddings in your own data. When you create Vectorizers in a Timescale
-Cloud database, embeddings are automatically created and synchronized in the background.
-On a self-hosted Postgres installation, you use a [Vectorizer
-Worker](/docs/vectorizer-worker.md) to asynchronously processes your
-vectorizers.
+to implement embeddings in your own data. On a self-hosted Postgres
+installation, you use a [Vectorizer Worker](/docs/vectorizer-worker.md) to
+asynchronously processes your vectorizers. When you create Vectorizers in a
+Timescale Cloud database, embeddings are automatically created and synchronized
+in the background. Note: Timescale Cloud only supports embedding with OpenAI.
```
Askir, line 208: In theory you can disable scheduling and run the vectorizer worker with ollama yourself right? While still having the db hosted on cloud?

alejandrodnm (approved):

README.md
JamesGuthrie: Yes! Do you think that's something that we should suggest to users?

docs/vectorizer-api-reference.md
```diff
@@ -25,7 +25,7 @@ operations such as INSERT, UPDATE, and DELETE.
 
 - **Configurable embedding process**: a vectorizer is highly configurable, 
    allowing you to specify:
-    - The embedding model and dimensions. For example, OpenAI's `text-embedding-3-small`.
+    - The embedding model and dimensions. For example, Ollama's `nomic-embed-text`.
```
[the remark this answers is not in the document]
JamesGuthrie: done

README.md
Askir: I would definitely be up for that. Hosting databases is always a bit tricky because you run the risk of losing data if you do something stupid. But the vectorizer is stateless so if I was building an LLM application I think I'd be fine with such a setup.

PR #274, opened 2024-12-03
fix: fail fast when api key is missing and once is set

smoya:


projects/pgai/pgai/cli.py
```diff
@@ -286,6 +286,9 @@ def vectorizer_worker(
                 for vectorizer_id in valid_vectorizer_ids:
                     vectorizer = get_vectorizer(db_url, vectorizer_id)
                     if vectorizer is None:
+                        log.error("error fetching vectorizer", vectorizer_id=vectorizer_id) # noqa: E501 (line too long)
```
JamesGuthrie, line 289: Nit: this makes me feel like it's about "fetching the vectorizer from the db", but as we discussed, there can be two reasons for a None value here. I would suggest making this error message a bit more generic. I miss a go-like `val, err := ...` pattern here though, which would propagate a more meaningful error.

smoya: Not a nit imho; it is an important suggestion because I kinda felt the same when writing this line but then my brain decided it was good enough.

> I miss a go-like val, err := ... pattern here though, which would propagate a more meaningful error.

Totally agree. I miss it in almost every language other than, eh... actually Go. Many people hate it but I love it because you have control and can acknowledge any error.

I believe the closest idiomatic behavior would be to raise an exception and catch it. I don't find it a bad solution in fact in this piece of code.

JamesGuthrie (approved):

PR #278, opened 2024-12-05
chore: enable extension tests in ci

Askir:
I replaced the ENABLE_ flags for each API provider with simply checking if a key is present. I think this makes more sense so you don't have to provide both env vars?

I also added a download fixture to the ollama tests so that the necessary models are downloaded during the test runs. This is quite slow, but I guess we can add caching later.

Other than that this was quite straight forward. Only issue is that voyageai already hits rate limits with my free key now.

### DIFF
--- a/projects/extension/build.py
+++ b/projects/extension/build.py
@@ -590,7 +590,13 @@ def where_am_i() -> str:
 
 def test_server() -> None:
     if where_am_i() == "host":
-        cmd = "docker exec -it -w /pgai/tests/vectorizer pgai-ext fastapi dev server.py"
+        cmd = """docker exec \
+            -e OPENAI_API_KEY=${OPENAI_API_KEY} \
+            -e ANTHROPIC_API_KEY=${ANTHROPIC_API_KEY} \
+            -e COHERE_API_KEY=${COHERE_API_KEY} \
+            -e VOYAGE_API_KEY=${VOYAGE_API_KEY} \
+            -it -w /pgai/tests/vectorizer pgai-ext fastapi dev server.py
+            """
         subprocess.run(cmd, shell=True, check=True, env=os.environ, cwd=ext_dir())
     else:
         cmd = "fastapi dev server.py"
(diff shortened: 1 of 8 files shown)


projects/extension/build.py
```diff
@@ -648,7 +648,7 @@ def docker_build() -> None:
 def docker_run() -> None:
     cmd = " ".join(
         [
-            "docker run -d --name pgai-ext -p 127.0.0.1:5432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust",
+            "docker run -d --name pgai-ext --network host -p 127.0.0.1:5432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust",
```
JamesGuthrie, line 651: Why was this change necessary?

projects/extension/tests/test_ollama.py
```diff
@@ -1,16 +1,58 @@
...
+    raise TimeoutError(f"Timeout waiting for model {model} to be ready")
+
+
+@pytest.fixture(scope="session", autouse=True)
+def ensure_models(ollama_host):
+    """
+    Ensure required models are downloaded before running tests.
+    This fixture runs automatically at the start of the test session.
+    """
+    required_models = ["llama3", "llava:7b"]
```
JamesGuthrie, line 50: Daaamn, these models are large (each ~4.7 GB).

projects/extension/build.py
Askir: because the ollama container and the container in which we run tests, otherwise cannot talk to each other.
The alternative would be to run both in a docker-compose or otherwise share a network. But I felt like this was simpler and should also help for local development if you run ollama separately.

projects/extension/tests/test_ollama.py
JamesGuthrie: I think for testing we're mostly trying to assert that it does _something_. I would suggest that we switch `llama3` to `llama3.2:1b`, and `llava:7b` to `smollm:138m`.

Askir: Updated and 5 minutes now, seems good to me.

JamesGuthrie (approved):
I added a commit with something that I need to get tests to pass locally. Otherwise LGTM.

jgpruitt:
nice!

Issue #23, opened 2024-06-10
Support bulk operations


jgpruitt:
What is the most performant/efficient way to embed lots of rows? Can we build functions or procedures to make this easy? If not, can we document guidance and provide example code?

kolaente:
I have hit this problem with an application I'm building (not yet with pgai). We were ingesting so much data into the system that we'd run into openai's rate limits. The solution here was to build a batch processing job which creates openai embedding batches and, on a schedule, checks if openai has processed the batch, then saving the returned embeddings into the databse.

I wonder if pgai could do something like this as well?

theodufort:
I met the same problem when I wanted to bulk generate embeddings for 25 million + rows, I can not do it without reaching `OSError: [Errno 24] Too many open files`

theodufort:
Right now only way that is working for me without having any error is doing it by small batches this way: 
`UPDATE public.subjects
SET subject_embedding = ai.ollama_embed('nomic-embed-text:v1.5', name)
WHERE id IN (
    SELECT id
    FROM public.subjects
    WHERE subject_embedding IS NULL
    LIMIT 100
);`

alejandrodnm:
Have you tried setting up a [vectorizer](https://github.com/timescale/pgai/blob/main/docs/vectorizer.md)? When you run the worker, you can specify the number of batches, concurrency, and poll interval. 

Batches are sent in a single request to openAI.

Of course, if you tried to do too much in a small period, you'll bound to be rate limited. We currently don't support the openAI's batch API. The alternative will be setting the vectorizer's config to stay below rate limit's threshold. This will work if you get spikes of ingest, and you're fine with some delay between inserts and generating the embeddings.

kolaente:
I haven't set one up yet, still exploring options.

Is it possible to either extend the vectorizer to support adding embeddings through openai's batch api or manually, via a service which I run? (The latter would then do all the checking and batch creation etc, but would require marking a chunk as "this will be created asynchronously, please do not do anything")

alejandrodnm:
@kolaente Supporting the batch API in the vectorizer worker docker image could take some effort, and it's not something in our current roadmap. But you can extend the vectorizer, and make your own worker. Once you have something running, we can discuss integrating that into the pgai repo.

When you create a vectorizer in your DB with the `ai.create_vectorizer` function, an embeddings store table, and a queue table will be created. The queue table will be populated whenever something changes in your source table.

These is the query we use to fetch items from the queue:

[...]

PR #280, opened 2024-12-05
feat: use openai's batch processing to create large volumes of embeddings

kolaente:
This PR adds support for OpenAI's [batch api](https://platform.openai.com/docs/guides/batch/batch-api?lang=python). The implemented functionality creates batches for async processing and then checks their status. This allows processing huge amounts of data a lot more quickly, because the batch api has significantly higher rate limits.

The PR is very much a draft at this point, please see the `TODO`s I've added in various places. I've run this locally, and it seems to work, but the way I implemented this is probably not the best way to do this - I attribute this to my lack of knowledge about the codebase and my python skills.
I have implemented a similar thing in the past (which is currently running in production), but in a Laravel application.

Resolves https://github.com/timescale/pgai/issues/23

## Open Questions and To-Dos:

These are things I need help with (because I don't know the codebase good enough). Some of them also have `TODO`s in code.

- [x] Use the configured embeddings model instead of hardcoding `text-embedding-3-small`
- [x] Use the existing openai client instead of recreating one each time
- [x] ~~Run the batch embedding only when the backlog of items to be processed is very large (ideally, this would be configurable)~~ Changed so that the switch needs to happen manually
- [x] Run at the correct place in the code, and only when Openai is configured and supported
- [x] Properly create the new tables with a migration or similar mechanism
- [x] Delete items that were submitted to batch processing from the queue
- [ ] Increase the number of items put in a batch - as the batch api can handle significantly higher volume than what fits in typical rate limits, we should use that
- [ ] Fix the concurrency issue ([see comment](https://github.com/timescale/pgai/pull/280#issuecomment-2521246062)) 
- [ ] Fix out of memory error ([see comment](https://github.com/timescale/pgai/pull/280#discussion_r1871975461))
- [x] Make sure the batch fetch query is save for concurrent use
- [x] Delete batch files once the batch is processed or failed


projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -622,6 +789,67 @@ async def _embed_and_write(self, conn: AsyncConnection, items: list[SourceRow]):
...
+                    resolved_pk = pk_names.split(',')
+                    item = {pk: id_value for pk, id_value in zip(resolved_pk, resolved_id)}
+                    item[self.vectorizer.config.chunking.chunk_column] = embedding_batch_chunks[custom_id]
+
+                    all_items.append(item)
+                    all_records.append([resolved_id + [chunk_seq, embedding_batch_chunks[custom_id]] + [np.array(embedding_data)]])
+
+        await self._delete_embeddings(conn, all_items)
+        for records in all_records:
+            await self._copy_embeddings(conn, records)
```
kolaente, line 849: This fails with postgres running out of memory:

```
ERROR:  out of memory
DETAIL:  Cannot enlarge string buffer containing 0 bytes by 2130706440 more bytes.
CONTEXT:  COPY wiki_orig_embeddings_store, line 1, column wiki_id
STATEMENT:  COPY "public"."wiki_orig_embeddings_store" ("wiki_id" ,"paragraph_id", chunk_seq, chunk, embedding) FROM STDIN WITH (FORMAT BINARY)
```

It first failed when I tried inserting all records at once, but inserting them one by one does not seem to work either.

kolaente:
Running this with more than one worker fails with this error:

```
2024-12-05 18:16:23.584 UTC [1518] ERROR:  deadlock detected at character 30
2024-12-05 18:16:23.584 UTC [1518] DETAIL:  Process 1518 waits for RowExclusiveLock on relation 24075 of database 16384; blocked by process 1517.
	Process 1517 waits for RowExclusiveLock on relation 24075 of database 16384; blocked by process 1518.
	Process 1518:
	                INSERT INTO ai.embedding_batches (
	                    openai_batch_id,
	                    input_file_id,
	                    output_file_id,
	                    status,
	                    errors,
	                    expires_at
	                ) VALUES (
	                    $1,
	                    $2,
	                    $3,
	                    $4,
	                    $5,
	                    $6
	                )

	Process 1517:
	                INSERT INTO ai.embedding_batches (
	                    openai_batch_id,
	                    input_file_id,
	                    output_file_id,
	                    status,
	                    errors,
	                    expires_at
	                ) VALUES (
	                    $1,
	                    $2,
	                    $3,
	                    $4,
	                    $5,
	                    $6
	                )

2024-12-05 18:16:23.584 UTC [1518] HINT:  See server log for query details.
2024-12-05 18:16:23.584 UTC [1518] STATEMENT:
	                INSERT INTO ai.embedding_batches (
	                    openai_batch_id,
	                    input_file_id,
	                    output_file_id,
	                    status,
	                    errors,
	                    expires_at
	                ) VALUES (
	                    $1,
	                    $2,
	                    $3,
	                    $4,
	                    $5,
	                    $6
	                )
```

Setting the worker count to 1 fixes this, but that's obviously not a solution.

Can someone point me in the right direction on how to fix this?

projects/pgai/pgai/vectorizer/embeddings.py
```diff
@@ -404,6 +406,50 @@ async def embed(
...
+
+        temp_file = tempfile.NamedTemporaryFile(delete=False, suffix='.jsonl', mode='w')
+
+        for document in documents:
+            entry = {
+                'custom_id': document['unique_full_chunk_id'],
+                'method': 'POST',
+                'url': '/v1/embeddings',
+                'body': {
+                    'model': 'text-embedding-3-small', # TODO how can I use the configured embeddings model?
```
Askir, line 432: We load the configured model from the database via pydantic so just calling `self.model` will give you the model name here.

projects/pgai/pgai/vectorizer/embeddings.py
```diff
@@ -404,6 +406,50 @@ async def embed(
...
+                    'input': document['chunk'],
+                },
+            }
+            temp_file.write(json.dumps(entry) + '\n')
+
+        temp_file.close()
+
+        client = openai.OpenAI() # TODO there has to be a client already which I could use instead?
```
Askir, line 440: `self._embedder` is the property that's currently used but I don't think this exposes the batch embedding api.
Probably makes sense to unify this under a simple `self._client` property.

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -464,13 +467,177 @@ async def run(self) -> int:
             await register_vector_async(conn)
             while True:
                 if not self._continue_processing(loops, res):
+                    await self._check_and_process_openai_batches(conn)
+                    # TODO how can we run this only after hitting the rate limit of the normal openai batch embedding api?
```
Askir, line 471: I am not entirely sure this is how it should work.
My gut idea would be to have a flag on the `create_vectorizer` call that allows enabling and disabling the batch_api. I'd then also put a configurable threshhold there, which would trigger the batch calls when the queue grows e.g. larger than 1000 items?

Not entirely sure about this interface. Maybe @alejandrodnm has another idea.

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -464,13 +467,177 @@ async def run(self) -> int:
...
+                    await self._do_openai_batch(conn)
                     return res
                 items_processed = await self._do_batch(conn)
                 if items_processed == 0:
                     return res
                 res += items_processed
                 loops += 1
 
+    async def _create_batch_table(self, conn: AsyncConnection):
+        # TODO this does not feel like the way to go, is there a way to do these kind of migrations properly?
```
Askir, line 481: Have a look at the extension code and the create_vectorizer function in particular. We create e.g. the queue table there so this code should probably also live in there.
I also think we need one of these for each vectorizer and not a shared one for all.

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -622,6 +789,67 @@ async def _embed_and_write(self, conn: AsyncConnection, items: list[SourceRow]):
 
         return len(records)
 
+    @tracer.wrap()
+    async def _embed_and_write_from_batch(
```
Askir, line 793: Shouldn't this just be called `_write_from_batch`? Since the embedding has already completed when the function is called.

Askir (commented):
Hey, this is super cool. Thank you so much for providing a PR already in an early stage so we can give some input.

I tried to answer some of your questions, but didn't look at everything in detail yet.

projects/pgai/pgai/vectorizer/embeddings.py
kolaente: So I'd just add that as a new property? AFAIK just initializing will look for credentials from the environment, is that what we want?

projects/pgai/pgai/vectorizer/vectorizer.py
kolaente: I think making this configurable makes sense. It will be easier to implement at least :grin: 

How would I create a new config variable for this? I know how to add a new env variable for this, but maybe that's not what we want?

projects/pgai/pgai/vectorizer/embeddings.py
kolaente: Done.

projects/pgai/pgai/vectorizer/vectorizer.py
kolaente: Renamed it to `_write_embeddings_from_batch` - the original name was inspired by the same function which does this using the regular api.

projects/pgai/pgai/vectorizer/vectorizer.py
kolaente: I imagine I'll need to create a migration for this? Do I need to redefine everything or only the `create_vectorizer` function?

alejandrodnm:
Hey @kolaente, awesome work. I'm going to start reviewing this. Since this is a non-trivial change, we are going to consider this a Strawman PR 

https://github.com/timescale/pgai/blob/main/CONTRIBUTING.md#rfc-stages

Do you think you could explain more the solution to the feature? Considerations you've made? Something like:

- Are you creating new tables? (I saw the batches table briefly)
- How are you changing the vectorizer to deal with batches? For example, on every loop of the vectorizer a new batch is created, the batch is stored in the new table.
- When are the batches retrieved?
- What happens on errors?

I'm already reviewing and leaving some feedback.

projects/extension/sql/ai--0.6.0.sql
```diff
@@ -0,0 +1,403 @@
...
+        scheduling = pg_catalog.jsonb_insert(scheduling, array['job_id'], to_jsonb(_job_id));
+    end if;
+
+    embedding_batch_schema = coalesce(embedding_batch_schema, 'ai');
+    embedding_batch_table = coalesce(embedding_batch_table, pg_catalog.concat('_vectorizer_embedding_batches_', _vectorizer_id));
+    embedding_batch_chunks_table = coalesce(embedding_batch_chunks_table, pg_catalog.concat('_vectorizer_embedding_batch_chunks_', _vectorizer_id));
+
+    -- create batch embedding tables
+    select (embedding operator (pg_catalog.->> 'implementation'))::text into _implementation;
+    if _implementation = 'openai' then
```
alejandrodnm, line 310: This needs to be created only if the embedding implementation the user set is for openai batch. We have, I think 2 options. Add a new boolean attribute to the current `ai.embedding_openai` like `batch_api=true`, or `use_batch_api=true`. The other is to create a new embedding configuration for openai batch api.

The benefit I see of creating a new configuration, is that it'll be easier to handle the different logics. Also, both APIs don't support the exact same options, for example, the batch API doesn't support the `user` option and the non-batch doesn't require a `completion_window`.

You can create a new implementation and can call it `ai.embedding_openai_batch`. You can use the `ai.embedding_openai` one for reference:

https://github.com/timescale/pgai/blob/59c32947972bea851193a6a382006c169eed1969/projects/extension/sql/idempotent/008-embedding.sql?5C#L3-L22

You can create it in the same file.

alejandrodnm (changes requested):
This is an initial pass on the sql code.

I'm not a SQL or extensions experts. This will have to be re-reviewed by my other colleagues. 

I haven't reviewed the python code. But, if you add a new openai batch implementation, you can create a new `OpenAIBatch` embedder in `projects/pgai/pgai/vectorizer/embeddings.py`, and let that drive the batching logic.

I haven't review that part, I'm guessing we'll need something else to trigger the fetch of the batches. 

Can you explain how do you plan to handle that part? In the same loop? Another worker?

kolaente:
@alejandrodnm Thanks for the review! Let me outline the the solution more generally:

I've discovered the problem I'm solving here in an application we're building, which creates embeddings from a lot of documents. The documents are basically everything the company has ever done, around 3m text documents. We were ingesting so much data into the database that we'd hit OpenAI's rate limit. It would have taken multiple months just with the limits of OpenAI's regular embeddings api. To solve this, I've implemented a system like the one in this PR, which batches the changes and then uses OpenAI's batch api to create embeddings. Because that has significantly higher rate limits, this would be a lot faster, even though a batch might take 24h to complete.
I've built this into my Laravel application and it seems to work. Now, I want to migrate everything over to pgai as I can then retire the chunking service that I've built, which is another part of the application.

The implementation here is a (almost) 1:1 translation of the one I implemented in my Laravel application. The main difference is the trigger - here it should automatically start to use the batch api instead of requiring a manual trigger (in my Laravel app, this was a cli command which would submit everything that has no embeddings yet to the batch api).

I've tried my best in creating a diagram to illustrate how I think it should work:

image

kolaente:
> I haven't reviewed the python code. But, if you add a new openai batch implementation, you can create a new OpenAIBatch embedder in projects/pgai/pgai/vectorizer/embeddings.py, and let that drive the batching logic.

AFAIK there can (or should) be only one vectorizer per table? Because then if I implement the batch processing as another vectorizer, the "fallback after hitting OpenAI's rate limit" will not work.

> I haven't review that part, I'm guessing we'll need something else to trigger the fetch of the batches.
> 
> Can you explain how do you plan to handle that part? In the same loop? Another worker?

This should be a seperate loop, but could also be included in the openai worker main loop.

projects/extension/sql/ai--0.6.0.sql
```diff
@@ -0,0 +1,403 @@
...
+, grant_to name[]
+) returns void as
+$func$
+declare
+    _sql text;
+begin
+    -- create the batches table
+    select pg_catalog.format
+    ( $sql$create table %I.%I(
+    openai_batch_id VARCHAR(255) PRIMARY KEY,
```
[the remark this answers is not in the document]
kolaente: The ID I'm using here is the one that's returned by OpenAI, that's why I called it that way. Renaming won't be a problem, but how can I make it clear then this ID is "owned" by OpenAI?

I'll have to take another look, but I think OpenAI is the only provider with a batch api for embeddings.

projects/extension/sql/ai--0.6.0.sql
```diff
@@ -0,0 +1,403 @@
...
+    end if;
+end;
+$func$
+language plpgsql volatile security invoker
+set search_path to pg_catalog, pg_temp
+;
+
+-------------------------------------------------------------------------------
+-- create_vectorizer
+create or replace function ai.create_vectorizer
```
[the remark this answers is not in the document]
kolaente: Done now.

projects/extension/sql/ai--0.6.0.sql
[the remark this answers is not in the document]
kolaente: Took another look and it seems like [cohere is the only other currently supported provider](https://docs.cohere.com/v2/docs/embed-jobs-api) which supports batch embedding jobs like OpenAI's. I would assume there will be others in the future at some point.

Unsure how to proceed here?

projects/extension/sql/ai--0.6.0.sql
```diff
@@ -0,0 +1,403 @@
+-------------------------------------------------------------------------------
+-- _vectorizer_create_queue_table
+create or replace function ai._vectorizer_create_embedding_batches_table
```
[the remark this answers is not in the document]
kolaente: Done.

kolaente:
> About files: I read that OpenAI keeps the files until they are manually deleted by the user. I’m not entirely sure about the implications of this since it’s not well-documented. If we decide to delete them manually, we need to store file IDs to make the extra requests for deletion.

Looks like it. Since [there's a limit on the total size of files in an org](https://platform.openai.com/docs/api-reference/files/create), we should delete the files once the batch is processed.

> Do we need to store all the [batch metadata](https://github.com/timescale/pgai/pull/280/files#diff-c13cec5f42c8cb178a30dd95012bb5ce152fe61cebf00f1b2f652e75365f1d71R17-R24)? I think storing just the batch ID and the last time it was checked is enough.

I think we should at least store the status, that way it is possible to check the status of in progress batches with one sql query. I've used that to get a quick overview in my application about this.

> The flow for sending chunks, generating batches, and saving batch IDs might replace the current process, though I might be missing something.

I would think of it more like extending the current process, but it's also possible to model it separately.

> The array index could represent the chunk sequence.

The problem with this approach is [the output order is not guaranteed](https://platform.openai.com/docs/guides/batch/getting-started?lang=node#5-retrieving-the-results). We should not rely on the order and only use the `custom_id`. 

> Filtering out oversized chunks for retries, as we do now, might become too complicated.

I think we can circumvent this by leaving well enough margin to the batch limit and send more batches instead.

> The only potential issue I see is in systems with continuous data ingestion, where the regular queue might never fully empty, delaying batch processing.

If the regular queue is never fully empty, doesn't that indicate some kind of backstop?

> @kolaente mentioned dynamically switching between non-batch and batch APIs based on rate limits. Is this something we want to support? It might add significant complexity.

This is my original use-case, that's why I proposed this. I'm unsure about moving the implementation to a separate vectorizer because then it's an either or decision. Ideally, I'd like to use this for backfilling large amounts of data and use the current implementation for everything else.

smoya:
> I think we should at least store the status, that way it is possible to check the status of in progress batches with one sql query. I've used that to get a quick overview in my application about this.

What would all the status be? Correct me if I'm wrong but from how I see it, that column will report the **last** status but not the real up-to-date status. Wondering if we really need to have such info. WDYT?

alejandrodnm:
> > I think we should at least store the status, that way it is possible to check the status of in progress batches with one sql query. I've used that to get a quick overview in my application about this.
> 
> What would all the status be? Correct me if I'm wrong but from how I see it, that column will report the **last** status but not the real up-to-date status. Wondering if we really need to have such info. WDYT?

I see a use for this. When going over the queue of batches to import we can have a flow like this:

- Lock the row with `select for update skip locked`, and update a `last_checked`, or `delay` column in the row. This will make it concurrently safe, and by updating the field, we increase the delay so if the import fails, or the batch is not ready we don't check it again until significant time has passed.
- Create a save point. This save point allows the import to fail, but maintain the `delay` of the transaction.
- Import the file and update or not the status to `done`.
- Create a save point. This save point allows to keep the done status, so that if deleting the file fails, or we exit at this point, we can just retry the file deletion, and not re-importing.
- Delete the file.
- Delete the item from the queue.

@kolaente Do you need the list of files/batches you already imported? Once a batch is imported and the file deleted, having them still in the DB might not give you any kind of value. The important thing is the embeddings, and you can find which items in your source queue have not been embedded.

You'd still have 2 queue, the regular embedding queue, which tells you the items that need to be embedded. The other is the batches queue, which will list, pending batches, or batches that have been imported but the files have not been deleted. 

Ideally, if a batch doesn't have a problem, you won't see it as done in the queue.
---
To support the use case of switching seamlessly between batch and non-batch API lets go with extending the existing embeddings configuration.

I'll leave out some pointers in the PR diff, because there's something else I want you to move there. 

With this approach, you'd just have to run an `UPDATE` statement to update the vectorizer config to switch between the 2 APIs. Is that good enough for your use case? I'm a little hesitant about auto fallback to batch API.

--- 

> The problem with this approach is [the output order is not guaranteed](https://platform.openai.com/docs/guides/batch/getting-started?lang=node#5-retrieving-the-results). We should not rely on the order and only use the custom_id.

That docs says that the order of the requests is not guaranteed. Each request will have a custom_id, that's correct. Your implementation matches one request to a chunk, and that's also correct. My suggestion, is to reduce the number of requests we create in the file, instead of one request per chunk, we can do one request per document, making the chunk_id equal to the pk.

In an embed request (not the batch file) you can send a string (or list of tokens), or a list of strings (or list of list of tokens). When you use the list of strings (or list of list of tokens) the response returns a list of the same size with the embeddings, the positions of the items in the response (embeddings) match the position of the items send in the `input`. [See this forum discussion](https://community.openai.com/t/does-the-index-field-on-an-embedding-response-correlate-to-the-index-of-the-input-text-it-was-generated-from/526099). This is kind of minor, we can start with what you did, the limits are the same either way, maybe less requests can be processed faster by openAI, but I don't know.
---

> If the regular queue is never fully empty, doesn't that indicate some kind of backstop?

I don't know if it would be a real case scenario, just wanted to pointed out. Your implementation waits for all the batches to be created (no more items in the regular queue), before start polling for finished batches. If you have a system that keeps adding items in the queue (continuous insert/updates), you'll be forever creating batches and never processing them. Again, this might not be a real scenario, and we expect the batch API to be used for backfilling. We can leave as it is, and come back to it later if there's any issues.

---

I'm going to do another pass, now that we've settled on some things, and understand the problem better.  It'll mostly be:

- Extending the openAI embeddings configuration with the batch properties.
- Handling the batch queue in a concurrently safe manner. Adding save points that allow us to recover in case of failures.

kolaente:
> we increase the delay so if the import fails, or the batch is not ready we don't check it again until significant time has passed.

What do you think is significant? From my experience, most batches complete faster than the 24h window, so it would make sense to check a lot more often.

> @kolaente Do you need the list of files/batches you already imported? Once a batch is imported and the file deleted, having them still in the DB might not give you any kind of value. The important thing is the embeddings, and you can find which items in your source queue have not been embedded.

I don't think there's a need for having old batch data once the embeddings are stored.

> With this approach, you'd just have to run an UPDATE statement to update the vectorizer config to switch between the 2 APIs. Is that good enough for your use case? I'm a little hesitant about auto fallback to batch API.

It will require a little more plumbing on my side, but probably work for the use-case yeah.

What makes you hesitant about auto fallback?

> instead of one request per chunk, we can do one request per document, making the chunk_id equal to the pk.

Gotcha. I didn't know the embeddings api could handle more than one input string in a single request.

> This is kind of minor, we can start with what you did, the limits are the same either way, maybe less requests can be processed faster by openAI, but I don't know.

Both is fine by me. Which of the two would increase the chances of this getting merged?

> Your implementation waits for all the batches to be created (no more items in the regular queue), before start polling for finished batches. If you have a system that keeps adding items in the queue (continuous insert/updates), you'll be forever creating batches and never processing them.

Ahhhh yes, that should be changed. My intention was to check batches in every run of the worker, but only after the current embedding run was done and processed. How would I add this? Seems like I didn't understand the code good enough.

projects/extension/sql/idempotent/013-vectorizer-api.sql
```diff
@@ -31,6 +29,9 @@ create or replace function ai.create_vectorizer
 , queue_table pg_catalog.name default null
 , grant_to pg_catalog.name[] default ai.grant_to()
 , enqueue_existing pg_catalog.bool default true
+, embedding_batch_schema pg_catalog.name default null
+, embedding_batch_table pg_catalog.name default null
+, embedding_batch_chunks_table pg_catalog.name default null
```
alejandrodnm, line 34: Let's move this arguments to 

https://github.com/timescale/pgai/blob/59c32947972bea851193a6a382006c169eed1969/projects/extension/sql/idempotent/008-embedding.sql#L4-L22

Let's also add another attribute to the  json returned by `ai.embedding_openai` called `use_batch_api` that'll take a boolean.

That way the batch logic can be contain in the openAI embedding, since it's the only one that uses it. Users will enable it like:

```sql
select ai.create_vectorizer(
    'public.blogs'::regclass
  , embedding=>ai.embedding_openai(
      'text-embedding-3-small', 
      1536, 
      api_key_name=>'OPENAI_API_KEY', 
      use_batch_api=>true)
  , chunking=>ai.chunking_recursive_character_text_splitter('content')
  , formatting=>ai.formatting_python_template('title: $title url: $url content: $chunk')
);
```

projects/extension/sql/idempotent/013-vectorizer-api.sql
```diff
@@ -225,6 +227,31 @@ begin
         scheduling = pg_catalog.jsonb_insert(scheduling, array['job_id'], pg_catalog.to_jsonb(_job_id));
     end if;
 
+    embedding_batch_schema = coalesce(embedding_batch_schema, 'ai');
+    embedding_batch_table = coalesce(embedding_batch_table, pg_catalog.concat('_vectorizer_embedding_batches_', _vectorizer_id));
+    embedding_batch_chunks_table = coalesce(embedding_batch_chunks_table, pg_catalog.concat('_vectorizer_embedding_batch_chunks_', _vectorizer_id));
```
alejandrodnm, line 232: These should be moved also to:

https://github.com/timescale/pgai/blob/59c32947972bea851193a6a382006c169eed1969/projects/extension/sql/idempotent/008-embedding.sql#L4-L22

projects/extension/sql/idempotent/013-vectorizer-api.sql
```diff
@@ -225,6 +227,31 @@ begin
...
+        if pg_catalog.to_regclass(pg_catalog.format('%I.%I', embedding_batch_schema, embedding_batch_chunks_table)) is not null then
+            raise exception 'an object named %.% already exists. specify an alternate embedding_batch_chunks_table explicitly', queue_schema, queue_table;
+        end if;
+
+        perform ai._vectorizer_create_embedding_batches_table
+                (embedding_batch_schema
+            , embedding_batch_table
+            , embedding_batch_chunks_table
+            , grant_to
+                );
```
alejandrodnm, line 252: Let's move this logic to its own function. It'll receive the embedding variable as argument, and it'll retrieve the schema and name from it.

projects/extension/sql/idempotent/013-vectorizer-api.sql
```diff
@@ -259,6 +286,9 @@ begin
       , 'formatting', formatting
       , 'scheduling', scheduling
       , 'processing', processing
+      , 'embedding_batch_schema', embedding_batch_schema
+      , 'embedding_batch_table', embedding_batch_table
+      , 'embedding_batch_chunks_table', embedding_batch_chunks_table
```
alejandrodnm, line 291: We won't need this because they will be stored in the `embedding` variable.

projects/extension/sql/idempotent/016-openai-batch-api.sql
```diff
@@ -0,0 +1,96 @@
...
+           ) into strict _sql
+    ;
+    execute _sql;
+
+    -- create the batch chunks table
+    select pg_catalog.format
+           ( $sql$create table %I.%I(
+    id                 VARCHAR(255) PRIMARY KEY,
+    embedding_batch_id VARCHAR(255) REFERENCES %I.%I (openai_batch_id),
+    text               TEXT
```
alejandrodnm, line 45:
```suggestion
    chunk              TEXT NOT NULL
```

That's what we use in the store table, since this is an intermediate storage, let's use the same name.

alejandrodnm (changes requested):
Based on the previous discussion. We are going to add an `use_batch_api` attribute to `ai.embedding_openai`, these changes should allow us to easily change between the 2 APIs with something like:

```sql
UPDATE ai.vectorizer
 SET config = jsonb_set(
     config,
     '{embedding,use_batch_api}',
     'false',
     false
 )
 WHERE id = 1;
```

This review only covers the extension part. I'll do another for the python code. 

There's a part missing, which is updating existing vectorizers. I think we can tackle that in a different PR.

alejandrodnm (changes requested):
Second review, I need to stop a little for now. I'll try to continue later today.

projects/extension/sql/idempotent/013-vectorizer-api.sql
kolaente: Doesn't it make sense to move this to `ai._validate_embedding`?

projects/extension/sql/idempotent/013-vectorizer-api.sql
kolaente: Done.

projects/extension/sql/idempotent/013-vectorizer-api.sql
kolaente: Done.

projects/extension/sql/idempotent/013-vectorizer-api.sql
kolaente: Done.

projects/extension/sql/idempotent/016-openai-batch-api.sql
kolaente: Done as well

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -81,6 +87,9 @@ class Config:
         LangChainCharacterTextSplitter | LangChainRecursiveCharacterTextSplitter
     ) = Field(..., discriminator="implementation")
     formatting: PythonTemplate | ChunkValue = Field(..., discriminator="implementation")
+    embedding_batch_schema: str | None
+    embedding_batch_table: str | None
+    embedding_batch_chunks_table: str | None
```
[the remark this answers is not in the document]
kolaente: Done.

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -319,6 +328,89 @@ def insert_errors_query(self) -> sql.Composed:
             self.errors_table_ident,
         )
 
+    @cached_property
+    def fetch_batches_to_process_query(self) -> sql.Composed:
+        return sql.SQL("""
+            SELECT openai_batch_id, output_file_id FROM {}.{}
+            WHERE status not in('failed', 'processed', 'prepared')
```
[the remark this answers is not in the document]
kolaente: This makes sense.

Does that also mean changing the implementation so that only in-progress batches are stored in the database, getting rid of storing and evaluating the status? I noticed your query does not contain the `WHERE status not in('failed', 'processed', 'prepared')` anymore.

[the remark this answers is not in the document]
kolaente: I've now implemented the changes to the query as per your suggestion.

alejandrodnm (changes requested):

alejandrodnm:
> > There's a part missing, which is updating existing vectorizers. I think we can tackle that in a different PR.
> 
> If we add an attribute to use the batch embedding api, do we really need to update existing vectorizers? Can't we just evaluate if the attribute exists and fall back to false if it does not exist?

My point here is what would happen for an existing vectorizer that would like to change to the async API once the feature is release. But let's not worry about that for now.

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -466,13 +552,117 @@ async def run(self) -> int:
             await self.vectorizer.config.embedding.setup()
             while True:
                 if not self._continue_processing(loops, res):
+                    await self._check_and_process_openai_batches(conn)
```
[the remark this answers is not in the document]
kolaente: Sounds good! (and a lot better than what I originally cooked up)

> Mark the batch job as imported, or something like that. The point of this status is to signal that we imported the embeddings, but we haven't cleaned up the file. Maybe there was an error the last time we tried.

You're talking about cleaning up the file stored in openai here?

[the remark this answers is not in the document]
kolaente: I've now implemented this, please take another look.

I also added a new type `AsyncBatch` to make passing the data from the database around a little clearer wrt what to expect.
The implementation seems to have cleared a few items from my checklist (first comment here) as well.

kolaente:
With that being done, I'll be pretty much offline for the next two weeks (Jan 6th) - happy holidays etc :) so please don't expect me to respond quickly.

alejandrodnm:
> With that being done, I'll be pretty much offline for the next two weeks (Jan 6th) - happy holidays etc :) so please don't expect me to respond quickly.

I'm also out for the holidays. I'll try reviewing on the 7th when I get back.

Thanks for your effort. We really appreciate it.

kolaente:
Hey @alejandrodnm, I'll probably do another round on this next week. I'll add some tests as you outlined.

How would I trigger the errors, using mocks in the openai client?

alejandrodnm:
Hey @kolaente That's an option. Another could be to set up the vcr cassette to return a 500 http error.

alejandrodnm:
Hey @kolaente how are you doing? Is there anything I can do to help you finish the PR?

I can also take over it if you want. I have some bandwidth and you've already done most of the work.

Cheers.

kolaente:
Hey @alejandrodnm, if you have the bandwidth, I would appreciate some help with the tests, I've never really done these kinds of tests in python before. And do you have some input on the three open points in the PR description?

I plan to make some time for this next week as well.

alejandrodnm:
Superseded by https://github.com/timescale/pgai/pull/564

PR #281, opened 2024-12-06
docs: mention the option to host vectorizer even with timescale cloud

Askir:


README.md
```diff
@@ -207,7 +207,11 @@ to implement embeddings in your own data. On a self-hosted Postgres
 installation, you use a [Vectorizer Worker](/docs/vectorizer-worker.md) to
 asynchronously processes your vectorizers. When you create Vectorizers in a
 Timescale Cloud database, embeddings are automatically created and synchronized
-in the background. Note: Timescale Cloud only supports embedding with OpenAI.
+in the background.
+
+Note: Timescale Cloud currently only supports embedding with OpenAI natively.
+If you want to use another provider (e.g. ollama) you can however disable [scheduling](/docs/vectorizer-api-reference.md#scheduling-configuration)
+and run the vectorizer worker yourself.
```
JamesGuthrie, line 214: IMHO we should
1. Frame this around Ollama specifically
2. Show how to do this inline

```suggestion
Note: Timescale Cloud currently only supports embedding with OpenAI.
If you want to use a self-hosted Ollama API, you can disable [scheduling](/docs/vectorizer-api-reference.md#scheduling-configuration) with `scheduling => ai.scheduling_none()`
and [self-host](/docs/vectorizer-worker.md#install-and-configure-vectorizer-worker) the vectorizer worker.
```

Askir: fair about ollama. Doing it inline, I'm not so sure about. I feel like it breaks with the high level flow of the Readme a lot (I honestly don't even like this detailed comment here so much but I don't know where else to put it)

billy-the-fish: Really good point. I would be careful with the words self-hosted, in the docs that means a self-hosted deployment of timescaledb in your own postgres deployment.  This is what I understand your sentence to mean. 

Note: Timescale Cloud currently supports embedding natively with OpenAI. To use Ollama on the data in your Timescale Cloud service,  set [scheduling => ai.scheduling_none()](/docs/vectorizer-api-reference.md#scheduling-configuration) in the configuration for your service, then [install vectorizer worker locally](/docs/vectorizer-worker.md#install-and-configure-vectorizer-worker) and configure it to connect to your Timescale Cloud service.

Askir: That sounds great to me I changed it 👍

billy-the-fish (approved):
LGTM

PR #284, opened 2024-12-09
feat!: remove `truncate` parameter from Ollama/Voyage APIs

JamesGuthrie:
Removes the optional `truncate` parameter from the following functions:
- ai.embedding_ollama
- ai.embedding_voyageai
- ai.voyageai_embed

Truncation is on by default for Ollama and Voyage AI APIs, it automatically truncates inputs which are larger than the embedding model's context window. The `truncate` parameter provided the ability to opt out of this behaviour. Vector embeddings would fail to be produced for a document with chunks larger than the model's context window.

Our judgement is that this provided too much flexibility and opportunity for error/misconfiguration, so we have chosen to remove this option from the API.


alejandrodnm (approved):

Askir (approved):
-300 lines looks glorious 🚀 
Let's hope we don't have to add them back 😁

PR #293, opened 2024-12-12
docs: fix psql command and pstgres data volume in quickstarts

Askir:


docs/vectorizer-quick-start.md
```diff
@@ -53,7 +53,7 @@ Now we can create and run a vectorizer. A vectorizer is a pgai concept, it proce
 
 1. **Connect to the database in your local developer environment**
 
-   - Docker: `docker compose exec -it pgai psql`
+   - Docker: `docker exec -it pgai-db-1 psql -U postgres`
```
JamesGuthrie, line 56: Why did you change this back?

Askir: The original command was wrong it would need to be `-it db` not `-it pgai`. Hervè called this out.
The other two quickstarts simply use `docker exec` instead of `docker compose exec` though, so I just aligned them to that.

I don't really have any preference.

JamesGuthrie: Oof, ya you're right. I think that might have been me 🙈

JamesGuthrie: I have a slight preference for `docker compose` because you can just use the service name (without the `pgai-` and `-` prefixes and suffixes).

JamesGuthrie (approved):

docs/vectorizer-quick-start.md
Askir: alright fair, changed it :D

PR #309, opened 2024-12-16
chore: support uv in extension install, use for dev

JamesGuthrie:
We would like to support uv in an opt-in fashion. The `build.py install` now uses uv if it is available, otherwise falling back to pip. To use uv, we now install it in our dev containers.


JamesGuthrie:
Note: reduces install time from 1 min 30 seconds to 22 seconds.

Before:

Screenshot 2024-12-16 at 15 39 16

After:

Screenshot 2024-12-16 at 15 39 03

PR #311, opened 2024-12-16
feat: load api keys from db in self hosted vectorizer

Askir:


docs/vectorizer-quick-start-openai.md
```diff
@@ -32,7 +32,6 @@ On your local machine:
         image: timescale/pgai-vectorizer-worker:v0.2.1
         environment:
           PGAI_VECTORIZER_WORKER_DB_URL: postgres://postgres:postgres@db:5432/postgres
-          OPENAI_API_KEY: <your-api-key>
```
Askir, line 35: While this will work with the new version, this is actually going to be confusing until we push a new release. Unless we change this to refer to main for the vectorizer worker? Is that image kept up to date on each `main` push?

projects/pgai/pgai/cli.py
```diff
@@ -102,8 +102,15 @@ def get_vectorizer(db_url: str, vectorizer_id: int) -> Vectorizer:
         # The Ollama API doesn't need a key, so `api_key_name` may be unset
         if "api_key_name" in embedding:
             api_key_name = embedding["api_key_name"]
-            api_key = os.getenv(api_key_name, None)
+            cur.execute(
```
smoya, line 105: As a matter of override, wouldn't make sense to follow just the opposite flow? In particular, consider the env var always as top priority and, if not set, then pull from DB.

Askir: yeh you're right that makes more sense, thanks :D

docs/vectorizer-quick-start-openai.md
cevian: yeah we should wait to merge this until this is resolved

projects/pgai/pgai/cli.py
```diff
@@ -104,6 +104,13 @@ def get_vectorizer(db_url: str, vectorizer_id: int) -> Vectorizer:
             api_key_name = embedding["api_key_name"]
             api_key = os.getenv(api_key_name, None)
             if api_key is None:
+                cur.execute(
```
cevian, line 107: LGTM

cevian (commented):
The code changes look good. Leaving the eval of the testing to other reviewers

projects/pgai/pgai/cli.py
JamesGuthrie, line 107: Could you add some (debug?) logging so that we can see where the vectorizer got the secret from? Something like this diff:

```diff
diff --git a/projects/pgai/pgai/cli.py b/projects/pgai/pgai/cli.py
index c890dff..fedccbb 100644
--- a/projects/pgai/pgai/cli.py
+++ b/projects/pgai/pgai/cli.py
@@ -103,13 +103,17 @@ def get_vectorizer(db_url: str, vectorizer_id: int) -> Vectorizer:
         if "api_key_name" in embedding:
             api_key_name = embedding["api_key_name"]
             api_key = os.getenv(api_key_name, None)
-            if api_key is None:
+            if api_key is not None:
+                log.debug(f"obtained secret '{api_key_name}' from environment")
+            else:
                 cur.execute(
                     "select ai.reveal_secret(%s)",
                     (api_key_name,),
                 )
                 row = cur.fetchone()
                 api_key = row["reveal_secret"] if row is not None else None
+                if api_key is not None:
+                    log.debug(f"obtained secret '{api_key_name}' from database")
             if not api_key:
                 raise ApiKeyNotFoundError(
                     f"api_key_name={api_key_name} vectorizer_id={vectorizer_id}"
```

JamesGuthrie (approved):
LGTM, one minor comment

docs/vectorizer-quick-start-openai.md
Askir: I'll put this in a separate PR

PR #313, opened 2024-12-17
feat: text-to-sql

jgpruitt:


projects/extension/sql/idempotent/905-text-to-sql.sql
```diff
@@ -0,0 +1,510 @@
...
+    , 'openai_user': openai_user
+    absent on null
+    )
+$func$ language sql immutable security invoker
+set search_path to pg_catalog, pg_temp
+;
+
+-------------------------------------------------------------------------------
+-- text_to_sql_ollama
+create or replace function ai.text_to_sql_ollama
```
cevian, line 352: append _config to the names?
```suggestion
create or replace function ai.text_to_sql_ollama_config
```

projects/extension/sql/idempotent/905-text-to-sql.sql
```diff
@@ -0,0 +1,510 @@
...
+    , 'top_p': top_p
+    absent on null
+    )
+$func$ language sql immutable security invoker
+set search_path to pg_catalog, pg_temp
+;
+
+-------------------------------------------------------------------------------
+-- text_to_sql
+create or replace function ai.text_to_sql
```
cevian, line 410: ugh this seem so repetitive and redundant to functions like ai.openai_chat_complete. Maybe we just have function to generate the full prompt and have people call the chat_complete functions explicitly?

projects/extension/sql/idempotent/905-text-to-sql.sql
```diff
@@ -0,0 +1,510 @@
...
+
+    return _ddl;
+end
+$func$ language plpgsql stable security invoker
+set search_path to pg_catalog, pg_temp
+;
+
+-------------------------------------------------------------------------------
+-- _text_to_sql_prompt
+create or replace function ai._text_to_sql_prompt
```
cevian, line 96: we should think about how to make these functions more composable and customizable. This feels too canned for me.

cevian (changes requested):
I think we should discuss a bit more about what the  devx/function calling structure looks like

cevian (approved):
Approving so we can checkpoint - but we need to address devx issues before taking  it out of feature flag

PR #315, opened 2024-12-17
chore: split embedders in individual files

smoya:
This PR refactors the existing `projects/pgai/pgai/embeddings.py` file by splitting it into smaller files, each containing a single class for a specific embedding provider (atm: voyageai, openai, ollama). 

The primary goal is to improve maintainability and prevent the file from growing indefinitely over time, which can make it harder to manage and navigate.


projects/pgai/pgai/vectorizer/embeddings.py
```diff
@@ -1,37 +1,15 @@
...
-from typing_extensions import TypedDict, override
-
-MAX_RETRIES = 3
-
-TOKEN_CONTEXT_LENGTH_ERROR = "chunk exceeds model context length"
 
-openai_token_length_regex = re.compile(
-    r"This model's maximum context length is (\d+) tokens"
-)
+# Import needed to make the embedders available under this module, avoiding any breaking change after the split refactor.
```
smoya, line 11: Any idea on how to avoid this? My idea is to not add a breaking change with the splitting (the import paths for the embedders are changing).

projects/pgai/pgai/vectorizer/embeddings.py
```diff
@@ -1,37 +1,16 @@
...
-MAX_RETRIES = 3
-
-TOKEN_CONTEXT_LENGTH_ERROR = "chunk exceeds model context length"
 
-openai_token_length_regex = re.compile(
-    r"This model's maximum context length is (\d+) tokens"
-)
+# Import needed to make the embedders available under this module, 
+# avoiding any breaking change after the split refactor.
+from .embedders import *  # noqa: F403
```
smoya, line 13: Any idea on how to avoid this? My idea is to avoid adding a breaking change with this refactor as the import paths for the embedders are changing. 
If I don't add this import here, import paths for embedders such as Ollama will need to change from `from .embeddings import Ollama` to `from .embedders import Ollama`., causing a breaking change (and the need of releasing a major version)

projects/pgai/pgai/vectorizer/embeddings.py
```diff
@@ -1,37 +1,16 @@
...
-MAX_RETRIES = 3
-
-TOKEN_CONTEXT_LENGTH_ERROR = "chunk exceeds model context length"
 
-openai_token_length_regex = re.compile(
-    r"This model's maximum context length is (\d+) tokens"
-)
+# Import needed to make the embedders available under this module,
+# avoiding any breaking change after the split refactor.
+from .embedders import *  # noqa: F403
```
smoya, line 13: Any idea on how to avoid this? My idea is to avoid adding a breaking change with this refactor as the import paths for the embedders are changing.
If I don't add this import here, import paths for embedders such as Ollama will need to change from from .embeddings import Ollama to from .embedders import Ollama., causing a breaking change (and the need of releasing a major version)

projects/pgai/pgai/vectorizer/embedders/voyageai.py
```diff
@@ -0,0 +1,72 @@
+from collections.abc import Sequence
+from functools import cached_property
+from typing import Literal
+
+import voyageai
+import voyageai.error
+from pydantic import BaseModel
+from typing_extensions import override
+
+from ..embeddings import (
```
smoya, line 10: Class is unmodified, except for this required imports.

projects/pgai/pgai/vectorizer/embedders/openai.py
```diff
@@ -0,0 +1,202 @@
...
+from functools import cached_property
+from typing import Any, Literal
+
+import openai
+import tiktoken
+from openai import resources
+from pydantic import BaseModel
+from typing_extensions import override
+
+from ..embeddings import (
```
smoya, line 12: Class is unmodified, except for this required imports.

projects/pgai/pgai/vectorizer/embedders/ollama.py
```diff
@@ -0,0 +1,158 @@
...
+from typing import (
+    Any,
+    Literal,
+)
+
+import ollama
+from pydantic import BaseModel
+from typing_extensions import TypedDict, override
+
+from ..embeddings import (
```
smoya, line 13: Class is unmodified, except for this required imports.

cevian (approved):
LGTM but lets also have @JamesGuthrie review

projects/pgai/pgai/vectorizer/embedders/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .ollama import Ollama
+from .openai import OpenAI
+from .voyageai import VoyageAI
+
+__all__ = ["Ollama", "OpenAI", "VoyageAI"]
```
JamesGuthrie, line 5: Do we need this `__all__`? According to [this](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package):
>  if a package’s `__init__.py` code defines a list named `__all__`, it is taken to be the list of module names that should be imported when `from package import *` is encountered

This seems like somewhat esoteric functionality, which we're not actively using, and I don't see the need to support. By removing it we can remove one more instruction in the readme above.

smoya: I made this in order to be able to import all providers through the same module: `from .embedders import Ollama, OpenAI, VoyageAI`  The alternative (unless you know another one) would be to use individual imports such as:

```python
from .embedders.ollama import Ollama
from .embedders.openai import OpenAI
from .embedders.voyageai import VoyageAI
```

JamesGuthrie: Thanks for the explanation! I don't have a strong opinion, but the second suggestions (imports in `__init__.py`) is "less magical", so I would prefer it over `__all__`.

JamesGuthrie (approved):

projects/pgai/pgai/vectorizer/embedders/__init__.py
```diff
@@ -1,5 +1,3 @@
-from .ollama import Ollama
-from .openai import OpenAI
-from .voyageai import VoyageAI
-
-__all__ = ["Ollama", "OpenAI", "VoyageAI"]
+from .ollama import Ollama as Ollama
```
smoya, line 1: The redundant alias is needed, otherwise the linter will complain and provide the only alternative of adding it to the `__all__` var

PR #316, opened 2024-12-17
chore: build latest image on push

Askir:
The idea here:
- Keep `latest` always up-to-date with main (or call it `main`?)
- Don't break main if possible, fix it fast if it breaks (we've been doing that anyways)
- Refer to the new `latest` tag everywhere in the docs
- Mention that for production use people should use a specific version instead

Problems:
- What if the main version of the vectorizer worker becomes incompatible with the currently released pgai extension version? We don't have a `latest` in development timescale-ha image. Is this something we should plan for?

### DIFF
--- a/.github/workflows/build-worker.yml
+++ b/.github/workflows/build-worker.yml
@@ -1,0 +1,44 @@
+on:
+  push:
+    branches:
+      - main
+
+name: build-worker
+
+jobs:
+  build-worker:
+    name: Publish to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Docker meta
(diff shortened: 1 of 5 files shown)


JamesGuthrie (changes requested):
- I don't think that we should call this `latest`, which IMHO is more for "latest tagged release" I like `edge`, which says "this is sharp and might cut you".

- could we merge `build-worker.yml` into `release-please.yml`, and rename `release-please.yml`?

- You should be able to use the `docker/metadata-action` to get all of the different tags that we want with a well-crafted `tags` input (see https://github.com/docker/metadata-action?tab=readme-ov-file#tags-input).

Something like this:

```
tags: | 
  type=raw,value=${{ env.RELEASE_TAG }},enable=${{needs.release-please.outputs.pgai_release_created}}
  type=raw,value=latest,enable=${{needs.release-please.outputs.pgai_release_created}}
  type=edge
```

Askir:
okay it seems like I don't have a lot of support for this workflow so I'm gonna close this PR

PR #317, opened 2024-12-17
docs: remove openai mention, fix opclasses in hnsw index

Askir:
We actually don't allow l2 distance according to the code:
https://github.com/timescale/pgai/blob/main/projects/extension/sql/ai--0.6.0.sql#L2112 (although it would probably work if we'd change that)

The vectorizer quickstart is ollama focused on openai not sure how that snuck in there.

### DIFF
--- a/docs/vectorizer-api-reference.md
+++ b/docs/vectorizer-api-reference.md
@@ -602,7 +602,7 @@ HNSW is suitable for in-memory datasets and scenarios where query speed is cruci
 ```sql
   SELECT ai.create_vectorizer(
       'blog_posts'::regclass,
-      indexing => ai.indexing_hnsw(min_rows => 50000, opclass => 'vector_l2_ops'),
+      indexing => ai.indexing_hnsw(min_rows => 50000, opclass => 'vector_l1_ops'),
       -- other parameters...
   );
 ```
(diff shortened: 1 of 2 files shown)


docs/vectorizer-api-reference.md
```diff
@@ -602,7 +602,7 @@ HNSW is suitable for in-memory datasets and scenarios where query speed is cruci
 ```sql
   SELECT ai.create_vectorizer(
       'blog_posts'::regclass,
-      indexing => ai.indexing_hnsw(min_rows => 50000, opclass => 'vector_l2_ops'),
+      indexing => ai.indexing_hnsw(min_rows => 50000, opclass => 'vector_l1_ops'),
```
JamesGuthrie, line 605: @jgpruitt do you remember if there was a specific reason that we did not have `vector_l2_ops` in this list: https://github.com/timescale/pgai/blob/main/projects/extension/sql/ai--0.6.0.sql#L2112?

Askir: We had a slack discussion on this before (I just forgot to push the docs change afterwards and remembered now):
> I believe the last three are relatively recent additions to pgvector and probably didn’t exist when our code was first written

JamesGuthrie (approved):
I'm happy with the docs changes, but I would like to suggest that we also make a code change to accept the new vector ops types.

Askir:
@JamesGuthrie 
I created a linear issue for it. I doubt this has priority until someone asks for it though :D

PR #318, opened 2024-12-18
test: spawn the ollama container through testcontainers

smoya:
This PR spawns a new `ollama/ollama:latest` container through testcontainers when a test requires an ollama server to be up.

As a mechanism to override this behavior, the container won't be used if a `OLLAMA_HOST` env var is found, using the value of it instead.

Open to better alternatives.

### DIFF
--- a/projects/pgai/tests/vectorizer/conftest.py
+++ b/projects/pgai/tests/vectorizer/conftest.py
@@ -7,6 +7,7 @@ import tiktoken
 import vcr  # type:ignore
 from testcontainers.core.image import DockerImage  # type:ignore
 from testcontainers.postgres import PostgresContainer  # type:ignore
+from testcontainers.ollama import OllamaContainer  # type:ignore
 
 from pgai.vectorizer.vectorizer import TIKTOKEN_CACHE_DIR
 
@@ -70,3 +71,9 @@ def postgres_container():
         driver=None,
     ) as postgres:
         yield postgres
+
+@pytest.fixture(scope="session")
+def ollama_container():
+    with OllamaContainer(image="ollama/ollama:latest") as ollama:
+        yield ollama
+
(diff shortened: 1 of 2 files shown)


adolsalamanca (approved):
thanks for this!!

projects/pgai/tests/vectorizer/conftest.py
```diff
@@ -70,3 +71,9 @@ def postgres_container():
         driver=None,
     ) as postgres:
         yield postgres
+
+@pytest.fixture(scope="session")
+def ollama_container():
+    with OllamaContainer(image="ollama/ollama:latest") as ollama:
```
Askir, line 77: Could this use a volume mount to cache models?

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
```diff
@@ -151,19 +152,28 @@ def configured_ollama_vectorizer_id(
...
 ) -> int:
     """Creates and configures an ollama vectorizer for testing"""
     _, concurrency, batch_size, chunking, formatting = test_params
     _, conn = cli_db
 
+    base_url = ollama_container.get_endpoint()
+    if "OLLAMA_HOST" in os.environ: 
+        # The user is running it's own Ollama container
+        base_url = os.environ["OLLAMA_HOST"]
+        ollama_container.stop()
```
Askir, line 165: Could move this to the fixture I think. e.g. just call the fixture `ollama_url` and only start the container if the environment variable isn't set.

Askir (approved):
minor comments lgtm

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
smoya: You mean moving this logic to the `ollama_container` fixture?

smoya: Fixed!

projects/pgai/tests/vectorizer/conftest.py
smoya: Done. 

```python
# Passing the ollama_home lets us reuse models that have already
# been pulled to the `~/.ollama` path on the host machine.
ollama_home=Path.home() / ".ollama",
```

PR #320, opened 2024-12-18
WIP litellm integration

JamesGuthrie:


projects/extension/tests/test_litellm.py
```diff
@@ -0,0 +1,311 @@
...
+        """
+        select vector_dims
+        (
+            ai.litellm_embed
+            ( 'amazon.titan-embed-text-v1'
+            , 'hello world'
+            , extra_options => jsonb_build_object(
+                'aws_access_key_id', %s::text,
+                'aws_secret_access_key', %s::text,
+                'aws_region_name', %s::text
```
Askir, line 275: Is there any way to autoload these from env vars if they exist? same for the api_key

projects/extension/ai/secrets.py
```diff
@@ -24,27 +24,21 @@ def remove_secret_from_cache(sd_cache: dict[str, str], secret_name: str):
...
 
     if secret_name is None or secret_name == "":
-        plpy.error("secret_name is required")
-
-    secret = reveal_secret(plpy, secret_name, sd_cache)
-    if secret is None:
-        plpy.error(f"missing {secret_name} secret")
-        # This line should never be reached, but it's here to make the type checker happy.
-        return ""
+        return None
```
jgpruitt, line 39: is there existing code that depends on this condition throwing an error?

JamesGuthrie: I don't know. I still need to do a full audit of all of the callers.

jgpruitt:
lookin' good

projects/extension/ai/secrets.py
JamesGuthrie: I've now audited the callers - from what I can tell none of our code depends on this throwing an error.

The motivation for this change is that with litellm there is no sensible "default secret name" that we could configure because every provider has different conventions for secret naming. This motivated the change on line 29 to allow for `secret_name_default` to be `None`.

projects/extension/tests/test_litellm.py
```diff
@@ -0,0 +1,311 @@
...
+    for input_type in input_types:
+        cur.execute(
+            """
+            select vector_dims
+            (
+                ai.litellm_embed
+                ( %s
+                , 'hello world'
+                , api_key=>%s
+                , extra_options => jsonb_build_object('input_type', %s::text)
```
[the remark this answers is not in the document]
JamesGuthrie: Unfortunately not.

projects/extension/tests/test_litellm.py
```diff
@@ -0,0 +1,311 @@
...
+    cur.execute(
+        f"select set_config('ai.{api_key_name.lower()}', %s, false) is not null",
+        (api_key_value,),
+    )
+    cur.execute(
+        """
+        select vector_dims
+        (
+            ai.litellm_embed
+            ( %s
```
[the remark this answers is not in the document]
JamesGuthrie: The model names are internally prefixed by the provider, e.g. `openai/text-embedding-3-small`.

projects/extension/ai/litellm.py
```diff
@@ -0,0 +1,35 @@
+import litellm
+from typing import Optional, Generator
+
+
+def embed(
+    model: str,
+    input: list[str],
+    api_key: str,
+    user: Optional[str] = None,
```
Askir, line 9: Is the extension still python 3.9? Otherwise we should be able to use union `| None` syntax no?
Bit of a nit, just confused.

projects/pgai/pgai/vectorizer/embedders/litellm.py
```diff
@@ -0,0 +1,132 @@
...
+            case "huggingface":
+                return 2048  # NOTE: There is not documented limit. In testing we got a response for a request with 10k (short) inputs.
+            case "mistral":
+                return 128  # FIXME: this is chosen somewhat arbitrarily. There is no limit on # chunks per batch, but there is a (low) limit on # tokens per batch (16384)
+            case "vertex":
+                return 250  # FIXME: this applies to `us-central-1` only (otherwise 5). Additionally there is a limit on # tokens per batch (20k) see https://cloud.google.com/vertex-ai/generative-ai/docs/embeddings/get-text-embeddings#get_text_embeddings_for_a_snippet_of_text
+            case "voyage":
+                return 128  # see https://docs.voyageai.com/reference/embeddings-api
+            case _:
+                raise UnknownProviderError(custom_llm_provider)
```
Askir, line 112: props for actually figuring all these out ngl I would've set it just to some small number and called it a day.

projects/pgai/pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "datadog_lambda>=6.9,<7.0",
     "pytimeparse>=1.1,<2.0",
     "voyageai>=0.3.1,<0.4.0",
+    "litellm>=1.57.5,<1.58.0",
```
Askir, line 28: any errors in 1.58? or just cautious?

projects/pgai/pgai/vectorizer/embedders/litellm.py
```diff
@@ -0,0 +1,132 @@
...
+        await logger.adebug(f"Chunks produced: {len(documents)}")
+        return await self._batcher.batch_chunks_and_embed(documents)
+
+    @cached_property
+    def _batcher(self) -> BatchApiCaller[StringDocument]:
+        return BatchApiCaller(self._max_chunks_per_batch(), self.call_embed_api)
+
+    @override
+    def _max_chunks_per_batch(self) -> int:
+        print(f"model: {self.model}")
```
Askir, line 92: Might want to make this a debug log or just remove it?

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
```diff
@@ -1035,3 +1047,112 @@ def test_vectorization_successful_with_null_contents(
...
+                "--concurrency",
+                str(concurrency),
+                "--log-level",
+                "debug",
+            ],
+            catch_exceptions=False,
+        )
+
+    if result.exception:
+        print(result.output)
```
Askir, line 1151: this is a bit annoying, I mean we set `catch_exception=False` why doesn't it just throw the original one 😅

Askir (approved):
Looked mostly at the vectorizer code but that looks pretty good to me 👍

projects/extension/ai/litellm.py
JamesGuthrie: We actually have a >= 3.10 in the pyproject.toml, so no. We have Optional[...] all over the place. I'd prefer that we address a change like that separately, and ensure that linting is configured correctly to enforce.

projects/pgai/pgai/vectorizer/embedders/litellm.py
JamesGuthrie: Thanks.

projects/pgai/pyproject.toml
JamesGuthrie: Actually I think when I started working on this 1.58 wasn't out yet. Will update.

projects/pgai/pgai/vectorizer/embedders/litellm.py
JamesGuthrie: Oof - I did not intend to add that. Thanks.

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
JamesGuthrie: A couple of things:

- I didn't know what the mechanism of `catch_exceptions=False` is. It basically says "if the CLI throws an exception which is not `SystemExit`, then re-throw that exception as-is".
- We've got try/except wrapping around the internals anyway, to ensure that the worker doesn't exit if an error occurs (because it's supposed to be a long-running thing). As a result, we should never throw anything that isn't `SystemExit`.

Askir: We don't really log the exceptions well either. I tried adding exc_info to our error log in the cli before to get the full stack trace, but somehow that makes the cli never finish, not sure what exactly the issue is there.
But the dev experience on the tests is a bit annoying I always end up running the debugger for this.

projects/extension/tests/test_litellm.py
JamesGuthrie: Yes, it does work like that. At least, somewhere in the bowels of LiteLLM it's happening. TBH I'm not sure if all providers are supported in the same way, but I assume so.

I could add a test which illustrates that it works that way (by passing the `AWS_*` env vars into the test's DB container). Unfortunately, with the way the tests are set up at the moment (there is one global container that all tests run against), I can't easily "show"/prove that the two different methods work.

projects/extension/ai/secrets.py
jgpruitt: Okay, if we don't pass a secret_name or a secret_name_default, we return None instead of throwing an error. Why is this change beneficial? Are there cases when LiteLLM can still function if the secret is None?

Can we emit a warning here before returning None? I imagine in the case of other providers it will help with debugging.

projects/extension/ai/secrets.py
```diff
@@ -24,27 +24,21 @@ def remove_secret_from_cache(sd_cache: dict[str, str], secret_name: str):
...
 
     if secret_name is None:
         secret_name = secret_name_default
 
     if secret_name is None or secret_name == "":
-        plpy.error("secret_name is required")
-
-    secret = reveal_secret(plpy, secret_name, sd_cache)
-    if secret is None:
-        plpy.error(f"missing {secret_name} secret")
```
jgpruitt, line 43: What is the motivation for this change? If must avoid throwing an error, I think we should still emit a warning.

projects/pgai/pyproject.toml
```diff
@@ -15,7 +15,7 @@ dependencies = [
     "langchain-openai>=0.1,<1.0",
     "langchain-text-splitters>=0.2,<1.0",
     "pydantic>=2.0,<3.0",
-    "ollama>=0.4.5,<0.5.0",
+    "ollama>=0.3.0,<0.5.0",
```
jgpruitt, line 18: is litellm requiring an older version of ollama?

projects/extension/sql/idempotent/017-litellm.sql
```diff
@@ -0,0 +1,58 @@
...
+as $python$
+    #ADD-PYTHON-LIB-DIR
+    import ai.litellm
+    import ai.secrets
+    options = {}
+    if extra_options is not None:
+        import json
+        options = {k: v for k, v in json.loads(extra_options).items()}
+
+    api_key_resolved = ai.secrets.get_secret(plpy, api_key, api_key_name, None, SD)
```
jgpruitt, line 20: Is an api_key always required? If so, we should abort before making the API call.

jgpruitt (changes requested):

projects/extension/sql/idempotent/017-litellm.sql
JamesGuthrie: No, it's not required.

projects/extension/ai/secrets.py
JamesGuthrie: Yes, LiteLLM functions fine if the secret is None. It can optionally be passed a secret. If no secret is provided, it attempts to internally resolve a secret on its own from "known" environment variables. One case in which we rely on this is bedrock, which needs three parameters (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_REGION`). Similar for Vertex.

projects/extension/ai/secrets.py
JamesGuthrie: We must avoid throwing an error. I'm okay with emitting something here, but I'm not convinced it should be a warning.

Two alternatives:
1. We could move the error to the caller, for callers who need the secret to be populated.
2. We could let the caller tell this function if it wants to error on missing secret or not (but tbh I would prefer option 1).

WDYT?

projects/pgai/pyproject.toml
JamesGuthrie: No, I think this was a result of bad merge conflict resolution. Will fix.

projects/extension/ai/secrets.py
jgpruitt: I'm okay with option 1

JamesGuthrie: I ended up taking hidden option 3: don't modify anything in `secrets.py` and handle this in the caller of `ai.secrets.get_secret`.

projects/extension/pyproject.toml
```diff
@@ -8,14 +8,16 @@ dynamic = ["version"]
 requires-python = ">=3.10"
 
 dependencies= [
-    "openai==1.44.0",
+    "openai==1.56.0",
```
cevian, line 11: why this change?

cevian (approved):
Nice one small comment

docs/vectorizer.md
```diff
@@ -45,11 +45,19 @@ textual, data analysis, and semantic search:
 
 ## Select an embedding provider and set up your API Keys
 
-Vectorizer supports the following vector embedding providers:
+Vectorizer supports the following vector embedding providers as first-party integrations:
 - [Ollama](https://ollama.com/)
 - [Voyage AI](https://www.voyageai.com/)
 - [OpenAI](https://openai.com/)
 
+Additionally, through the [LiteLLM](https://litellm.ai) provider we support:
```
cevian, line 53: can we add a a table of embedding providers to the front-page README (can do in separate PR).

projects/extension/pyproject.toml
JamesGuthrie: LiteLLM requires >=1.55.3
https://github.com/BerriAI/litellm/blob/a835baacfc5e4c64690ac2dc60f2015e91222b4d/pyproject.toml#L24

projects/extension/tests/test_litellm.py
```diff
@@ -0,0 +1,323 @@
...
+            secrets["AWS_ACCESS_KEY_ID"],
+            secrets["AWS_SECRET_ACCESS_KEY"],
+            secrets["AWS_REGION_NAME"],
+        ),
+    )
+    actual = cur.fetchone()[0]
+    assert actual == 1536
+
+
+@pytest.mark.skip(reason="testing vertex is hard")
```
jgpruitt, line 302: 😆

jgpruitt (approved):

PR #325, opened 2024-12-20
ci: install all dependencies through just pgai install

smoya:
- Updated the `just pgai install` to include the `--all-extras` flag.
- Modified the command for installing dependencies in GH actions to use `just pgai install`.

### DIFF
--- a/projects/pgai/justfile
+++ b/projects/pgai/justfile
@@ -25,7 +25,7 @@ build:
 
 # Install the wheel package locally
 install:
-	@uv sync
+	@uv sync --all-extras
 
 # Remove the installed pgai package
 uninstall:
--- a/projects/pgai/justfile
+++ b/projects/pgai/justfile
@@ -51,7 +51,8 @@ type-check:
 format:
 	@uv run ruff format --diff ./
 
-ci: lint type-check format test build
+# CI pipeline. Runs all recipes needed for ensuring the code is ready to be integrated. Triggered by GH Actions
+ci: install lint type-check format test build
 
 # Build Docker image with version tag
 docker-build:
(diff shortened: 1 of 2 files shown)


alejandrodnm (approved):

projects/pgai/justfile
```diff
@@ -25,7 +25,7 @@ build:
 
 # Install the wheel package locally
 install:
-	@uv sync
+	@uv sync --all-extras
```
JamesGuthrie, line 28: I was wondering if this is strictly correct, but AFAICT we don't have any optional dependencies at the moment anyway, so it's kind of moot.

JamesGuthrie (approved):

projects/pgai/justfile
alejandrodnm: @JamesGuthrie we do need them, otherwise sqlachemy won't be installed, and you'll get an error when running tests.

alejandrodnm:
@smoya should we add `just pgai install` [here](https://github.com/timescale/pgai/blob/a4298c3939ac7817fa37298e47a1220484354c82/DEVELOPMENT.md#L367).

projects/pgai/justfile
JamesGuthrie: Aah, I see now. I had missed that we now _do_ have an optional dependency.

smoya:
> @smoya should we add `just pgai install` [here](https://github.com/timescale/pgai/blob/a4298c3939ac7817fa37298e47a1220484354c82/DEVELOPMENT.md#L367)? Or maybe have it as a dependency of `just pgai test` or `just pgai ci`?

- Added to `just pgai ci`. 
- Used `just pgai ci` in GH instead.

PR #326, opened 2024-12-20
test: use psycopg over psycopg2 for sqlalchemy tests

MasterOdin:
PR replaces usage of `psycopg2` with `psycopg` for the sqlalchemy tests that were added in #265, as we already were installing `psycopg` for other tests, and that installing `psycopg2` can be a bit annoying for folks on macOS (see https://github.com/psycopg/psycopg2/issues/1200 as an example).


smoya (approved):
Yeah!💯💯💯

PR #329, opened 2024-12-22
Default PG_BIN

rdu90:
If one runs this w/o just, then this variable won't necessarily be set in the environment and the code will fail This patch adds a default for the getenv call

### DIFF
--- a/projects/extension/build.py
+++ b/projects/extension/build.py
@@ -385,7 +385,7 @@ def clean_sql() -> None:
 
 
 def postgres_bin_dir() -> Path:
-    bin_dir = os.getenv("PG_BIN")
+    bin_dir = os.getenv("PG_BIN", "/usr/bin")
     if bin_dir is not None and Path(bin_dir).is_dir():
         return Path(bin_dir).resolve()
     else:


rdu90:
I picked /usr/bin because that is where pg_config was installed on my Debian system. there might be better variables to use in the code.
Also, I could change it so that it errors out but tells the user to set that variable...but. that seems...tedious for the person running it

JamesGuthrie:
@rdu90 we did have an issue related to `PG_BIN`, which was in the 0.6.0 release. In the meantime we merged this PR which fixed that problem: #286.

I suspect that the fix in #286 will also fix the problem that you encountered.

Can you confirm which version you used when you encountered the problem? If you were not testing on the latest main, could you try re-testing on the latest main and see if the issue still occurs.

If the issue you encountered was not fixed in #286, can you give exact steps to reproduce the issue that you see?

rdu90:
I am using that version

I was following the instructions here

https://github.com/timescale/pgai/blob/main/docs/install_from_source.md

Which says to follow at the latest tagged release and gives the .60 version 

I wonder if the other issue I had will also be solved by using the newest version (I had to remove a few packages installed by apt in order to get things to work on my virtual environment) 

I will get back to you in a few hours

rdu90:
I looked over the code in main and the patch there addresses the problem

I do see something that I'll make another PR for evaluation, nothing big, just a nitpick because I'm like that ;-)

PR #340, opened 2025-01-06
fix: handle 'null' value in chunking 'chunk_column'

### LINKED ISSUE #334: [Bug]: Null items in source table cause problems
### What happened?

Opening on behalf of user

https://discord.com/channels/1246241636019605616/1249554562004680715/1323321996007833621

### pgai extension affected

_No response_

### pgai library affected

_No response_

### PostgreSQL version used

X

X

Docker

Not applicable

_No response_

```bash
Make source column nullable and add null values
```

None

JamesGuthrie:
The chunking configuration takes a 'chunk_column' parameter, which determines which column of the source row is chunked.

This commit treats a null entry as being equivalent to the empty string, and no embeddings are generated.

Fixes #334

### DIFF
--- a/projects/pgai/pgai/vectorizer/chunking.py
+++ b/projects/pgai/pgai/vectorizer/chunking.py
@@ -80,7 +80,8 @@ class LangChainCharacterTextSplitter(BaseModel, Chunker):
         Returns:
             list[str]: A list of chunked strings.
         """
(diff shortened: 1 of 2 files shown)


jgpruitt (approved):

Askir (approved):
Looks good to me, tests seem to fail because of incorrect API keys? Bit confused, did someone upload new ones or did ours time out?

PR #351, opened 2025-01-09
feat: allow users to configure a base_url for the vectorizer OpenAI embedder

smoya:
This PR introduces the ability to configure the base_url in the OpenAI embedded for vectorizers. This enables greater flexibility for connecting to custom OpenAI API endpoints. This change adds support not only for private API deployments or testing environments, but future implementation of embedders that use the same API as OpenAI, such as [Azure OpenAI](https://learn.microsoft.com/en-us/azure/ai-services/openai/reference).

The changes include adding a new class `BaseURLMixin` which can be used in any other compatible embedder.


projects/extension/sql/idempotent/008-embedding.sql
JamesGuthrie, line 1: There are a couple of failing test - you need to recreate the outputs of the `contents/test_contents.py ` and `privileges/test_privileges.py ` tests.

The `upgrade/test_upgrade.py ` is also failing. This test validates that a fresh install at the current version has the same functions, etc. in it as an install which was upgraded from the previous version to the current version. It's failing because you've changed the signature of `ai.embedding_openai`, but not explicitly dropped the old function. As a result, the upgrade path has both versions of `ai.embedding_openai`. To fix this, you need to add an incremental migration to drop the old signature for this function.

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
```diff
@@ -235,23 +245,41 @@ def test_params(request: pytest.FixtureRequest) -> tuple[int, int, int, str, str
...
-                "chunking_character_text_splitter('content')",
-                "formatting_python_template('$chunk')",
+                (
+                    1,
+                    1,
+                    1,
+                    "chunking_character_text_splitter('content')",
+                    "formatting_python_template('$chunk')",
+                ),
+                None,  # No base_url is set. Use default (https://api.openai.com/v1)
```
JamesGuthrie, line 258: This parameter needs to flow into the casette name

projects/pgai/tests/vectorizer/cassettes/openai-character_text_splitter-chunk_value-items=1-batch_size=1.yaml
JamesGuthrie, line 1: How did you get this content into the cassette file? I don't see how I would easily be able to reproduce this test if I need to at a later date.

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
smoya: Good point. Added 👍

projects/pgai/tests/vectorizer/cassettes/openai-character_text_splitter-chunk_value-items=1-batch_size=1.yaml
smoya: I changed back to what I used when creating this file. I ran a `mitmproxy` in my local env  (localhost:8000) rewriting the request to OpenAI API endpoint. Made this by running this oneliner: 

```sh
mitmproxy --listen-port 8000 --mode reverse:https://api.openai.com
```

Do you believe it's worth to add a comment in the test with this? Do you think there is a better alternative though? I though on rather create the reverse proxy by code in python but since this is just "one time" for creating the cassette, I discarded it.

smoya: I tried to write a simple reverse proxy in the conftest.py file and, even though it works, VCR intercepts the call and the cassette ends up thinking the final URL is the openai one. Also entered in a infinite loop but that's most probably related to the fact I'm running the proxy in a separate thread (didn't spent more time on it).

Unless you know a simplest alternative, I will go with adding a comment with the mitmproxy command as the way to do it.

smoya: I managed to make it work integrating mitmproxy in the test code. Now It runs a proxy on that test, making the cassette be reproducible again.

projects/pgai/tests/vectorizer/conftest.py
jgpruitt, line 1: So the addition of the mitmproxy was so that we have something to test with? The proxy just forwards to OpenAI so we have a non-openai URL that acts just like OpenAI (because it is OpenAI)?

smoya: Yes. You have more context in [this comment](https://github.com/timescale/pgai/pull/351#discussion_r1910360474). 

Basically the proxy added in those tests is acting as a simple reverse proxy that rewrites requests made to a particular url+port (e.g. localhost:8000) to the OpenAI API. The expectations of such tests are basically the same as if requesting **directly** to the OpenAI API, in this way we can properly test that any provider that implements the same exact API as OpenAI does, will be compatible with our OpenAI vectorizer embedded config.

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
```diff
@@ -298,6 +324,8 @@ def test_process_vectorizer(
                 catch_exceptions=False,
             )
 
+        # assert result.stdout != ''
+        # assert result.stderr == ''
```
smoya, line 328: :) need to remove this

projects/pgai/pgai/vectorizer/embeddings.py
```diff
@@ -165,6 +165,26 @@ async def setup(self) -> None:  # noqa: B027 empty on purpose
...
+    def set_base_url(self, base_url: str | None):
+        """
+        Sets the base URL from the provided secrets.
+
+        Args:
+            base_url (str | None): The base URL to be set.
+        """
+        self.base_url = base_url
+
+
```
adolsalamanca, line 187: Others might know better, but I believe it would be a bit more idiomatic to use
```
@base_url.setter
    def base_url(self, base_url):
        self._base_url = base_url
```

adolsalamanca (approved):
Looks good, left a couple of comments

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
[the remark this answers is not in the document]
smoya: Yeah those are gone now

projects/pgai/pgai/vectorizer/embeddings.py
smoya: Interesting. I see we are not being very consistent on the usage of annotations regarding properties. However, it seems is the way to go, as I'm seeing in other well-known libraries. 
Will make the change 💯

JamesGuthrie (approved):

projects/pgai/tests/vectorizer/conftest.py
```diff
@@ -113,3 +119,50 @@ def postgres_container(
     load_openai_key: bool = params.get("load_openai_key", True)  # type: ignore
 
     return postgres_container_manager(load_openai_key=load_openai_key)  # type: ignore
+
+
+class ReverseProxy:
```
AldoFusterTurpin, line 124: Maybe this sound stupid, but I think it would be better to name this ReverseProxyConfig as this is not really the proxy itself, this just carries the config, otherwise I would expect this class to provide a method "run" to execute the proxy. 🙂

projects/pgai/tests/vectorizer/conftest.py
```diff
@@ -113,3 +119,50 @@ def postgres_container(
...
+    # mitmproxy relies on asyncio for its event loop
+    def run_event_loop():
+        asyncio.run(start_proxy())
+
+    proxy_thread = threading.Thread(target=run_event_loop, daemon=True)
+    proxy_thread.start()
+
+
+@pytest.fixture(scope="function")
+def openai_proxy_url(request: pytest.FixtureRequest):
```
AldoFusterTurpin, line 157: Is there a good reason tu use yield here ? Maybe I am missing something, but as we always return just a single value, I think using always return instead of yield would simpler.

smoya: Not anymore. You are right.

projects/pgai/tests/vectorizer/conftest.py
smoya: Renamed to `ReverseProxyAddon` which is what it is actually. Does it make sense?

projects/pgai/tests/vectorizer/conftest.py
smoya: fixed

projects/pgai/pgai/vectorizer/embeddings.py
```diff
@@ -165,6 +165,17 @@ async def setup(self) -> None:  # noqa: B027 empty on purpose
         """
 
 
+class BaseURLMixin:
```
AldoFusterTurpin, line 168: I prefer to always provide constructor to classes to avoid inconsistent state between instantiation and usage of classes, but this is just MHO and more about OOP in general.

AldoFusterTurpin (commented):
LGTM! Thanks!
(Not formal "approve" as I am missing context)

projects/pgai/pgai/vectorizer/embeddings.py
smoya: Mixed feelings here. Simplicity applies here IMHO: not having to deal with calls to sup classes from child classes gives me confidence on this. Otherwise, I would even opt for not creating sub a mixin class. However, I wanted to start having a class to be reused by any other embedded that can set their base_url, so if needed, we can know if a class implements that.

Having said that, I'm a completely open to any recommendation from any python expert here as I do not have enough experience in this language unfortunately. I always advocate for creating idiomatic code rather than focus on general practices because Developer Experience is what matters the most here.

PR #354, opened 2025-01-10
fix: two usability issues with sqlalchemy

Askir:
I am fixing two quite annoying issues with the SQLAlchemy integration in this PR:
1. It wasn't possible to have the same field name configured on two different entities. So you can't have `content_embeddings` twice, because the name was colliding in the Mapper configuration. So I changed the dynamically generated Model name to contain the parent Model.

2. The way we were configuring the relationship into the other direction via `backref` was somewhat lazily evaluated and is also no longer recommended by SQLAlchemy. It's considered "legacy". I changed it with some back and forth to be simply two relationships in both directions and added a `parent_kwargs` field that allows to configure the relationship into the reverse direction.

### DIFF
--- a/projects/pgai/tests/vectorizer/extensions/test_sqlalchemy_relationship.py
+++ b/projects/pgai/tests/vectorizer/extensions/test_sqlalchemy_relationship.py
@@ -1,8 +1,8 @@
 from typing import Any
 
 import numpy as np
-from sqlalchemy import Column, Engine, Integer, Text
-from sqlalchemy.orm import DeclarativeBase, Session
+from sqlalchemy import Column, Engine, Integer, Text, select
+from sqlalchemy.orm import DeclarativeBase, Session, backref, joinedload
 from sqlalchemy.sql import text
 from testcontainers.postgres import PostgresContainer  # type: ignore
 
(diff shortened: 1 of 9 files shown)


projects/pgai/tests/vectorizer/extensions/test_sqlalchemy_relationship.py
```diff
@@ -88,3 +88,77 @@ def test_vectorizer_embedding_creation(
...
+    some development with the extension.
+    The nature of the vectorizer_relationship being a descriptor messes
+    with sqlalchemys relationship resolution.
+    It was previously using `backref` to propagate the parent field,
+    which is resolved later and an immediate access
+    to build queries like this would fail.
+    """
+    select(Document.content_embeddings).options(
+        joinedload(Document.content_embeddings.parent)  # type: ignore
+    )  # type: ignore
```
adolsalamanca, line 164: I guess it's intended not to have assertions here, to just verify it builds?

adolsalamanca (approved):
Looks good, just left a small comment.

projects/pgai/tests/vectorizer/extensions/test_sqlalchemy_relationship.py
Askir: yes, it was throwing an exception before 😅

cevian (commented):
Is there a test for issue 1 added somewhere? (I might have missed it)

Askir:
@cevian yes the select test covers this too, because the same file has two models with `content_embeddings`.

PR #362, opened 2025-01-14
feat: add warning when trying to install pre-release version

JamesGuthrie:
Pre-release versions of the pgai extension do not have a defined upgrade
path to released versions. Installing these versions could leave
databases in a problematic state that cannot be cleanly upgraded.

This change prevents accidental installation of development builds by
directing users to official releases, while still allowing intentional
installation of pre-release versions via the explicit build-install
command.

### DIFF
--- a/projects/extension/build.py
+++ b/projects/extension/build.py
@@ -7,6 +7,7 @@ import shutil
 import subprocess
 import sys
 import tempfile
+import textwrap
 from pathlib import Path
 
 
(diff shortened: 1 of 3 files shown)


projects/extension/build.py
```diff
@@ -583,7 +584,33 @@ def build() -> None:
...
+
+                If you are certain that you want to install a pre-release version, run:
+                    `{command}`
+            """)
+        )
+        exit(1)
+
+
 def install() -> None:
+    error_if_pre_release()
```
jgpruitt, line 613: Should this be called here or in install_sql()?

jgpruitt: i.e. do we only want to catch this if they do `just install` or also if they do `just install-sql` as well?

jgpruitt (approved):
One question. Otherwise LGTM

projects/extension/build.py
JamesGuthrie: I put it here specifically (at the top of this function), and not in `install_sql` so that it aborts early (and not after doing a bunch of pip installs etc.). I can put it in `install_sql` too, although I'm not too concerned about that path because we don't really document it anywhere.

jgpruitt: makes sense.

PR #366, opened 2025-01-15
Fixed embedding table name typos in vectorizer.md

ihis-11:
I updated the embedding table name in the **The embedding storage table** section to ensure that it follows [pgai Vectorizer's API rules on destination table naming](https://github.com/timescale/pgai/blob/main/docs/vectorizer-api-reference.md#parameters).


adolsalamanca (approved):
looks good, but commit message should be adjusted for CI to pass.

PR #367, opened 2025-01-15
chore: allow the creation of blank issues in this repo

smoya:
I believe we should not be that restrictive with users that want to create an issue and allow them to create blank issues if they don't find a template that fits with their needs.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
   - name: pgai discord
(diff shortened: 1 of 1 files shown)


adolsalamanca (approved):

projects/extension/requirements-locked.txt
```diff
@@ -0,0 +1,202 @@
+# This file was autogenerated by uv via the following command:
```
alejandrodnm, line 1: @smoya are you hacking us? 🙈

smoya: Omg 🤦 veeeeery slow brain today

smoya: solved

PR #371, opened 2025-01-16
chore: lock transitive dependencies using uv

JamesGuthrie:
This change switches to using uv to manage our dependencies. This is
primarily motivated by the desire to lock the versions of our transitive
dependencies. Also uv is fast.

uv tracks locked dependencies in `uv.lock`.

In order to not impose uv on end-users who want to install the extension,
we provide a locked requirements file `requirements-lock.txt` which is used
in the installation process. This file must be updated when `uv.lock` is
updated, with `uv export -o requirements-lock.txt`.


Askir (approved):
hehe I can approve my own changes. JK though, someone else please take a look :D

PR #380, opened 2025-01-21
docs(dev): improve and split DEVELOPMENT.md

smoya:
This PR improves the Development documentation by:

- It splits current `DEVELOPMENT.md` file into 3 files:
  1. Moves all guidelines related to the extension to `projects/extension/DEVELOPMENT.md`.
  2. Moves all guidelines related to the pgai library to `projects/pgai/DEVELOPMENT.md`. 
  3. Keeps the current root file as index, adding links to each of the files listed above.
- It improves (I hope it does) the `DEVELOPMENT.md` root file by making it a bit more comprehensible and by adding missing context.

Additionally, I fixed the `just install-commit-hook` command as it was failing. I didn't move that change into its own single PR because the change is sooo small, but if you believe it should be done in that way, just mention it, I'm fine.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,418 +1,96 @@
-# Develop and test changes to pgai
+# Contributing to pgai
 
-pgai brings embedding and generation AI models closer to the database. Want to contribute to the pgai project?
-Start here.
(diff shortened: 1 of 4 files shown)


jgpruitt (approved):
LGTM!

smoya:
I'm gonna quickly merge this to avoid losing documentation if this PR sticks for a while here.

PR #381, opened 2025-01-21
feat(text_to_sql): anthropic working with tool use

jgpruitt:
Reworked text-to-sql to present tools to the LLM. This technique allows the LLM to ask for more database context if information is missing that is needed to provide a confident answer. It also allows the LLM to indicate which of the context elements were relevant to the user's question.

This is not production-ready. It is feature-flagged. The tests are out-of-date and thus disabled. I need to merge this WIP so that Matt and I can each branch from this point and work on separate but related parts.


alejandrodnm:
It seems that the changes are not just about anthropic support for tools. I see many changes to the semantic catalog functions. 

For example `projects/extension/sql/idempotent/905-semantic-catalog-render.sql` are all those new functions just for adding tools support?

jgpruitt:
> It seems that the changes are not just about anthropic support for tools. I see many changes to the semantic catalog functions.
> 
> For example `projects/extension/sql/idempotent/905-semantic-catalog-render.sql` are all those new functions just for adding tools support?

This is a WIP for text-to-sql overall. All of the text-to-sql stuff is in development and changing constantly. Tool support was the "most profound" change in this PR.

JamesGuthrie (approved):

PR #382, opened 2025-01-22
chore: install extension dependencies from requirements-lock.txt

JamesGuthrie:


projects/extension/build.py
```diff
@@ -530,38 +530,34 @@ def install_py() -> None:
...
+    ):  # delete package info if exists
+        shutil.rmtree(d)
+    cmd = f'{bin} install -v --no-deps --compile --target "{version_target_dir}" "{ext_dir()}"'
+    subprocess.run(
+        cmd,
+        check=True,
+        shell=True,
+        env=os.environ,
+        cwd=str(ext_dir()),
+    )
```
Askir, line 560: Re-iterating my question from the thread:
Do we actually need the code to be a package? Can't we just do `pip install -r requirements-lock.py` and copy the python source files into e.g. a python directory and be done with it?

JamesGuthrie: TBH I don't know - what do we gain from that approach vs. this?

smoya: > Can't we just do pip install -r requirements-lock.py and copy the python source files into e.g. a python directory and be done with it?

If this is what we are gonna ask users to do in case they want to use pgai library in their own code base, I don't think it is viable.

JamesGuthrie: No, but it also doesn't make sense to have pinned/locked deps in a lib.

JamesGuthrie: That is to say: this approach doesn't impose our locked deps on lib users. We do still have the issue of pretty heavily-pinned deps, which we should perhaps consider relaxing a little for users of the lib.

Askir: >TBH I don't know - what do we gain from that approach vs. this?

We'd remove a step from the installation, which removes the dependency of setuptools or any other build tool.

And no this has nothing to do with the pgai library. I'm just questioning if the extensions python code should rly be handled like a "library" which is what we are doing right now. Or if it actually is an application.

PR #384, opened 2025-01-22
docs(dev): add missing 'just ext freeze' step into the extension release docs

smoya:


smoya:
Not required. freeze() step added into the `just ext build-release` command in https://github.com/timescale/pgai/pull/383/commits/48bd318f670f61db234ec84717b274c45f737ae5

Issue #352, opened 2025-01-09
[Feature]: Support host changes for openai calls


djt:
### What problem does the new feature solve?

This will allow you to use other openai compatible providers to be slotted in with minimal work on pgai.  More providers is a better offering.

### What does the feature do?

Allows third party openai compatible API providers to be used with pgai.

Best example currently is https://console.groq.com/docs/openai.

### Are you going to work on this feature?

🆘 No, could someone else please consider working on it?

JamesGuthrie:
@djt currently the `ai.openai_list_models`, `ai.openai_embed`, `ai.openai_chat_complete`, and `ai.openai_moderate` functions do provide this option through the `base_url` parameter. This is not supported in the vectorizer configuration, but we're actively working on adding this now (see #351)

smoya:
Hi @djt , as a matter of an update, my PR that implements setting up a `base_url` to the OpenAI embedder is in final review phase.

Once merged and, most important, released, I will ping you so you can test it (if you are willing to, of course).

djt:
That would be great, thanks.

smoya:
Hi @djt sorry for the delay. We've just released that feature and is available in the latest docker image of the vectorizer https://hub.docker.com/layers/timescale/pgai-vectorizer-worker/latest/images/sha256-f678a4948d16be3cd7395199c27d785193203fbef1eaf51a5f06c57aa80e1e3f

You can check the addition of the `base_url` param to the docs [here](https://github.com/timescale/pgai/blob/main/docs/vectorizer-api-reference.md#parameters-3).

Looking forward to your feedback!

BW-Projects:
I'm getting an error saying, that the tokenizer for the custom embedding model could not be found:

`Could not automatically map BAAI/bge-m3 to a tokeniser. Please use `tiktoken.get_encoding` to explicitly get the tokeniser you expect.`

Why is there even a need for the tokenizer - to split strings that have more token than the embedding model context size?

Here the full log:
```
2025-01-23 14:25:34 [debug    ] obtained secret 'OPENAI_API_KEY' from environment
2025-01-23 14:25:34 [info     ] running vectorizer             vectorizer_id=23
2025-01-23 14:25:34 [debug    ] Items pulled from queue: 50
2025-01-23 14:25:34 [error    ] unexpected error: 'Could not automatically map BAAI/bge-m3 to a tokeniser. Please use `tiktoken.get_encoding` to explicitly get the tokeniser you expect.'
[...]

PR #390, opened 2025-01-24
feat(vectorizer): disable OpenAI tokenization when a model does not have a tokenizer match

smoya:
## Context

At this moment, we are tokenizing the chunks before being sent to OpenAI embedding API. The way that's done is through matching a tokenizer with the given OpenAI model thanks to tiktoken. See [here](https://github.com/timescale/pgai/blob/main/projects/pgai/pgai/vectorizer/embedders/openai.py#L203-L205).

That works because tiktoken has a map of `openai_model => tokenizer`.  
Notice that **is only compatible with OpenAI models**.

## Description
This PR allows to skip tokenization whenever we are not able to find a tokenizer for a given model. 

Why this is needed? Because two reasons:

1. Since tiktoken is only compatible with OpenAI models, the call made to match a non-openai model with a tokenizer will always fail.
2. Not all LLM providers accept tokenized input but rather just plain text.

That, among the recently `base_url` param we added in our latest release, this will allow to call any other LLM with an API compatible with OpenAI. 

Relates to https://github.com/timescale/pgai/issues/352

### DIFF
--- a/projects/pgai/pgai/vectorizer/embedders/openai.py
+++ b/projects/pgai/pgai/vectorizer/embedders/openai.py
@@ -1,7 +1,7 @@
 import re
-from collections.abc import Sequence
+from collections.abc import Iterable, Sequence
 from functools import cached_property
-from typing import Any, Literal
+from typing import Any, Literal, cast
 
 import openai
 import tiktoken
@@ -14,11 +14,11 @@ from ..embeddings import (
     BaseURLMixin,
     BatchApiCaller,
     ChunkEmbeddingError,
+    Document,
     Embedder,
     EmbeddingResponse,
     EmbeddingVector,
     StringDocument,
-    TokenDocument,
     Usage,
     logger,
 )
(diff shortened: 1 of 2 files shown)


projects/pgai/pgai/vectorizer/embedders/openai.py
```diff
@@ -69,9 +70,9 @@ def _embedder(self) -> resources.AsyncEmbeddings:
     def _max_chunks_per_batch(self) -> int:
         return 2048
 
-    async def call_embed_api(self, documents: list[TokenDocument]) -> EmbeddingResponse:
+    async def call_embed_api(self, documents: list[Document]) -> EmbeddingResponse:
         response = await self._embedder.create(
-            input=documents,
+            input=cast(list[str] | Iterable[Iterable[int]], documents),
```
smoya, line 75: :/ I'm meeeelting. Thinking on adding an ignore comment here instead 😓

alejandrodnm (approved):

projects/pgai/pgai/vectorizer/embedders/openai.py
JamesGuthrie: The typing is subtly incorrect for this whole thing, and is driving me mad (I was the one who added the incorrect stuff). I think that we can merge this as-is though.

JamesGuthrie (approved):

PR #393, opened 2025-01-24
build: ignore sql files with 'x' prefix

jgpruitt:
While developing, sometimes I want to implement something a few different ways and compare. The easiest thing to do is to keep each version in a separate SQL file. But I can only have one at a time built and tested.

I don't want to

1. move the SQL file to another directory - this is a pain and I still want to be able to look at the file in my editor in the context of the project
2. comment out the entire body of the SQL - also a pain and destroys and IDE niceties
3. delete the file - I haven't decided to discard the work

What I want is an easy way to temporarily ignore a SQL file while I'm developing.

This PR allows you to rename a SQL file to add an 'x' prefix to tell the build to ignore the file.

### DIFF
--- a/projects/extension/build.py
+++ b/projects/extension/build.py
@@ -121,7 +121,9 @@ def idempotent_sql_dir() -> Path:
 
 
 def idempotent_sql_files() -> list[Path]:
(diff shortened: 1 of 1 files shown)


smoya (approved):

JamesGuthrie (commented):
I'm confused - did anything change other than formatting?

PR #394, opened 2025-01-24
chore: get rid of nested parametrization in vectorizer tests

Askir:
Refactors the vectorizer test suite to be more easily readable by:
- Removing nested test_params fixture (num_items, concurrency, batch_size, chunking, formatting) in favor of direct parameters that are then passed to setup functions
- Creating dedicated configuration functions (configure_openai_vectorizer, configure_ollama_vectorizer, etc.) to replace the fixtures
- Using a common `run_vectorizer_worker` helper function standardizing CLI invocation

I also want to split the file up, but wondering how and if I should do that separately.

### DIFF
--- a/projects/pgai/tests/vectorizer/test_vectorizer_cli.py
+++ b/projects/pgai/tests/vectorizer/test_vectorizer_cli.py
@@ -11,14 +11,13 @@ from typing import Any
 import openai
 import psycopg
 import pytest
-from click.testing import CliRunner
 from psycopg import Connection, sql
 from psycopg.rows import dict_row
 from testcontainers.ollama import OllamaContainer  # type: ignore
 from testcontainers.postgres import PostgresContainer  # type: ignore
 
-from pgai.cli import vectorizer_worker
 from tests.vectorizer import expected
+from tests.vectorizer.utils import run_vectorizer_worker
 
 count = 10000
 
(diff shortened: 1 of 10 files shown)


projects/pgai/uv.lock
```diff
@@ -2053,7 +2053,6 @@ wheels = [
 
 [[package]]
 name = "pgai"
-version = "0.5.0"
```
Askir, line 2056: when i ran `uv sync --all-extras` this was removed, I didn't do that 😅

JamesGuthrie: Ya I don't quite get what's going on there. I suspect that different minor versions (0.4 vs 0.5) of uv treat this parameter differently, and some write it in and others remove it.

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
```diff
@@ -88,39 +87,37 @@ def cli_db_url(cli_db: tuple[TestDatabase, Connection]) -> str:
...
+        values = [(i, i, f"post_{i}") for i in range(1, number_of_rows + 1)]
         cur.executemany(
             "INSERT INTO blog (id, id2, content) VALUES (%s, %s, %s)", values
         )
     return table_name
 
 
 @pytest.fixture(scope="session")
-def ollama_connection_url():
+def ollama_container_url_or_url():
```
JamesGuthrie, line 120: This confuses me, what about just `ollama_url`?

JamesGuthrie (approved):
Wow - this is so much better. Thank you!

projects/pgai/tests/vectorizer/test_vectorizer_cli.py
Askir: yeh fair, I was just confused why this is a fixture and not a constant 😅 which is why i put the container in there, but tbh fair to encapsulate and hide the details.

PR #395, opened 2025-01-24
chore(extension): enable more linting rules

JamesGuthrie:
This enables some more linting rules in the extension, aligning linting rules across the pgai library and the extension.


smoya (approved):
LGTM! 🚀🌔

PR #396, opened 2025-01-24
chore: pass all provider env vars into docker container

JamesGuthrie:
This change ensures that the environment variables for all of our model providers are passed into the extension container. This makes local development a bit easier, but I'm not 100% convinced that it's warranted.


projects/extension/build.py
jgpruitt, line 1: I like it. I wish there was a more dynamic way though. Like maybe if there is a .env file add all the env vars from that file to the docker container?

jgpruitt (approved):

PR #397, opened 2025-01-24
Improve LiteLLM docs

JamesGuthrie:
- document ai.litellm_embed()
- clarify huggingface vectorizer

### DIFF
--- a/docs/vectorizer-api-reference.md
+++ b/docs/vectorizer-api-reference.md
@@ -350,7 +350,7 @@ Note: Mistral limits the maximum input per batch to 16384 tokens.
 
 ##### Azure OpenAI
 
-To set up a vectorizer with Azure OpenAI you require three values from the Azure AI Foundry console:
+To set up a vectorizer with Azure OpenAI you require these values from the Azure AI Foundry console:
 - deployment name
 - base URL
 - version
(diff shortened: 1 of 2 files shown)


jackyliang:
I see openai used as an example for the litellm Vectorizers. Can we include other examples such as HuggingFace? 

https://huggingface.co/spaces/mteb/leaderboard

^ Here are the top embedding models on HF

alejandrodnm (approved):

docs/vectorizer-api-reference.md
```diff
@@ -378,6 +378,34 @@ Configure the vectorizer, note that the base URL and version are configured thro
...
+with serverless inference to force the call to block until the model has been
+loaded. If you do not use `wait_for_model`, it's likely that vectorization will
+never succeed.
+
+```sql
+    SELECT ai.create_vectorizer(
+        'my_table'::regclass,
+        embedding => ai.embedding_litellm(
+          'huggingface/BAAI/bge-small-en-v1.5',
+        , 384
```
jackyliang, line 400: Does this not need `api_key_name => 'HUGGINGFACE_API_KEY',`?

JamesGuthrie: It doesn't _need_ it, and TBH it's confusing when it does and doesn't. If the `HUGGINGFACE_API_KEY` environment variable is set in the vectorizer worker, then it doesn't need to be set. If not, and the vectorizer should get the API key from the database, then the `api_key_name` parameter must be set.

PR #402, opened 2025-01-27
feat: pause/resume vectorizer

alejandrodnm:
Add ability to enable/disable vectorizers using ai.enable_vectorizer_schedule
and ai.disable_vectorizer_schedule functions. This applies to vectorizers
configured with ai.scheduling_none that use polling-based workers.

Key changes:
- Add disabled column to ai.vectorizer table
- Expose disabled status through ai.vectorizer_status view
- Add checks in Vectorizer Worker to stop processing when disabled
- Improve stopping behavior for ai.scheduling_timescaledb vectorizers
- Add Features class to manage version-dependent functionality

The Vectorizer Worker now checks disabled status before processing each
batch. Disabled vectorizers (both ai.scheduling_none and
ai.scheduling_timescaledb) exit after completing their current batch,
improving on previous behavior where timescaledb vectorizers processed
their full queue before stopping.

### DIFF
--- a/projects/pgai/pgai/vectorizer/features/features.py
+++ b/projects/pgai/pgai/vectorizer/features/features.py
@@ -1,0 +1,19 @@
+from functools import cached_property
+
+from packaging import version
+
+
+class Features:
+    """Feature flags and version-dependent functionality manager."""
+
+    def __init__(self, ext_version: str) -> None:
+        self.ext_version = version.parse(ext_version)
+
+    @cached_property
+    def disable_vectorizers(self) -> bool:
+        """If the disable vectorizer feature is supported by the extension.
+
+        The feature consists of a `disabled` column in the `ai.vectorizer`
+        table, and the `ai.vectorizer_status` view.
+        """
+        return self.ext_version > version.parse("0.7.0")
--- a/projects/pgai/pgai/vectorizer/vectorizer.py
+++ b/projects/pgai/pgai/vectorizer/vectorizer.py
@@ -24,6 +24,7 @@ from .chunking import (
 )
 from .embedders import LiteLLM, Ollama, OpenAI, VoyageAI
 from .embeddings import ChunkEmbeddingError
+from .features import Features
 from .formatting import ChunkValue, PythonTemplate
 from .processing import ProcessingDefault
 
@@ -115,6 +116,8 @@ class Vectorizer:
     source_pk: list[PkAtt]
     errors_schema: str = "ai"
     errors_table: str = "vectorizer_errors"
+    schema: str = "ai"
+    table: str = "vectorizer"
 
 
 class VectorizerQueryBuilder:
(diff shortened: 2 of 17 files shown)


projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -461,18 +482,51 @@ async def run(self) -> int:
...
+            async with conn.cursor() as cursor:
+                await cursor.execute(
+                    self.queries.is_vectorizer_disabled_query,
+                    [self.vectorizer.id],
+                )
+                row = await cursor.fetchone()
+                if not row:
+                    raise Exception("vectorizer row not found")
+                if row[0]:
+                    return False
```
jgpruitt, line 526: Shouldn't this return the value of `disabled` from the row in the database rather than a hardcoded `False`?

jgpruitt (changes requested):
1 question

projects/pgai/pgai/vectorizer/vectorizer.py
alejandrodnm: No because if it's not disable we still need to check the `self._continue_processing_hook` to see if we should continue processing.

The hook on cloud is a function that calculates if we are near the 80% execution time that we are allowed to run.

jgpruitt: Ahh! gotcha.

jgpruitt (approved):

adolsalamanca (approved):
neat, thanks!

projects/pgai/pgai/vectorizer/features/features.py
```diff
@@ -0,0 +1,19 @@
...
+        self.ext_version = version.parse(ext_version)
+
+    @cached_property
+    def disable_vectorizers(self) -> bool:
+        """If the disable vectorizer feature is supported by the extension.
+
+        The feature consists of a `disabled` column in the `ai.vectorizer`
+        table, and the `ai.vectorizer_status` view.
+        """
+        return self.ext_version > version.parse("0.7.0")
```
smoya, line 19: 💯

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -461,18 +482,51 @@ async def run(self) -> int:
...
-                if not self._continue_processing(loops, res):
+                if not await self._continue_processing(conn, loops, res):
                     return res
                 items_processed = await self._do_batch(conn)
                 if items_processed == 0:
                     return res
                 res += items_processed
                 loops += 1
 
+    async def _continue_processing(
```
smoya, line 499: Nit: by reading the func name, seems you are telling and not asking. In other words, it seems the intention is to actually resume processing instead of asking if the process should continue.

smoya (approved):
Small nits. LGTM otherwise 💯💯

PR #405, opened 2025-01-28
Fixes to ai.openai_chat_complete and openai support for text-to-sql

jgpruitt:
Made improvements to `ai.openai_chat_complete`
* `null` arguments are fully omitted from call to openai sdk
* `tool_choice` type changed from `jsonb` to `text` so that [all possible values can be expressed](https://platform.openai.com/docs/api-reference/chat/object)

The changes above were required for implementing support for openai for text-to-sql, which is the second commit in this PR.

Not in this PR but we should consider:

1. updating the openai python lib to a recent version
2. adding the multiple new parameters to the chat completion api


projects/extension/sql/idempotent/001-openai.sql
```diff
@@ -154,7 +154,7 @@ create or replace function ai.openai_chat_complete
 , temperature float8 default null
 , top_p float8 default null
 , tools jsonb default null
-, tool_choice jsonb default null
+, tool_choice text default null
```
Askir, line 157: I guess there is no unions in sql functions? :D

Askir (approved):

projects/extension/sql/idempotent/001-openai.sql
jgpruitt: Unfortunately no

alejandrodnm:
I wanted to discuss if it makes sense to put all these extra args into an `extra_client_args` and `extra_request_args` parameters, so that we don't have to keep adding more to the function signature.

PR #407, opened 2025-01-29
feat: tweak chunking func to support doc chunking in the future

smoya:
# Summary

This PR modifies all chunking functions to let empty `chunk_column` arguments. 
An extra validation has been added to the `ai._validate_chunking` function that raises an exception if an empty `chunk_column` is found and there is no `loader` configured in the `ai.create_vectorizer` function call.

> [!NOTE]  
> This change **does not** represent any breaking change in the API so it is safe to be released at any point.

We are introducing the concept of `loader`, which in the future we will use for allowing loading data from external sources, such as documents. Parsing is, at this very early stage, implicit in the loader responsibility (for simplicity sake).

The trade-off of this actual PR is that the user could create a vectorizer passing an arbitrary or empty json as `loader` value. We didn't add a validation in purpose in this PR for simplifying the review. However, it will come in a next PR.
Feel free to say your opinion about this decision; we can add a simple validation if you believe it is mandatory for this PR.

This is a first step for moving into an API such as:

```sql
SELECT ai.create_vectorizer
( 'public.wiki'::regclass
, loader => ai.loader_from_document(file_uri_column => 'url')
, chunking=>ai.chunking_recursive_character_text_splitter() # the change here is allowing empty chunk_column field, nothing else.
, embedding=>ai.embedding_openai('text-embedding-3-small', 1536)
)
```


projects/extension/sql/idempotent/005-chunking.sql
```diff
@@ -2,11 +2,12 @@
 -------------------------------------------------------------------------------
 -- chunking_character_text_splitter
 create or replace function ai.chunking_character_text_splitter
-( chunk_column pg_catalog.name
+( chunk_column pg_catalog.name default ''
 , chunk_size pg_catalog.int4 default 800
 , chunk_overlap pg_catalog.int4 default 400
 , separator pg_catalog.text default E'\n\n'
 , is_separator_regex pg_catalog.bool default false
+, chunk_document pg_catalog.bool default false
```
Askir, line 10: Does it make sense instead of explicitly requiring this flag, just making the chunk column optional?
Then create_vectorizer method could check if a parser was specified and validate that a chunk column is only given if no parser is specified and the other way around (require a chunk column if no parser is there).

Would simplify the API a bit and again avoid misconfiguration (chunk document is false but a parser is specified).

Askir: Could also go as far and just ignore the chunk_column if a parser is specified.

smoya: Having a parser shoulnd't be a decision maker for assuming you are loading a file, and vice-versa; if the file doesn't really need from parsing (i.e. a plain text file) why would you configure one?

Regarding the signature of the chunkers, I agree we could just make the `chunk_column` optional (as this PR does) and not require from an extra param to indicate you want to chunk whatever the input is (whatever the loader + parser gives). In fact, that was our first idea iirc, however I believe @jgpruitt suggested this alternative because it didn't find it ideal. 

Summoning @jgpruitt again just in case we misunderstood your suggestion 🙏

smoya: Additionally, adding the `chunk_document`, even though defaulting to `false`, is breaking our upgrade tests because previous versions don't have `chunk_document` argument and we expect them to be present and set to `false`. 
We are skipping `null` values but not `false` (see [here](https://github.com/timescale/pgai/blob/c7aad8709320baf3aaeec9b25ec70935bdce6ac3/projects/extension/sql/idempotent/005-chunking.sql#L22)), and I don't think we need to change that behaviour either.

jgpruitt: >Having a parser shoulnd't be a decision maker for assuming you are loading a file, and vice-versa; if the file doesn't really need from parsing (i.e. a plain text file) why would you configure one?

Agreed. Also, we should consider the case where you might want a parser for content already in a database column. i.e. you may want a parser without loading a file.

jgpruitt: GOAL: if you are creating a vectorizer that does not load from S3 and you omit the chunk_column, this should fail at the time of create vectorizer, NOT at vectorizer run time.

So, the idea of having a `chunk_document` boolean was really just to provide a signal as to whether chunk_column is required or not.

If there is a better idea, I'm open to it.

jgpruitt: One idea: what if for the S3 vectorizer, chunk_column was the column containing the S3 URL?

jgpruitt: Is there a way we could make chunk_column just go away? Or configured at a higher level?

Will there be a "loader config" that we could use to signify that the chunk_column isn't needed? I mean we could have a loader config that specifies the env vars needed to do the S3 downloads.

jgpruitt: >Adding the chunk_document, even though defaulting to false, is breaking our upgrade tests because previous versions don't have chunk_document argument and we expect them to be present and set to false.

Okay, so if we move forward with the chunk_document field, you'll need to add an incremental script that looks at all the vectorizer configs with an old version and updates them to have the new field. This would get applied at upgrade time. (This is exactly why we have the upgrade tests -- to catch these things 😁)

smoya: > Is there a way we could make chunk_column just go away? Or configured at a higher level?

That's actually a great point. We chatted this offline, and both seemed to agreed this is responsibility of the loader.
The following is a big change on the API (breaking change actually), but is what I suggested as the ideal API for the next step.

```sql
SELECT ai.create_vectorizer(
    'table_name'::regclass,
    ## LOADER ##
    loader => ai.loader_from_document_column('uri_column'),
    # OR
    loader => ai.loader_from_column('table_column'), # this represents the current loading method
    ############
    parser => ai.parser_pdf() # defaults to ai.parser_pymupdf(),
    chunking => ai.chunking_recursive_character_text_splitter(), # note how we don't even need to specify the source of the data (column, or document).
    formatting => ai.formatting_python_template('$filename: $chunk'),
    embedding => ai.embedding_openai('text-embedding-3-small', 1536)
) as vectorizer_id
```

That allows: 

- To use loaders and parsers individually. 
- To remove the responsibility of loading the source of data in chunkers, limiting the responsibility to just transforming ANY input to an output.
- To allow parsing from raw data stored in a column (as you @jgpruitt also suggested in one of the previous comments)
- Ultimately, this aligns more with a real pipeline definition.

This obviously introduces breaking changes but tbh, it is a change I feel is going to, in the end, save us a lot of time and headaches when adding patches over patches on every new feature we add.
We are in a very early stage on our library and extension and I don't see a better time than now to do things like this.

Thoughts? :) 

cc some committers @adolsalamanca @Askir @jgpruitt

adolsalamanca (approved):

projects/extension/tests/vectorizer/test_vectorizer.py
adolsalamanca, line 1: Those tests were removed given those are validated in the privileges explicit tests.

projects/extension/tests/vectorizer/test_vectorizer.py
```diff
@@ -260,37 +261,6 @@ def test_vectorizer_timescaledb():
...
-                f"select has_table_privilege('bob', '{vec.view_schema}.{vec.view_name}', 'select')"
-            )
-            actual = cur.fetchone()[0]
-            assert actual
-
-            # bob should have select on the vectorizer table
-            cur.execute("select has_table_privilege('bob', 'ai.vectorizer', 'select')")
-            actual = cur.fetchone()[0]
-            assert actual
-
```
adolsalamanca, line 293: Those tests were removed as there are explicit validations already in the privileges tests.

projects/extension/sql/idempotent/013-vectorizer-api.sql
```diff
@@ -31,6 +31,7 @@ create or replace function ai.create_vectorizer
 , queue_table pg_catalog.name default null
 , grant_to pg_catalog.name[] default ai.grant_to()
 , enqueue_existing pg_catalog.bool default true
+, loader pg_catalog.jsonb default null
```
Askir, line 34: We're not using this right now right?

adolsalamanca: we're using it to perform the validation of the chunking specified.

adolsalamanca: https://github.com/timescale/pgai/blob/adol/tweak-chunking-for-s3/projects/extension/sql/idempotent/013-vectorizer-api.sql#L74-L76

projects/extension/sql/idempotent/005-chunking.sql
```diff
@@ -26,7 +26,7 @@ set search_path to pg_catalog, pg_temp
 -------------------------------------------------------------------------------
 -- chunking_recursive_character_text_splitter
 create or replace function ai.chunking_recursive_character_text_splitter
-( chunk_column pg_catalog.name
+( chunk_column pg_catalog.name default ''
```
jgpruitt, line 29: Isn't null a more appropriate default?

jgpruitt (changes requested):

projects/extension/sql/idempotent/005-chunking.sql
smoya: Fixed!

PR #412, opened 2025-01-29
Fix typo in python-integration.md

solugebefola:


JamesGuthrie (approved):

Askir:
Oh thanks for the fix, we use convention commits though so you'll need something like `fix: typo in python-integration.md` to get this merged 😇

Askir:
actually I can change it while merging, so I'll just do that.

PR #415, opened 2025-02-03
Add generate_description and generate_column_descriptions functions

MasterOdin:
PR adds two new functions that generate descriptions for database objects:

* `generate_description`
* `generate_column_descriptions`

Usage is similar to the `set_*` variants in that you'd pass the database object name (e.g. `public.foo`), with the addition of being able to pass in parameters for setting `catalog_name` and `config` which has similar setup as the `ai.text_to_sql` function.


projects/extension/sql/idempotent/901-semantic-catalog-set.sql
```diff
@@ -115,6 +136,27 @@ end;
...
+, save boolean default false
+, overwrite boolean default false
+) returns table
+( column_name text
+, description text
+)
+as $python$
+    #ADD-PYTHON-LIB-DIR
+    import ai.semantic_catalog
+    yield from ai.semantic_catalog.generate_column_descriptions(plpy, relation, catalog_name, config, save, overwrite)
```
jgpruitt, line 154: whoa! i'd never seen `yield from` before. neat.

projects/extension/ai/semantic_catalog.py
```diff
@@ -0,0 +1,377 @@
...
+from typing import TypedDict
+
+
+class GeneratedDescription(TypedDict):
+    name: str
+    description: str
+
+
+def render_obj_sample(plpy, relation: str) -> str:
+    result = plpy.execute(f"select * from {relation}", 5)
```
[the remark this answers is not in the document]
jgpruitt: We may want to use the TABLESAMPLE clause in the future. LIMIT will only get us the most recent rows. TABLESAMPLE can get us a more random sample efficiently

projects/extension/ai/semantic_catalog.py
```diff
@@ -0,0 +1,383 @@
...
+
+    ret_obj = f"""<data id="{relation}">"""
+    for r in result:
+        ret_obj += f"""\n  insert into {relation} ({', '.join(r.keys())}) values ({', '.join([str(v) for v in r.values()])});"""
```
jgpruitt, line 16: Use the utility functions to quote the table name and quote the values as literals.

- plpy.quote_ident()
- plpy.quote_literal()
- plpy.quote_nullable()

https://www.postgresql.org/docs/current/plpython-util.html

projects/extension/ai/semantic_catalog.py
```diff
@@ -0,0 +1,383 @@
...
+    ret_obj += "\n</data>"
+
+    return ret_obj
+
+
+def get_parsed_config(plpy, catalog_name: str, config: dict | None) -> dict:
+    if config is None:
+        result = plpy.execute(
+            f"select x.text_to_sql from ai.semantic_catalog x where x.catalog_name = '{catalog_name}';"
+        )
```
jgpruitt, line 26: we should use a parameterized query and pass catalog_name as a parameter

```suggestion
        plan = plpy.plan(
            f"select x.text_to_sql from ai.semantic_catalog x where x.catalog_name = $1;", ["text"]
        )
        result = plpy.execute(plan, [catalog_name], 1)
```

projects/extension/ai/semantic_catalog.py
```diff
@@ -0,0 +1,383 @@
...
+    plpy, relation: str | None = None, fn: str | None = None
+) -> str:
+    objid = (
+        f"'{relation}'::pg_catalog.regclass::pg_catalog.oid"
+        if relation
+        else f"'{fn}'::regprocedure::pg_catalog.oid"
+    )
+    result = plpy.execute(
+        f"select ai.render_semantic_catalog_obj(0, 'pg_catalog.pg_class'::pg_catalog.regclass::pg_catalog.oid, {objid}) as description;"
+    )
```
jgpruitt, line 46: `pg_proc` will be needed instead of `pg_class` for functions

jgpruitt, line 46: It might be better to have two separate queries. One for functions/procs. One for tables/views. Let's use parameterized queries instead of format strings for safety.

projects/extension/ai/semantic_catalog.py
```diff
@@ -0,0 +1,383 @@
...
+            response["choices"][0]["message"]["tool_calls"][0]["function"]["arguments"]
+        )["description"]
+    elif provider == "cohere":
+        pass
+    else:
+        raise Exception(f"provider {provider} not found")
+
+    if save:
+        result = plpy.execute(
+            f"select 1 from ai.semantic_catalog_obj x where x.objsubid = 0 and x.classid = 'pg_catalog.pg_class'::pg_catalog.regclass::pg_catalog.oid and x.objid = '{relation}'::pg_catalog.regclass::pg_catalog.oid"
```
jgpruitt, line 151: `pg_proc` instead of `pg_class` for functions

jgpruitt: oh you have a separate function for function descriptions. nvm

projects/extension/sql/idempotent/905-semantic-catalog-render.sql
```diff
@@ -290,3 +290,17 @@ as $func$
 $func$ language sql stable security invoker
 set search_path to pg_catalog, pg_temp
 ;
+
+-------------------------------------------------------------------------------
+-- render_obj_sample
+create or replace function ai.render_obj_sample
```
jgpruitt, line 296: I think "render_table_sample" or just "render_sample" might be better understood by users

projects/extension/ai/semantic_catalog.py
jgpruitt, line 1: We need to use parameterized queries instead of string manipulation throughout.

jgpruitt (changes requested):

projects/extension/ai/semantic_catalog.py
MasterOdin: Done in 4967a3d51918ff09a597e0f28a1e6daa6a6826a6

projects/extension/sql/idempotent/905-semantic-catalog-render.sql
MasterOdin: Done in 6be2c632acad662561e3846611ec93f91ca7a313

projects/extension/ai/semantic_catalog.py
MasterOdin: Done in 4967a3d51918ff09a597e0f28a1e6daa6a6826a6

projects/extension/ai/semantic_catalog.py
MasterOdin: Fixed in 4967a3d51918ff09a597e0f28a1e6daa6a6826a6, where I've made two queries, one for tables/views, one for functions. Makes it easier to make these sorts of tweaks without having a bunch of string concatentation.

projects/extension/ai/semantic_catalog.py
MasterOdin: Done in 4967a3d51918ff09a597e0f28a1e6daa6a6826a6.

I don't think there's a `plpy.plan` function, used `plpy.prepare` instead.

projects/extension/ai/semantic_catalog.py
MasterOdin: Reworked this in 4967a3d51918ff09a597e0f28a1e6daa6a6826a6. Though, I wonder if don't need to use `plpy.quote_ident` for `relation`, given that it's a fully quantified `schema_name.table_name` structure. If need to use it, then need to first split into two separate parts, and then apply `plpy.quote_ident` to each part.

I did the latter in 4967a3d51918ff09a597e0f28a1e6daa6a6826a6, happy to revise though.

projects/extension/ai/semantic_catalog.py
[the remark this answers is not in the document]
MasterOdin: Updated to use `plpy.quote_ident` in 4967a3d51918ff09a597e0f28a1e6daa6a6826a6.

I'm not sure exactly the best way to get `TABLESAMPLE` to work given that its argument is a percentage of the table that should be returned, so we'd need to run a count I guess, and then calculate the percentage off that, but not sure if that count query could be expensive...

projects/extension/sql/idempotent/905-semantic-catalog-render.sql
```diff
@@ -290,3 +290,17 @@ as $func$
...
+( relation pg_catalog.regclass
```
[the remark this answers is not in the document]
MasterOdin: done in b25659275e3c3bc23f44ad59a5e42427b5c6a615 / a04271c6b957c1708d8c45ca5f2c66dbcb59af65.

projects/extension/ai/semantic_catalog.py
```diff
@@ -0,0 +1,433 @@
...
+    result = plpy.execute(f"select * from {ident}", total)
+
+    ret_obj = f"""<data id="{relation}">"""
+    for r in result:
+        values = []
+        for v in r.values():
+            if isinstance(v, str):
+                values.append(plpy.quote_nullable(v))
+            else:
+                values.append(str(v))
```
jgpruitt, line 22: This still has some issues.

I did this:
```
create table bob
( id int not null primary key generated always as identity
, a text
, b bool
, c timestamptz default now()
, d date
, e float8 default random()
);
insert into bob (a, b, d) values ('foo', true, 'today'::date);
insert into bob (a, b, d) values ('bar', false, 'yesterday'::date);
insert into bob (a, b, d) values (null, null, null);
```
And render_sample yielded this:
```
select ai.render_sample('bob'::regclass);
                                                               render_sample
--------------------------------------------------------------------------------------------------------------------------------------------
 <data id="public.bob">                                                                                                                    +
   insert into public.bob (id, a, b, c, d, e) values (1, 'foo', True, '2025-02-06 19:59:24.705332+00', '2025-02-06', 0.25927368863286215); +
   insert into public.bob (id, a, b, c, d, e) values (2, 'bar', False, '2025-02-06 19:59:42.785997+00', '2025-02-05', 0.31713476013543906);+
   insert into public.bob (id, a, b, c, d, e) values (3, None, None, '2025-02-06 20:00:01.591488+00', None, 0.17700116924718712);          +
 </data>
(1 row)
```
We are getting `None` instead of `null` and python `True`/`False` instead of SQL `true`/`false`.

projects/extension/ai/semantic_catalog.py
```diff
@@ -0,0 +1,433 @@
...
+        )
+        result = plpy.execute(
+            plan, [model, json.dumps(messages), json.dumps(map_tools_to_openai(tools))]
+        )
+        response = json.loads(result[0]["openai_chat_complete"])
+        description = json.loads(
+            response["choices"][0]["message"]["tool_calls"][0]["function"]["arguments"]
+        )["description"]
+    elif provider == "cohere":
+        pass
```
jgpruitt, line 170: raise exception instead?

projects/extension/ai/semantic_catalog.py
```diff
@@ -0,0 +1,433 @@
...
+            """,
+            ["text", "jsonb", "text", "jsonb"],
+        )
+        result = plpy.execute(
+            plan, [model, json.dumps(messages), system_prompt, json.dumps(tools)]
+        )
+        response = json.loads(result[0]["anthropic_generate"])
+        description = response["content"][0]["input"]["description"]
+    elif provider == "ollama":
+        pass
```
jgpruitt, line 144: raise exception instead?

projects/extension/ai/semantic_catalog.py
```diff
@@ -0,0 +1,433 @@
...
+        )
+        result = plpy.execute(
+            plan, [model, json.dumps(messages), system_prompt, json.dumps(tools)]
+        )
+        columns = json.loads(result[0]["anthropic_generate"])["content"][0]["input"][
+            "columns"
+        ]
+
+    elif provider == "ollama":
+        pass
```
jgpruitt, line 279: raise exception

projects/extension/ai/semantic_catalog.py
```diff
@@ -0,0 +1,433 @@
...
+        )
+        result = plpy.execute(
+            plan, [model, json.dumps(messages), json.dumps(map_tools_to_openai(tools))]
+        )
+        response = json.loads(result[0]["openai_chat_complete"])
+        columns = json.loads(
+            response["choices"][0]["message"]["tool_calls"][0]["function"]["arguments"]
+        )["columns"]
+    elif provider == "cohere":
+        pass
```
jgpruitt, line 305: raise exception

projects/extension/ai/semantic_catalog.py
```diff
@@ -0,0 +1,433 @@
...
+        )
+        result = plpy.execute(
+            plan, [model, json.dumps(messages), json.dumps(map_tools_to_openai(tools))]
+        )
+        response = json.loads(result[0]["openai_chat_complete"])
+        description = json.loads(
+            response["choices"][0]["message"]["tool_calls"][0]["function"]["arguments"]
+        )["description"]
+    elif provider == "cohere":
+        pass
```
jgpruitt, line 418: raise exception

projects/extension/ai/semantic_catalog.py
```diff
@@ -0,0 +1,433 @@
...
+            """,
+            ["text", "jsonb", "text", "jsonb"],
+        )
+        result = plpy.execute(
+            plan, [model, json.dumps(messages), system_prompt, json.dumps(tools)]
+        )
+        response = json.loads(result[0]["anthropic_generate"])
+        description = response["content"][0]["input"]["description"]
+    elif provider == "ollama":
+        pass
```
jgpruitt, line 392: raise exception

jgpruitt (changes requested):
A few small changes. Let's get it merged and iterate from there.

projects/extension/ai/semantic_catalog.py
MasterOdin: Good catch, fixed in 27747f75cd726be51e1c56339324b22fde52f504.

projects/extension/ai/semantic_catalog.py
MasterOdin: Done in 27853696d476c88c7c8ae47fa003081e711dc8ed

projects/extension/ai/semantic_catalog.py
MasterOdin: Done in 27853696d476c88c7c8ae47fa003081e711dc8ed

projects/extension/ai/semantic_catalog.py
MasterOdin: Done in 27853696d476c88c7c8ae47fa003081e711dc8ed

projects/extension/ai/semantic_catalog.py
MasterOdin: Done in 27853696d476c88c7c8ae47fa003081e711dc8ed

projects/extension/ai/semantic_catalog.py
MasterOdin: Done in 27853696d476c88c7c8ae47fa003081e711dc8ed

projects/extension/ai/semantic_catalog.py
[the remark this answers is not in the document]
jgpruitt: yeah. let's stick with what you have. It will work with views and TABLESAMPLE won't

projects/extension/ai/semantic_catalog.py
```diff
@@ -0,0 +1,441 @@
...
+            if isinstance(v, str) or v is None:
+                values.append(plpy.quote_nullable(v))
+            elif isinstance(v, bool):
+                values.append("true" if v else "false")
+            else:
+                values.append(str(v))
+        ret_obj += f"""\n  insert into {ident} ({', '.join(plpy.quote_ident(key) for key in r.keys())}) values ({', '.join(values)});"""
+    ret_obj += "\n</data>"
+
+    return ret_obj
```
jgpruitt, line 28: I still think there are issues here, but let's merge and fix later. I think using SQL to render will be the most bulletproof approach.

jgpruitt (approved):

PR #416, opened 2025-02-03
fix(openai): allow reasoning models (o1, o3) to work with chat completion

smoya:
# Description

Starting from o1, all o* models (o1 and o3 atm) deprecated the `max_tokens` parameter in favor of `max_completion_tokens`. See https://platform.openai.com/docs/guides/reasoning#controlling-costs

Without this fix, we are not able to call `ai.openai_chat_complete` function with any of those models: 

```
ERROR:  openai.BadRequestError: Error code: 400 - {'error': {'message': "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.", 'type': 'invalid_request_error', 'param': 'max_tokens', 'code': 'unsupported_parameter'}}
```


projects/extension/tests/test_openai.py
```diff
@@ -284,6 +284,32 @@ def test_openai_chat_complete(cur, openai_api_key):
     assert actual is True
 
 
+def test_openai_chat_complete_with_new_o_models(cur, openai_api_key):
```
smoya, line 287: BTW, tried to unify some of the tests but ended up having a bit less of lines but adding a lot of complexity penalizing readability.
Sometimes, duplication is good :)

projects/extension/sql/idempotent/001-openai.sql
```diff
@@ -242,7 +242,7 @@ as $python$
         logit_bias=ai.openai.str_arg_to_dict(logit_bias),
         logprobs=logprobs,
         top_logprobs=top_logprobs,
-        max_tokens=max_tokens,
+        max_completion_tokens=max_tokens,
```
jgpruitt, line 245: Do the old models still take max_tokens? Should we provide both as parameters to our function?

projects/extension/tests/test_openai.py
```diff
@@ -284,6 +284,32 @@ def test_openai_chat_complete(cur, openai_api_key):
...
+    cur.execute(
+        """
+        with x as
+        (
+          select ai.openai_chat_complete
```
jgpruitt, line 292: Should we have a test that passes max_tokens to try it out?

jgpruitt (changes requested):

projects/extension/sql/idempotent/001-openai.sql
smoya: `max_completion_tokens` is retrocompatible with all models. No need to pass `max_tokens` anymore.

smoya: @jgpruitt There is the need to use one or another param based on the model; for reasoning models `max_completion_tokens` should be used, `max_token` otherwise.

I used a regex to match reasoning models, if you feel there is a better alternative please share.
I discarded the following matching approaches:

- Matching literal `o1`, `o1-mini`, `o3-mini` because whenever a new reasoning o* model goes out, it won't be supported.
- Reverse matching previous models. The list is so big and there seems to be variants of those models (including datetimes in the slug name).

jgpruitt (changes requested):
I would prefer that we make both `max_tokens` and `max_completion_tokens` parameters to the function and not try to do any mapping between the two. We can rely on the OpenAI library/API to catch incorrect usage and supply error messages. If we try to do this for users, I think we'll be chasing the changes to OpenAI models. WDYT?

smoya:
> I would prefer that we make both `max_tokens` and `max_completion_tokens` parameters to the function and not try to do any mapping between the two. We can rely on the OpenAI library/API to catch incorrect usage and supply error messages. If we try to do this for users, I think we'll be chasing the changes to OpenAI models. WDYT?

@jgpruitt It's a thing now !

projects/extension/ai/vectorizer.py
```diff
@@ -8,7 +8,7 @@
 from .utils import get_guc_value
 
 GUC_VECTORIZER_URL = "ai.external_functions_executor_url"
-DEFAULT_VECTORIZER_URL = "http://localhost:8000"
+DEFAULT_VECTORIZER_URL = "http://0.0.0.0:8000"
```
smoya, line 11: This change is fixing CI errors connecting to the test server.

projects/extension/tests/test_openai.py
jgpruitt: yeah. duplication gets a bad rap

jgpruitt (approved):

projects/extension/ai/vectorizer.py
jgpruitt: nice catch!

PR #418, opened 2025-02-04
feat(chunking): add chunking functions to the extension

jgpruitt:
Adds two new functions to pgai to allow for in-database chunking.

1. `ai.chunk_text`
2. `ai.chunk_text_recursively`


docs/chunking.md
```diff
@@ -0,0 +1,107 @@
...
+## chunk_text
+
+Splits text into chunks using a separator
+
+| Name                | Type  | Default | Required | Description                                               |
+|---------------------|-------|---------|----------|-----------------------------------------------------------|
+| input               | text  | -       | ✔        | The text to split into chunks                             |
+| chunk_size          | int   | -       | ✖        | The target size of a chunk in characters                  |
+| chunk_overlap       | int   | -       | ✖        | The target amount of overlapping characters in each chunk |
+| separator           | text  | -       | ✖        | The text to split on                                      |
```
JamesGuthrie, line 40: I find this somewhat confusing. Under the hood we're using LangChain's CharacterTextSplitter, so while _we_ don't set a default value in the `chunk_text` function, the defaults for these are `chunk_size = 4000`, `chunk_overlap = 200`, and `separator = '\n\n'`. I would suggest that we surface both the fact that we're using CharacterTextSplitter, and what the actual defaults are.

docs/chunking.md
```diff
@@ -0,0 +1,107 @@
...
+## chunk_text_recursively
+
+Recursively splits text into chunks using multiple separators in sequence.
+
+| Name               | Type   | Default | Required | Description                                               |
+|--------------------|--------|---------|----------|-----------------------------------------------------------|
+| input              | text   | -       | ✔        | The text to split into chunks                             |
+| chunk_size         | int    | -       | ✖        | The target size of a chunk in characters                  |
+| chunk_overlap      | int    | -       | ✖        | The target amount of overlapping characters in each chunk |
+| separators         | text[] | -       | ✖        | An array of texts to split on                             |
```
JamesGuthrie, line 78: Same comment as above, but for RecursiveCharacterTextSplitter.

JamesGuthrie (changes requested):

docs/chunking.md
jgpruitt: done

JamesGuthrie (approved):

PR #419, opened 2025-02-04
chore: internal refactor build.py

JamesGuthrie:
Introduces the Actions class which acts as a container for all public-facing actions available in build.py.

To add a new action to the build.py, one need only add a new member function to the Actions class.

### DIFF
--- a/projects/extension/build.py
+++ b/projects/extension/build.py
@@ -8,37 +8,479 @@ import subprocess
 import sys
 import tempfile
 import textwrap
+from collections import OrderedDict
 from pathlib import Path
 
-HELP = """Available targets:
-- help             displays this message and exits
-- build-install    runs build followed by install
-- install          installs the project
-- install-sql      installs the sql files into the postgres installation
-- install-prior-py installs the extension's python package for prior versions
-- install-py       installs the extension's python package
-- uninstall        uninstalls the project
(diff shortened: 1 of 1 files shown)


adolsalamanca (approved):

projects/extension/build.py
```diff
@@ -8,37 +8,479 @@
...
-- lint             runs both sql and python linters
-- format-py        runs ruff to check formatting of the python source files
-- reformat-py      runs ruff to update the formatting of the python source files
-- docker-build     builds the dev docker image
-- docker-run       launches a container in docker using the docker image
-- docker-stop      stops the container
-- docker-rm        deletes the dev container
-- run              builds+runs the dev container and installs the extension"""
+
+class BuildPyActions:
```
jgpruitt, line 15: nit: WDYT about naming the class just "Actions"? 

```
        BuildPyActions.install_prior_py()
        BuildPyActions.install_py()
        BuildPyActions.install_sql()
```

would be. which for some reason feels much less intrusive to my brain

```
        Actions.install_prior_py()
        Actions.install_py()
        Actions.install_sql()
```

no biggie though

jgpruitt (approved):

projects/extension/build.py
JamesGuthrie: Yes, done!

PR #422, opened 2025-02-04
feat: add raw response variants for OpenAI functions

alejandrodnm:
This PR continues the work from https://github.com/timescale/pgai/pull/219 . That PR was deemed to big so we are splitting the functionality in smaller PRs.

### feat: add raw response variants for OpenAI functions 
 
 Add _with_raw_response variants for all OpenAI functions that return the
raw API response as JSONB. These variants accept the same arguments as
their regular counterparts, making it easy to access raw API responses
when needed.

New methods:

- ai.openai_list_models_with_raw_response
- ai.openai_embed_with_raw_response
- ai.openai_moderate_with_raw_response
- ai.openai_chat_complete_with_raw_response


projects/extension/ai/openai.py
```diff
@@ -88,3 +88,32 @@ def embed(
...
+
+    kwargs = create_kwargs(
+        dimensions=dimensions,
+        user=user,
+        extra_headers=str_arg_to_dict(extra_headers),
+        extra_query=str_arg_to_dict(extra_query),
+        extra_body=str_arg_to_dict(extra_body),
+        timeout=timeout,
+    )
+    response = client.embeddings.with_raw_response.create(
```
jgpruitt, line 116: `with_raw_response` is part of the OpenAI client? Is it undocumented? I can't find it.

jgpruitt (approved):
My only nit is that "with_raw_response" makes for super long function names, but if that mirrors what's in the OpenAI client, then I guess it's okay

cevian (approved):
kinda surprised that the text->jsonb conversion happens so seamlessly but if it works, it works

projects/extension/tests/test_openai.py
```diff
@@ -58,6 +58,24 @@ def test_openai_list_models(cur, openai_api_key):
...
+            api_key=>%s,
+            extra_headers=>'{"X-Custom-Header": "my-value"}',
+            extra_query=>'{"debug": true}',
+            timeout=>600::float8
+        )
+    """,
+        (openai_api_key,),
+    )
+    actual = cur.fetchone()[0]
+    print(actual)
```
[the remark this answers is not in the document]
alejandrodnm: no

projects/extension/ai/openai.py
alejandrodnm: Yeah it's in the readme

https://github.com/openai/openai-python?tab=readme-ov-file#accessing-raw-response-data-eg-headers

projects/extension/ai/openai.py
```diff
@@ -88,3 +88,32 @@ def embed(
         return None
     for obj in response.data:
         yield obj.index, obj.embedding
+
+
+def embed_with_raw_response(
```
smoya, line 93: Just wondering if we could reuse code from the `embed` function or the other way around.
Perhaps, `embed` function could call this new one and just return the parsed response `response.parse()`.

PR #424, opened 2025-02-04
build: move built sql files to an 'output' directory

jgpruitt:
Most of the time, we don't need to look at the SQL files that are created by `just build`. They are build artifacts. Most of the time we are looking at the source files in the `idempotent` and `incremental` directories. Previously, the built SQL files were placed in the `sql` directory, but since we now have quite a few of them, this was becoming obtrusive.

This PR moves the built SQL files to an `output` directory to get them "out of the way".

### Before:
image

### After:
image


smoya (approved):
Small change but big improvement on ux for those dealing with this code every day.

Thks!💯💯

PR #420, opened 2025-02-04
feat: add support for openai extra request parameters (headers, body, query)

alejandrodnm:
This PR continues the work from https://github.com/timescale/pgai/pull/219 . That PR was deemed to big so we are splitting the functionality in smaller PRs.

### feat: add support for openai extra request parameters (headers, body, query) 
 
 Add support for specifying extra_headers, extra_body, and extra_query
arguments in OpenAI function calls:

- ai.openai_list_models
- ai.openai_embed
- ai.openai_chat_complete
- ai.openai_moderate

These parameters are passed through to the
underlying OpenAI library methods that support them, enabling more
flexible request customization.

This change allows users to:

- Add custom HTTP headers via extra_headers
- Include additional body parameters via extra_body
- Specify extra query parameters via extra_query

All parameters are passed unchanged to the corresponding OpenAI library
methods.

PR #426, opened 2025-02-05
Add openai client config arguments

alejandrodnm:
This PR continues the work from https://github.com/timescale/pgai/pull/219 . That PR was deemed to big so we are splitting the functionality in smaller PRs.

### feat!: consolidate OpenAI client configuration 
 
 BREAKING CHANGE: Replace individual client configuration parameters with
a single `client_config` parameter. The `base_url` parameter has been
moved into the new configuration structure.

- Add new `ai.openai_client_config()` function to generate client
  configurations
- Remove `base_url` parameter from all OpenAI functions
- Add `client_config` parameter to all OpenAI functions
- Support additional client options: timeout, organization, project,
  max_retries, default_headers, default_query

Migration:
Old:

```sql
  SELECT ai.openai_embed(
    model => 'model',
    input_text => 'text',
    base_url => 'url'
  );
```

New:

```sql
  SELECT ai.openai_embed(
    model => 'model',
    input_text => 'text',
    client_config => ai.openai_client_config(base_url => 'url')
  );
 ```

### CODE REVIEW

**Comment by alejandrodnm** [comment]
This is currently using #422 as based, I'll change it to main once that's merged.

File: projects/extension/sql/idempotent/001-openai.sql [hunk 1]
```diff
@@ -599,3 +600,37 @@ $python$
...
 set search_path to pg_catalog, pg_temp
 ;
+
+
+-------------------------------------------------------------------------------
+-- openai_client_config
+-- Generate a JSON object with the configuration for the OpenAI client.
+create or replace function ai.openai_client_config
+( base_url text default null
+, timeout float8 default null
```
jgpruitt, line 610: is timeout in seconds or what? Should this be an `interval` instead of `float8`?

projects/extension/sql/incremental/014-extra-openai-args.sql
jgpruitt, line 1: Wait. Why isn't there a `drop function` for each of the openai functions? We changed the signature of all of them didn't we?

jgpruitt (changes requested):

projects/extension/sql/incremental/014-extra-openai-args.sql
alejandrodnm: I merged a recent PR that added the migration. Since everything will go in the same release, it's not needed in this PR.

https://github.com/timescale/pgai/pull/420/files#diff-db6ba9af95c071ad682926141401de6151d0d7eb953c00e00e99fe80eb45da11R1-R27

projects/extension/sql/idempotent/001-openai.sql
alejandrodnm: I put a float because it's what openai uses 

https://github.com/timescale/pgai/pull/420/files#diff-db6ba9af95c071ad682926141401de6151d0d7eb953c00e00e99fe80eb45da11R1-R27

Do you know if the interval is translated to `float` when it's converted to plpython?

jgpruitt: i'm confused as to why the upgrade tests aren't failing on this branch though.

projects/extension/sql/incremental/014-extra-openai-args.sql
jgpruitt: i'm confused as to why the upgrade tests aren't failing on this branch though.

projects/extension/sql/idempotent/001-openai.sql
jgpruitt: https://www.postgresql.org/docs/current/plpython-data.html#PLPYTHON-DATA-TYPE-MAPPING

`interval` gets converted to `str` in Python, unfortunately.

jgpruitt: Okay, I had to hunt through the OpenAI implementation to find that this comes from the httpx library. Then, I had to hunt through the httpx library's examples to infer that this is in seconds.

Can we name this `timeout_seconds` or similar to save our users from having to either guess or do the same exercise?

jgpruitt: or make it an `interval` and handle the conversion to seconds

projects/extension/sql/incremental/014-extra-openai-args.sql
alejandrodnm: Because main already has the drop of the old functions. So we are just creating the new functions.

projects/extension/sql/idempotent/001-openai.sql
alejandrodnm: I'm going with `timeout_seconds`. Since the interval gets passed as str, I don't think it's safe to parse it manually. You can strings like '00:00:01', '00:00:01.45', '1 day', '1 mon'.

adolsalamanca (approved):

jgpruitt (approved):

PR #431, opened 2025-02-05
feat: allow vectorizer to read documents via smart-open parse via pymupdf

Askir:
Uses our agreed upon API spec number 4.
Works end to end for local files and s3 (credential loading also works, see tests)

What's missing?
- Validation in create_vectorizer (currently there is no check that you use an empty chunking column with a file parser so if you configure one that will crash the vectorizer)
- Docs
- Fixing the tests in the extension (I did the minimal thing to get the vectorizer working)


projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -762,7 +768,14 @@ async def _generate_embeddings(
...
+            if self.vectorizer.config.loader is not None:
+                file_content = self.vectorizer.config.loader.load(item)
+                markdown_content = self.vectorizer.config.parser.parse_file_content(
+                    file_content
+                )
+                chunks = self.vectorizer.config.chunking.into_chunks(markdown_content)
+            else:
+                chunks = self.vectorizer.config.chunking.into_chunks(item)
             for chunk_id, chunk in enumerate(chunks, 0):
                 formatted = self.vectorizer.config.formatting.format(chunk, item)
```
cevian, line 780: I think it may be best to conceive of this as a processing pipeline where most processing nodes alter the row

```suggestion
            if self.vectorizer.config.loader is not None:
                # loader reads a "url" column and loads binary data into item['file_contents']
                self.vectorizer.config.loader.load(item)
            if self.vectorizer.config.parser is not None:
               # reads some column (default: file_contents) and puts result in item['parsed_file_content']
               self.vectorizer.config.parser.parse_file_content(item)
            chunks = self.vectorizer.config.chunking.into_chunks(item)
            for chunk_id, chunk in enumerate(chunks, 0):
                formatted = self.vectorizer.config.formatting.format(chunk, item)
                records_without_embeddings.append(pk + [chunk_id, formatted])
```

projects/extension/sql/idempotent/013-loader.sql
```diff
@@ -0,0 +1,14 @@
+-------------------------------------------------------------------------------
+-- loader_file_loader
+create or replace function ai.loader_from_document
```
jgpruitt, line 3: naming considerations:

1. our other configs for vectorizer end in `ing`: `embedding`, `formatting`, `chunking`. Should we follow the same convention?
2. would `object_storage` be more appropriate than `document`? We should describe "where" instead of "what". We may load documents from other places in the future

projects/extension/sql/idempotent/014-parser.sql
```diff
@@ -0,0 +1,11 @@
+-------------------------------------------------------------------------------
+-- parser_auto
+create or replace function ai.parser_auto() returns pg_catalog.jsonb
```
jgpruitt, line 3: `parsing` vs `parser` to follow the existing convention?

projects/extension/sql/idempotent/005-chunking.sql
```diff
@@ -2,7 +2,7 @@
 -------------------------------------------------------------------------------
 -- chunking_character_text_splitter
 create or replace function ai.chunking_character_text_splitter
-( chunk_column pg_catalog.name
+( chunk_column pg_catalog.name default ''
```
jgpruitt, line 5: i think this column will go away completely when we have a `loader_row` configuration, but in case it doesn't the default should be `null` rather than `''`

jgpruitt (changes requested):

smoya:
Closing in favor of the WIP work in https://github.com/timescale/pgai/pull/442

PR #434, opened 2025-02-05
docs: Add docs about s3 document integration

Askir:


Askir:
We're changing the API a bit more so closing this for now

PR #435, opened 2025-02-05
feat!: only install current extension version

JamesGuthrie:
Running `python3 build.py install` would install all versions of the extension (and their python dependencies).

This commit changes this behaviour to only install the sql scripts (including upgrade paths) and python dependencies for the current version of the extension.

For backwards compatibility, use `build.py install all`, which activates the old behaviour.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
         run: docker exec pgai-ext just check-requirements
 
       - name: Install extension
-        run: docker exec pgai-ext just install
+        run: docker exec pgai-ext just install-all
 
       - name: Run test server
         run: docker exec -d
--- a/projects/extension/DEVELOPMENT.md
+++ b/projects/extension/DEVELOPMENT.md
@@ -56,7 +56,7 @@ To make changes to pgai:
 
          ```bash
          just ext build
-         just ext install
+         just ext install-all
          ```
 
       1. Run the unit tests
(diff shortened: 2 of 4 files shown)


JamesGuthrie:
One open question for reviewers: If developers follow the existing documentation the upgrade tests will fail, because we don't specify that one should run `just build install-all` in the docker container. The upside is that we only install one extension version. I'm not sure which is the better option.

smoya:
> One open question for reviewers: If developers follow the existing documentation the upgrade tests will fail, because we don't specify that one should run `just build install-all` in the docker container. The upside is that we only install one extension version. I'm not sure which is the better option.

If understand this correctly, we just need to change the DEVELOPMENT.md file replacing the  `just install` with `just install-all`. I don't see an issue here, but perhaps I'm missing something extra.

smoya (approved):

jgpruitt:
> One open question for reviewers: If developers follow the existing documentation the upgrade tests will fail, because we don't specify that one should run `just build install-all` in the docker container. The upside is that we only install one extension version. I'm not sure which is the better option.

I think having instructions that "just work" is most important.

JamesGuthrie:
> > One open question for reviewers: If developers follow the existing documentation the upgrade tests will fail, because we don't specify that one should run `just build install-all` in the docker container. The upside is that we only install one extension version. I'm not sure which is the better option.
> 
> I think having instructions that "just work" is most important.

I have updated the code in the meantime.

jgpruitt:
This will require changes to the docker image repos, right?

JamesGuthrie:
> This will require changes to the docker image repos, right?

yes

PR #437, opened 2025-02-05
chore!: drop versions 0.1.0, 0.2.0, 0.3.0

JamesGuthrie:
We previously deprecated these versions, and there is no upgrade path from them to version 0.4.0. In order to upgrade you must drop the pgai extension and re-create it.


projects/extension/build.py
```diff
@@ -240,9 +236,6 @@ def build_sql() -> None:
             wf.flush()
             wf.close()
         for prior_version in prior_versions():
-            if prior_version in deprecated_versions():
-                # we don't allow upgrades from these versions. they are deprecated
-                continue
```
jgpruitt, line 245: can we leave this?

projects/extension/build.py
```diff
@@ -507,14 +497,6 @@ def prior_versions() -> list[str]:
...
 
 
-def deprecated_versions() -> set[str]:
-    return {
-        "0.3.0",  # deprecated
-        "0.2.0",  # deprecated
-        "0.1.0",  # deprecated
-    }
-
-
```
jgpruitt, line 517: can we leave this and return an empty set? then we're set up if we need to deprecate versions in the future

jgpruitt (approved):
left a couple of notes, but lgtm!

projects/extension/build.py
JamesGuthrie: yes! done!

projects/extension/build.py
JamesGuthrie: yes! done.

jgpruitt (approved):

PR #442, opened 2025-02-06
feat: add support for generating embeddings for external documents

smoya:
## Summary

  - Added document loading and parsing capabilities to vectorizers, supporting various file formats (PDF, XLSX, HTML, EPUB, etc.)
  - Implemented loading_uri to fetch documents from local files or remote storage (S3)
  - Added automated parsing via parsing_auto to intelligently process different document types
  - Integrated document retry mechanism for handling transient failures
  - Created example for generating embeddings directly from documents

## Test plan

  - Added comprehensive test suite for document loading and parsing
  - Added example with various document types to demonstrate functionality
  - Testing all document formats: PDF, XLSX, HTML, EPUB, Markdown
  - Verified retry mechanisms work correctly with proper error handling


projects/extension/sql/idempotent/012-vectorizer-int.sql
```diff
@@ -405,7 +405,10 @@ declare
 begin
     -- create the table
     select pg_catalog.format
-    ( $sql$create table %I.%I(%s, queued_at timestamptz not null default now())$sql$
+    ( $sql$create table %I.%I(%s
+    , queued_at pg_catalog.timestamptz not null default now()
+    , retries pg_catalog.int4 default 0
+    , retry_after pg_catalog.timestamptz)$sql$
```
jgpruitt, line 411:
```suggestion
    ( $sql$
      create table %I.%I
      ( %s
      , queued_at pg_catalog.timestamptz not null default now()
      , retries pg_catalog.int4 default 0
      , retry_after pg_catalog.timestamptz
      )
      $sql$
```

incredibly nit picky. sorry.

projects/extension/sql/idempotent/013-loading.sql
```diff
@@ -0,0 +1,83 @@
...
+    , 'config_type': 'loading'
+    , 'column_name': column_name
+    )
+$func$ language sql immutable security invoker
+set search_path to pg_catalog, pg_temp
+;
+
+-------------------------------------------------------------------------------
+-- loading_document
+create or replace function ai.loading_document
```
jgpruitt, line 18: IMO, `loading_uri` would be a more appropriate name. It describes "where" we are loading from.

projects/extension/sql/idempotent/013-loading.sql
```diff
@@ -0,0 +1,83 @@
+-------------------------------------------------------------------------------
+-- loading_row
+create or replace function ai.loading_row
```
jgpruitt, line 3: IMO, `loading_column` would be a more appropriate name.

projects/extension/sql/idempotent/015-vectorizer-api.sql
```diff
@@ -133,11 +135,22 @@ begin
...
+    -- validate the loading config
+    perform ai._validate_loading(loading, _source_schema, _source_table);
+
+    -- validate the parsing config
+    perform ai._validate_parsing(jsonb_build_object(
+        'parsing', parsing,
+        'loading', loading,
+        'source_schema', _source_schema,
+        'source_table', _source_table
+    ));
```
jgpruitt, line 147: I would pass these elements as individual arguments rather than constructing a jsonb object.

projects/extension/sql/idempotent/014-parsing.sql
```diff
@@ -0,0 +1,113 @@
...
+    , 'config_type': 'parsing'
+    )
+$func$ language sql immutable security invoker
+set search_path to pg_catalog, pg_temp
+;
+
+-------------------------------------------------------------------------------
+-- _validate_parsing
+create or replace function ai._validate_parsing
+( config pg_catalog.jsonb  -- has to contain both loading and parsing config
```
jgpruitt, line 52: I would create distinct parameters for each element you need. This makes it more clear what the function expects and can better enforce that everything needed is provided and in the right types.

jgpruitt, line 52:
```suggestion
( loading pg_catalog.jsonb 
, parsing pg_catalog.jsonb
, source_schema pg_catalog.name
, source_table pg_catalog.name
```

projects/extension/sql/idempotent/900-semantic-catalog-init.sql
```diff
@@ -38,7 +38,7 @@ begin
     , processing=>processing
     , grant_to=>grant_to
     , formatting=>ai.formatting_python_template() -- TODO: this ain't gonna work
-    , chunking=>ai.chunking_recursive_character_text_splitter('description') -- TODO
+    , chunking=>ai.chunking_recursive_character_text_splitter() -- TODO
```
jgpruitt, line 41: add the appropriate loading config so this doesn't break?

projects/extension/sql/idempotent/900-semantic-catalog-init.sql
```diff
@@ -51,7 +51,7 @@ begin
...
```
jgpruitt, line 54: add the appropriate loading config so this doesn't break?

projects/extension/sql/idempotent/013-loading.sql
jgpruitt: It would also "feel more ergonomic". You're passing a column name to this function

```sql
select ai.loading_column('my_column');
```

projects/extension/tests/vectorizer/server.py
jgpruitt, line 1: Can we add classes for the loading and parsing configs? I know it's not strictly necessary, but it does serve as a canary to catch changes to the format.

projects/extension/sql/idempotent/013-loading.sql
```diff
@@ -0,0 +1,83 @@
...
+( column_name pg_catalog.text)
```
jgpruitt, line 4:
```suggestion
( content_column pg_catalog.name)
```
maybe? It better describes which column we expect to be provided.

Also, use the `name` type

projects/extension/sql/idempotent/013-loading.sql
```diff
@@ -0,0 +1,83 @@
...
+    , 'column_name': column_name
+    )
+$func$ language sql immutable security invoker
+set search_path to pg_catalog, pg_temp
+;
+
+-------------------------------------------------------------------------------
+-- loading_document
+create or replace function ai.loading_document
+( column_name pg_catalog.text)
```
jgpruitt, line 19:
```suggestion
( uri_column pg_catalog.name)
```
maybe? It better describes which column we expect to be provided

Also, use the `name` type

projects/extension/sql/idempotent/013-loading.sql
```diff
@@ -0,0 +1,83 @@
...
+create or replace function ai._validate_loading
+( config pg_catalog.jsonb
+, source_schema pg_catalog.name
+, source_table pg_catalog.name
+) returns void
+as $func$
+declare
+    _config_type pg_catalog.text;
+    _implementation pg_catalog.text;
+    _column_name pg_catalog.text;
```
jgpruitt, line 42:
```suggestion
    _column_name pg_catalog.name;
```

projects/extension/sql/idempotent/014-parsing.sql
```diff
@@ -0,0 +1,113 @@
...
+) returns void
+as $func$
+declare
+    _column_type pg_catalog.text;
```
jgpruitt, line 56:
```suggestion
    _column_type pg_catalog.name;
```

projects/extension/tests/vectorizer/test_loading.py
```diff
@@ -0,0 +1,110 @@
...
+            """,
+            "function ai.loading_row() does not exist",
+        ),
+        (
+            """
+            select ai._validate_loading
+            ( ai.loading_document(), 'public', 'thing' )
+            """,
+            "function ai.loading_document() does not exist",
+        ),
```
jgpruitt, line 85: To ensure that our type check works, add a "bad" test for loading_row and loading_document on a column of an unsupported type, like the `weight` column.

Also, add "bad" tests for a column that doesn't exist.

projects/extension/tests/vectorizer/test_parsing.py
```diff
@@ -0,0 +1,171 @@
...
+             'parsing', ai.parsing_pymupdf(),
+             'loading', ai.loading_row('document'),
+             'source_schema', 'public',
+             'source_table', 'thing'
+            )
+        )
+        """,
+    ]
+    bad = [
+        (
```
jgpruitt, line 102: add a case for a non-existent column

projects/extension/tests/vectorizer/test_vectorizer.py
```diff
@@ -260,37 +272,6 @@ def test_vectorizer_timescaledb():
...
-                f"select has_table_privilege('bob', '{vec.view_schema}.{vec.view_name}', 'select')"
-            )
-            actual = cur.fetchone()[0]
-            assert actual
-
-            # bob should have select on the vectorizer table
-            cur.execute("select has_table_privilege('bob', 'ai.vectorizer', 'select')")
-            actual = cur.fetchone()[0]
-            assert actual
-
```
jgpruitt, line 293: why are we nuking these?

projects/extension/tests/vectorizer/test_vectorizer.py
```diff
@@ -193,6 +203,7 @@ def test_vectorizer_timescaledb():
         db_url("test"), autocommit=True, row_factory=namedtuple_row
     ) as con:
         with con.cursor() as cur:
+            cur.execute("set statement_timeout = '5s'")
```
jgpruitt, line 206: why is this added?

projects/extension/tests/vectorizer/test_vectorizer.py
```diff
@@ -1807,11 +1808,221 @@ def test_create_vectorizer_privs():
...
+                create table vec.doc_bytea
+                ( id bigint not null primary key generated always as identity
+                , content bytea not null
+                )
+            """)
+
+            # Insert a sample PDF as bytea
+            cur.execute("""
+                insert into vec.doc_bytea(content)
+                values (decode('255044462D312E340A25', 'hex'))  -- Start of PDF file magic bytes
```
jgpruitt, line 1841: neat!

jgpruitt (changes requested):
First pass. Only looked at the extension.

projects/extension/tests/vectorizer/test_vectorizer.py
adolsalamanca: It's related to a non resolved issue with that test locally,
for some reason the SQL call to the run_job fails, 
and at least this way it fails fast, not after a few minutes.

projects/extension/tests/vectorizer/test_vectorizer.py
adolsalamanca: those were added before we did specific validations under privileges suite,
so that became redundant. 

would you rather prefer to have validations in both sides?

docs/vectorizer/api-reference.md
```diff
@@ -148,6 +152,191 @@ in other management functions.
...
+|------|------|---------|----------|-------------|
+| column_name | text | - | ✔ | The name of the column containing the file path. |
+
+#### Returns
+
+A JSON configuration object that you can use in [ai.create_vectorizer](#create-vectorizers).
+
+## Parsing configuration
+
+You use the parsing configuration functions in `pgai` to define the way data, after being loaded, is parsed. This is useful if you want to load data that is not immediately text and therefore needs to be parsed first. I.e. a PDF document.
```
JamesGuthrie, line 221:
```suggestion
You use the parsing configuration functions in `pgai` to define how data is parsed after document loading. This is useful if for non-textual formats such as PDF documents.
```

docs/vectorizer/api-reference.md
```diff
@@ -148,6 +152,191 @@ in other management functions.
...
+- [ai.parsing_auto](#aiparsing_auto)
+- [ai.parsing_none](#aiparsing_none)
+- [ai.parsing_docling](#aiparsing_docling)
+- [ai.parsing_pymupdf](#aiparsing_pymupdf)
+
+### ai.parsing_auto
+
+You use `ai.parsing_auto` to automatically parse the data with a fitting available parser.
+This special parser will decide on the fly which parser fits with each of the documents by guessing the file type. 
+If the type can not be guessed, the document will not be processed and an error will be appended to the vectorizer errors.
```
JamesGuthrie, line 234:
```suggestion
You use `ai.parsing_auto` to automatically select an appropriate parser based on detected file types.
Documents with unrecognizable formats won't be processed and will generate an error (in the `ai.vectorizer_errors` table.
```

docs/vectorizer/api-reference.md
```diff
@@ -148,6 +152,191 @@ in other management functions.
...
+| None | - | - | - | - |
+
+#### Returns
+
+A JSON configuration object that you can use in [ai.create_vectorizer](#create-vectorizers).
+
+
+### ai.parsing_none
+
+You use `ai.parsing_none` to load the data as is from the source table.
```
JamesGuthrie, line 261:
```suggestion
You use `ai.parsing_none` to load the data as-is from the source table.
```

docs/vectorizer/api-reference.md
```diff
@@ -148,6 +152,191 @@ in other management functions.
...
+
+| Name | Type | Default | Required | Description |
+|------|------|---------|----------|-------------|
+| None | - | - | - | - |
+
+#### Returns
+
+A JSON configuration object that you can use in [ai.create_vectorizer](#create-vectorizers).
+
+### ai.parsing_docling
```
JamesGuthrie, line 288: Should we provide a suggestion on when one should use docling vs pymypdf?

docs/vectorizer/api-reference.md
```diff
@@ -148,6 +152,191 @@ in other management functions.
...
+You use the parsing configuration functions in `pgai` to define the way data, after being loaded, is parsed. This is useful if you want to load data that is not immediately text and therefore needs to be parsed first. I.e. a PDF document.
+
+The parsing functions are:
+
+- [ai.parsing_auto](#aiparsing_auto)
+- [ai.parsing_none](#aiparsing_none)
+- [ai.parsing_docling](#aiparsing_docling)
+- [ai.parsing_pymupdf](#aiparsing_pymupdf)
+
+### ai.parsing_auto
```
JamesGuthrie, line 230: Should we document how the underlying parser is chosen?

projects/extension/sql/idempotent/005-chunking.sql
```diff
@@ -2,8 +2,7 @@
 -------------------------------------------------------------------------------
 -- chunking_character_text_splitter
 create or replace function ai.chunking_character_text_splitter
-( chunk_column pg_catalog.name
-, chunk_size pg_catalog.int4 default 800
+( chunk_size pg_catalog.int4 default 800
```
JamesGuthrie, line 5: Should we keep versions of these functions around with the old signature which throw an error telling the user to use the new API?

projects/extension/sql/idempotent/005-chunking.sql
```diff
@@ -26,8 +24,7 @@ set search_path to pg_catalog, pg_temp
 -------------------------------------------------------------------------------
 -- chunking_recursive_character_text_splitter
 create or replace function ai.chunking_recursive_character_text_splitter
-( chunk_column pg_catalog.name
-, chunk_size pg_catalog.int4 default 800
+( chunk_size pg_catalog.int4 default 800
```
JamesGuthrie, line 27: ditto

projects/extension/sql/idempotent/015-vectorizer-api.sql
```diff
@@ -17,8 +17,10 @@ set search_path to pg_catalog, pg_temp
 create or replace function ai.create_vectorizer
 ( source pg_catalog.regclass
 , destination pg_catalog.name default null
+, loading pg_catalog.jsonb default null
```
JamesGuthrie, line 20: Again, should we keep the old create_vectorizer function and change its implementation to throw an error pointing out the changed API?

docs/vectorizer/api-reference.md
```diff
@@ -148,6 +152,191 @@ in other management functions.
...
+
+A JSON configuration object that you can use in [ai.create_vectorizer](#create-vectorizers).
+
+
+### ai.parsing_none
+
+You use `ai.parsing_none` to load the data as is from the source table.
+
+> [!NOTE]  
+> This is the default parser; adding this parser or skipping the `parsing` parameter does the same effect.
```
JamesGuthrie, line 264: This statement is incorrect? The default parser is `parsing_auto`.

JamesGuthrie (commented):
Initial pass.

docs/vectorizer/api-reference.md
Askir: done ✅

docs/vectorizer/api-reference.md
Askir: done ✅

projects/extension/sql/idempotent/900-semantic-catalog-init.sql
Askir: done ✅

projects/extension/sql/idempotent/900-semantic-catalog-init.sql
Askir: done ✅

projects/extension/tests/vectorizer/server.py
Askir: done ✅  what is this actually used for though?

projects/extension/sql/idempotent/013-loading.sql
Askir: I didn't rename the parameter name I feel like with `loading_column(column_name => '')` makes a lot of sense.

projects/extension/tests/vectorizer/test_loading.py
Askir: done ✅

projects/extension/tests/vectorizer/test_parsing.py
Askir: the parsing validation doesn't check that but it's in the loading validation I think that's fine?

projects/extension/sql/idempotent/005-chunking.sql
adolsalamanca: thanks, as it has been decided to move forward instead, marking this as resolved.

projects/extension/sql/idempotent/015-vectorizer-api.sql
adolsalamanca: thanks, as it has been decided to move forward instead, marking this as resolved.

projects/extension/sql/incremental/020-migrate-existing-vectorizers-to-loading.sql
```diff
@@ -0,0 +1,39 @@
...
+            );
+            
+            -- Update the vectorizer with new config
+            UPDATE ai.vectorizer 
+            SET config = _config
+            WHERE id operator(pg_catalog.=) _vectorizer.id;
+        END IF;
+    END LOOP;
+end;
+$block$;
```
adolsalamanca, line 39:
```suggestion
$block$;

```

jgpruitt (changes requested):

projects/pgai/pyproject.toml
```diff
@@ -28,6 +28,10 @@ dependencies = [
     "litellm>=1.58.2,<1.59.0",
     "google-cloud-aiplatform>=1.78.0,<2.0",
     "boto3>=1.35.0,<2.0",
+    "smart-open>=7.1.0",
+    "pymupdf4llm>=0.0.17",
+    "filetype>=1.2.0",
+    "docling>=2.21.0",
```
JamesGuthrie, line 34: I think that we should constrain the versions here.

projects/pgai/pgai/cli.py
```diff
@@ -136,18 +136,21 @@ def run_vectorizer(
     db_url: str,
     vectorizer: Vectorizer,
     concurrency: int,
+    loading_retries: int,
```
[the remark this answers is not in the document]
adolsalamanca: fair point, I would then suggest to keep it at the level it's now, but renaming the field. thoughts?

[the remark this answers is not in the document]
smoya: I've thinking (and discussed it with @adolsalamanca) about the possibility of having a global "retries" setting but from the user point of view, I'm not very confident it would become a real-life use case but instead setting that with some granularity. Even though I like starting from simple things, I believe we have the chance of setting up this to be granular enough. 

Enabling setting the max retries for loading atm, plus allowing in the future to set up max retries for embedding generations (which I believe is crucial due to the fact money is involved in most cases) can be done if we split the retries logic per unit of work: loading, embedding, etc. 

After a discussion with @adolsalamanca, I suggested we came up with the following 2 approaches:

### 1. Use individual columns for each type of retry

- `loading_retries`
- `embedding_retries`
- etc

Plus one for each `retry_after` (e.g. `loading_retry_after`)

### 2. Use one single `jsonb`-type  column for all retries info

Storing a json such as the following:

```json
{
    "loading":
    {
        "retries": 3,
        "retry_after": 1741090438
    },
    "embedding":
    {
        "retries": 1,
        "retry_after": 1741090122
    }
}
```

Concerns on performance querying jsonb are there, but I believe we are good in our use cases?

projects/pgai/tests/vectorizer/cli/test_vectorizer_document.py
```diff
@@ -0,0 +1,421 @@
...
+
+def test_simple_document_embedding_s3(
+    cli_db: tuple[TestDatabase, Connection],
+    cli_db_url: str,
+    vcr_: Any,
+):
+    """Test that a document is successfully embedded"""
+    connection = cli_db[1]
+    vectorizer_id = configure_document_vectorizer(
+        cli_db[1], base_path="s3://adol-docs-test"
```
JamesGuthrie, line 175: Can we rename this bucket, so that it's more obvious that pgai's tests rely on this bucket existing? Or could we use something like minio instead, and not have a dependency on s3?

projects/extension/sql/idempotent/013-loading.sql
```diff
@@ -0,0 +1,83 @@
...
+    select count(*) operator(pg_catalog.>) 0 into strict _found
+    from pg_catalog.pg_class k
+        inner join pg_catalog.pg_namespace n on (k.relnamespace operator(pg_catalog.=) n.oid)
+        inner join pg_catalog.pg_attribute a on (k.oid operator(pg_catalog.=) a.attrelid)
+        inner join pg_catalog.pg_type y on (a.atttypid operator(pg_catalog.=) y.oid)
+    where n.nspname operator(pg_catalog.=) source_schema
+        and k.relname operator(pg_catalog.=) source_table
+        and a.attnum operator(pg_catalog.>) 0
+        and a.attname operator(pg_catalog.=) _column_name
+        and y.typname in ('text', 'varchar', 'char', 'bpchar', 'bytea');
```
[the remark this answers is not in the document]
smoya: Good catch. I believe that was already in that way but probably removed the proper test when rebasing. Now all in place.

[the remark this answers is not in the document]
smoya: Good point. Check added.

projects/extension/sql/idempotent/013-loading.sql
```diff
@@ -0,0 +1,87 @@
...
+    if _implementation is null or _implementation not in ('column', 'uri') then
+        raise exception 'invalid loading config implementation';
+end if;
+
+    _column_name = config operator(pg_catalog.->>) 'column_name';
+     if _column_name is null then
+        raise exception 'invalid loading config, missing column_name';
+end if;
+
+    select y.typname into _column_type
```
smoya, line 65: @jgpruitt As we need to get the column type for later do proper validations based on it, I changed this query to just get the column type. Now the count(*) > 0 check is not needed. I had to remove the `strict` because the select could potentially return more than 1 row in this case now, however I didn't add a `limit 1` because I believe this won't ever happen (to return more than 1 result) right? If you think the `limit 1` is needed as consistency check, I can add it.

Is there anything else I'm not considering? 🙏

projects/pgai/tests/vectorizer/cli/test_vectorizer_document.py
smoya: Good point! I will use localstack as @MetalBlueberry suggested 💯

Askir: We don't actually rely on this bucket for running the tests, since vcr has the request and response recorded. You can probably just rename it in the test and the recorded request yaml and it will still work.

smoya: But you won't be able to recreate the VCR if we remove such a bucket. 
I ended up adding localstack via testcontainers as we did for other cases.

projects/extension/tests/vectorizer/test_parsing.py
jgpruitt: i don't feel strongly either way

projects/extension/sql/idempotent/013-loading.sql
smoya: Then it is fine to leave it as it is now with my change.

projects/extension/tests/vectorizer/test_vectorizer.py
adolsalamanca: I thought it used to fail with the server running too, but that's correct. 
removing this statement now.
thanks!

projects/extension/sql/idempotent/013-loading.sql
```diff
@@ -0,0 +1,96 @@
...
+    if _implementation is null or _implementation not in ('column', 'uri') then
+        raise exception 'invalid loading config implementation';
+end if;
+
+    _column_name = config operator(pg_catalog.->>) 'column_name';
+     if _column_name is null then
+        raise exception 'invalid loading config, missing column_name';
+end if;
+    
+    if (config operator(pg_catalog.->>) 'retries')::int < 0 then
```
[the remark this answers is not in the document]
adolsalamanca: oh that's a good point, thanks!

projects/extension/sql/idempotent/013-loading.sql
adolsalamanca: done, thanks!

projects/extension/sql/idempotent/013-loading.sql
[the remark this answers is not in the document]
adolsalamanca: done!

projects/extension/tests/vectorizer/server.py
```diff
@@ -110,6 +108,38 @@ class ProcessingDefault(BaseModel):
...
 
 
+class LoadingColumn(BaseModel):
+    implementation: Literal["column"]
+    config_type: Literal["loading"]
+    column_name: str
+
+
+class LoadingUri(BaseModel):
+    implementation: Literal["document"]
```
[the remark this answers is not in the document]
smoya: fixed

projects/extension/sql/idempotent/012-vectorizer-int.sql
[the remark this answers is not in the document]
adolsalamanca: done!

projects/extension/tests/vectorizer/server.py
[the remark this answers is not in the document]
adolsalamanca: fixed, thanks!

projects/extension/sql/incremental/020-migrate-existing-vectorizers-to-loading.sql
```diff
@@ -0,0 +1,41 @@
...
+           ),
+                'parsing', json_object(
+                    'implementation': 'auto',
+                    'config_type': 'parsing'
+                ),
+                'chunking', _chunking operator(pg_catalog.-) 'chunk_column'
+            );
+            
+            -- Update the vectorizer with new config
+            UPDATE ai.vectorizer
```
jgpruitt, line 35: We should update the `version` in the config to be the version of this release.

jgpruitt (changes requested):

projects/extension/tests/vectorizer/server.py
jgpruitt: when running the extension's tests, we run this test server. It allows us to calls `ai.execute_vectorizer` which sends an HTTP request with the vectorizer as a json payload. This test server then deletes everything from that vectorizer's queue.
The test server also facilitates tests where we fetch secrets.

jgpruitt: Here is an example where we create a vectorizer, check that the queue has items, simulate running the vectorizer (via this test service), and then check that the queue is empty.
https://github.com/timescale/pgai/blob/main/projects/extension/tests/vectorizer/test_vectorizer.py#L323

jgpruitt: the library depends on the extension. I didn't want extension tests to also depend on the library. that seemed bad, so I added this test server

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -354,6 +362,49 @@ def is_vectorizer_disabled_query(self) -> sql.Composed:
...
+                    {queue_table}
+                WHERE
+                    ({pk_fields}) IN ({pk_values})
+                AND
+                    (loading_retry_after is null or loading_retry_after < now())
+            ) AS source
+            ON {merge_predicates}
+            WHEN MATCHED
+                AND target.loading_retries >= %(loading_retries)s THEN
+                    DELETE
```
cevian, line 382: This makes me unhappy as this leaves no record of what permanently failed.
I see a few options
- add a tombstone column
- add a tombstone table
- just increment loading_retries again and have the "get stuff from work queue" logic ignore items that have to high a retry count.

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -504,12 +556,38 @@ async def run(self) -> int:
                             ),
                         ),
                     )
+
+                if e.__cause__ is not None:
+                    raise e.__cause__  # noqa
+                raise e
+
+            except UriLoadingError as e:
```
cevian, line 564: hmm handling this error at this level fails the rest of the rows in the batch too? what if 99 items succeeded and 1 failed? This seems not very good. why don't we handle these like we do in `_generate_embeddings`. specifically we should handle it like `ChunkEmbeddingError` and return but not raise.

cevian (changes requested):
Overall, good work.

My biggest concern is the worker retry handling. See comments

projects/extension/sql/incremental/020-migrate-existing-vectorizers-to-loading.sql
adolsalamanca: For the record, happy to apply this change if that's what we want, but currently tests are failing due to the previously mentioned script overwriting our intention:

image

projects/extension/sql/idempotent/012-loading.sql
```diff
@@ -0,0 +1,96 @@
...
+    )
+$func$ language sql immutable security invoker
+set search_path to pg_catalog, pg_temp
+;
+
+-------------------------------------------------------------------------------
+-- loading_uri
+create or replace function ai.loading_uri
+( column_name pg_catalog.name
+, retries pg_catalog.int4 default 6)
```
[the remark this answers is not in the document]
adolsalamanca: That's correct, given a previous discussion we have decided to limit where it applies now.
It was a generic retries field before, but only acting on loading, that's why we decided to iterate and make its naming clearly attached to the loading, given it's where it's acting as of today.

projects/extension/sql/idempotent/014-vectorizer-int.sql
```diff
@@ -399,7 +399,14 @@ declare
 begin
     -- create the table
     select pg_catalog.format
-    ( $sql$create table %I.%I(%s, queued_at timestamptz not null default now())$sql$
+    ( $sql$
+      create table %I.%I
+      ( %s
+      , queued_at pg_catalog.timestamptz not null default now()
+      , loading_retries pg_catalog.int4 not null default 0
```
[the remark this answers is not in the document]
adolsalamanca: correct, only applies to loading as of today.

projects/pgai/pgai/vectorizer/vectorizer.py
adolsalamanca: fixed, coming in a separated PR.

projects/pgai/pgai/vectorizer/migrations/v0_10_0.py
```diff
@@ -0,0 +1,91 @@
...
+)
+def migrate_to_0_10_0(old_conf: Config_0_9) -> dict[str, Any]:
+    # use the data as is from the previous version so we modify whatever is needed
+    result = old_conf.__dict__.copy()
+
+    result["loading"] = ColumnLoading(
+        implementation="column", column_name=old_conf.chunking.chunk_column
+    )
+    result["parsing"] = ParsingNone(implementation="none")
+    match old_conf.chunking.implementation:
```
cevian, line 73: lets throw an exception if an unexpected implementation found

projects/pgai/pyproject.toml
cevian: ping on this

projects/pgai/tests/vectorizer/cli/test_vectorizer_core.py
```diff
@@ -487,3 +492,59 @@ def test_regression_vectorizer_uuid_pk_with_error(
...
+    the loading_retries feature enabled"""
+    _, connection = cli_db
+    table_name = setup_source_table(connection, 1)
+
+    # Given a vectorizer and the feature `loading_retries` disabled.
+    vectorizer_id = configure_vectorizer(
+        table_name,
+        cli_db[1],
+    )
+    features = Features("0.9.0")
```
cevian, line 512: note: this will have to change with the next rebase

projects/pgai/tests/vectorizer/cli/test_vectorizer_document.py
```diff
@@ -0,0 +1,476 @@
...
+    vectorizer_id = configure_document_vectorizer(
+        cli_db[1], base_path=f"s3://{s3_bucket}"
+    )
+
+    # required for accessing this S3 bucket in particular
+    # os.environ["AWS_DEFAULT_REGION"] = "eu-central-1"
+
+    # This is necessary to prevent boto3 from pulling credentials
+    # from WS Instance Metadata Service when those are missing.
+    # Comment if you need to recreate the cassette.
```
cevian, line 234: can we detect whether we are recreating the cassete in code and then programmatically set the right things>

projects/pgai/tests/vectorizer/cli/test_vectorizer_document.py
```diff
@@ -0,0 +1,476 @@
...
+            " key: 'non_existing_doc.pdf' version: None"
+            " error: An error occurred (NoSuchKey) when"
+            " calling the GetObject operation: The specified"
+            " key does not exist.",
+            "is_retryable": False,
+        }
+
+        cur.execute(
+            "SELECT loading_retries, loading_retry_after" " FROM ai._vectorizer_q_1;"
+        )
```
cevian, line 408: what does this do? the result is never checked

cevian (changes requested):
Full review. Mainly small changes requested

docs/vectorizer/api-reference.md
```diff
@@ -145,6 +149,239 @@ in other management functions.
...
+| None | - | - | - | - |
+
+#### Returns
+
+A JSON configuration object that you can use in [ai.create_vectorizer](#create-vectorizers).
+
+
+### ai.parsing_none
+
+You use `ai.parsing_none` to load the data as-is from the source table.
```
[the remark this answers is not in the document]
adolsalamanca: done, thanks

projects/pgai/pgai/vectorizer/migrations/v0_10_0.py
adolsalamanca: fixed!

projects/pgai/tests/vectorizer/cli/test_vectorizer_core.py
adolsalamanca: fixed, thanks

projects/extension/sql/idempotent/014-vectorizer-int.sql
```diff
@@ -399,7 +399,14 @@ declare
...
+      , loading_retry_after pg_catalog.timestamptz
```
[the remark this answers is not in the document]
adolsalamanca: the alternative we were considering was having a jsonb to have connector/process specific tracking info, 
but given that this came up in a later stage of the process, we decided to keep it simple for now.

how about the moment we decide to extend the loading capabilities we fallback into that option,
keeping things as they're for now here? happy to discuss about it otherwise @cevian

projects/extension/sql/idempotent/012-loading.sql
```diff
@@ -0,0 +1,96 @@
+-------------------------------------------------------------------------------
+-- loading_column
+create or replace function ai.loading_column
+( column_name pg_catalog.name
+, retries pg_catalog.int4 default 6)
```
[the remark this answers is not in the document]
adolsalamanca: To be honest I believe it started as a consistency addition.
However, we have just experienced issues loading data from compressed tables.
could that be one of the cases to make this valuable? 
given that we might prefer to eventually drop on the retries to avoid
bleeding us or the customer on AWS lambda costs?

thoughts?

projects/extension/ai/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "0.9.1-dev"
+__version__ = "0.10.0-dev"
```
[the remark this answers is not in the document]
adolsalamanca: I'm not sure to be honest, not sure if Sergio had to do this for some reason.
happy to revert it in case it's required.

examples/embeddings_from_documents/README.md
```diff
@@ -0,0 +1,144 @@
+# Tutorial: Generating document embeddings with PGAI
+
+This tutorial will guide you through the process of generating embeddings for documents using the PGAI extension for PostgreSQL.
+
+## Prerequisites
+
+- Download this example subdirectory. You can quickly do it by generating a downloadable `.zip` file from [here](https://download-directory.github.io/?url=https%3A%2F%2Fgithub.com%2Ftimescale%2Fpgai%2Ftree%2Fmain%2Fexamples%2Fembeddings_from_documents).
+- PostgreSQL database with the PGAI extension installed. Refer to [pgai install](/docs/README.md#pgai-install) for installation instructions.
+- Documents to process (supports various formats including MD, XLSX, HTML, PDF). We will use those available in the [documents](documents) directory.
+- A running instance of the [Vectorizer Worker](/docs/vectorizer/worker.md). In order to load the documents from the [documents](documents) directory, you need to modify the `compose.yaml` file created in the previous step and add the following volume to the `vectorizer-worker` service:
```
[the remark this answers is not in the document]
adolsalamanca: I guess it might be a copy paste issue. I'll change to point to projects/pgai/compose-dev.yaml

projects/pgai/tests/vectorizer/cli/test_vectorizer_document.py
adolsalamanca: fixed, thanks!

projects/extension/sql/idempotent/005-chunking.sql
adolsalamanca: we decided not to.

projects/extension/pyproject.toml
```diff
@@ -19,6 +19,7 @@ dependencies= [
     "litellm==1.58.2",
     "google-cloud-aiplatform==1.74.0", # required for vertexAI (don't know why litellm doesn't include this)
     "langchain-text-splitters>=0.3.5",
+    "jinja2>=3.1.6",
```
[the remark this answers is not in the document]
adolsalamanca: seems like it was already in the requirements-lock, so it was possibly just added to the pyproject, but was already a dependency.

```
uv pip show jinja2
Using Python 3.12.0 environment at: projects/extension/.venv
Name: jinja2
Version: 3.1.6
Location: /home/adol/Work/pgai/projects/extension/.venv/lib/python3.12/site-packages
Requires: markupsafe
Required-by: litellm, pga
```

projects/extension/sql/incremental/020-migrate-existing-vectorizers-to-loading.sql
adolsalamanca: ping @cevian @jgpruitt, as this one I wasn't entirely sure about and might be relevant.

docs/vectorizer/api-reference.md
```diff
@@ -145,6 +149,239 @@ in other management functions.
...
+A JSON configuration object that you can use in [ai.create_vectorizer](#create-vectorizers).
+
+## Parsing configuration
+
+You use the parsing configuration functions in `pgai` to define how data is parsed after document loading. This is useful if for non-textual formats such as PDF documents.
+
+The parsing functions are:
+
+- [ai.parsing_auto](#aiparsing_auto): Automatically selects the appropriate parser based on file type.
+- [ai.parsing_none](#aiparsing_none): Converts various formats to Markdown. See [PyMuPDF](https://pymupdf.readthedocs.io/en/latest/) for supported formats.
```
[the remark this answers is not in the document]
adolsalamanca: fixed, thanks

projects/pgai/pgai/vectorizer/vectorizer.py
adolsalamanca: done

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -117,8 +140,9 @@ class Vectorizer:
 
     id: int
     config: Config
-    queue_table: str
     queue_schema: str
+    queue_table: str
+    queue_failed_table: str
```
Askir, line 145: This will crash on older pgai extension versions where this field doesn't exist yet.
I'm running into this right now trying to write a test with pgai 0.8.0 on this vectorizer version:

```
{'type': 'missing', 'loc': ('queue_failed_table',), 'msg': 'Field required', 'input': {'id': 1, 'config': {'version': '0.8.0', 'chunking': {'separator': '\n\n', 'chunk_size': 800, 'config_type': 'chunking', 'chunk_column': 'content', 'chunk_overlap': 400, 'implementation': 'character_text_splitter', 'is_separator_regex': False}, 'indexing': {'config_type': 'indexing', 'implementation': 'none'}, 'embedding': {'model': 'text-embedding-ada-002', 'dimensions': 1536, 'config_type': 'embedding', 'api_key_name': 'OPENAI_API_KEY', 'implementation': 'openai'}, 'formatting': {'template': '$chunk', 'config_type': 'formatting', 'implementation': 'python_template'}, 'processing': {'config_type': 'processing', 'implementation': 'default'}, 'scheduling': {'config_type': 'scheduling', 'implementation': 'none'}, 'destination': {'implementation': 'default', 'config_type': 'destination', 'target_schema': 'public', 'target_table': 'blog_embedding_store'}}, 'disabled': False, 'source_pk': [{'pknum': 1, 'attnum': 1, 'attname': 'id', 'typname': 'int4'}], 'view_name': 'blog_embedding', 'queue_table': '_vectorizer_q_1', 'view_schema': 'public', 'queue_schema': 'ai', 'source_table': 'blog', 'target_table': 'blog_embedding_store', 'trigger_name': '_vectorizer_src_trg_1', 'source_schema': 'public', 'target_schema': 'public'}, 'url': 'https://errors.pydantic.dev/2.10/v/missing'}
```

Askir (changes requested):
This vectorizer is not backwards compatible, blocking to make sure we discuss this.
Generally I'm not too happy with our "migration" approach here now.

projects/extension/sql/idempotent/014-vectorizer-int.sql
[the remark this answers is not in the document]
cevian: My only actual suggestion was to rename  `loading_retry_after` to `retry_after` but I am fine leaving as-is for now.

projects/extension/pyproject.toml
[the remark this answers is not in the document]
adolsalamanca: the version bump was to address a vulnerability btw

docs/vectorizer/api-reference.md
```diff
@@ -145,6 +149,228 @@ in other management functions.
...
+A JSON configuration object that you can use in [ai.create_vectorizer](#create-vectorizers).
+
+## Parsing configuration
+
+You use the parsing configuration functions in `pgai` to define how data is parsed after document loading. This is useful if for non-textual formats such as PDF documents.
+
+The parsing functions are:
+
+- [ai.parsing_auto](#aiparsing_auto): Automatically selects the appropriate parser based on file type.
+- [ai.parsing_none](#aiparsing_none): Converts various formats to Markdown.
```
alejandrodnm, line 242: I think this is a left over mistake. Below we say that this option `skip the parsing step`

projects/extension/sql/idempotent/012-loading.sql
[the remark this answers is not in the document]
alejandrodnm: The use case I can imagine is having a value that generates chunks that are bigger than what's allowed by the model, those will never succeed and will just stay in the queue forever. I haven't look into how this is used, but I guess it's to prevent embedding them if they keep failing, which would help in this case.

Askir (approved):
lgtm

projects/pgai/tests/test_model_download.py
```diff
@@ -0,0 +1,63 @@
...
+        assert result.exit_code == 0
+
+        # Verify that download_models was called with the correct parameters
+        mock_download.assert_called_once_with(
+            progress=True,
+            output_dir=DOCLING_CACHE_DIR,
+        )
+
+
+@pytest.mark.skip("Integration test")
```
JamesGuthrie, line 30: why is this skipped?

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -365,6 +389,102 @@ def fetch_work_query(self) -> sql.Composed:
...
+                ),
+                deleted_rows AS (
+                    DELETE FROM {queue_table} AS w
+                    USING locked_items AS l
+                    WHERE l.locked = true
+                    AND {delete_join_predicates}
+                )
+                SELECT {source_table}.*, {loading_retries}
+                FROM locked_items
+                LEFT JOIN {source_schema}.{source_table} USING ({pk_fields})
```
JamesGuthrie, line 456: Incorporate this: https://github.com/timescale/pgai/pull/581

projects/pgai/tests/test_model_download.py
Askir: Because that downloads gigabytes of data and runs for minutes. I didn't want to block CI for that long all the time. There is a mocked version above that runs always.

JamesGuthrie: 👍

PR #443, opened 2025-02-06
docs: pgai extension install requires sudo

JamesGuthrie:
The pgai extension install needs to write to system-owned paths, including `/usr/local/lib`, and the postgres extension path, typically `/usr/share/postgresql/<pg version>/extension`.

### DIFF
--- a/docs/install_from_source.md
+++ b/docs/install_from_source.md
@@ -36,20 +36,29 @@ To install pgai from source on a PostgreSQL server:
 1. Clone the pgai repo at the latest tagged release:
 
     ```bash
-    git clone https://github.com/timescale/pgai.git --branch extension-0.6.0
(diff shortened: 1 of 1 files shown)


docs/install_from_source.md
```diff
@@ -36,20 +36,25 @@ To install pgai from source on a PostgreSQL server:
...
     ```
 
 1. Install the `pgai` PostgreSQL extension:
 
     ```bash
-    just ext install
+    sudo just ext install
     ```
+    
+    Note: The install requires sudo to create the following files (in system-owned paths):
```
alejandrodnm, line 49: Should we add to the note explicitly mentioning that if you don't want to use sudo the user that's running the command needs write access to these folders?

alejandrodnm (approved):

PR #444, opened 2025-02-06
ci: add action for linting PR title instead of all commits

smoya:
# Description

Since our merge strategy is set to always squash, commits are squashed into one single commit, getting the title by default from the PR title. Then there is no need to lint every single commit in the PR and rather just the title.
This avoids the annoying moment when you just add a commit into the PR such as "fixing test a and b" and the CI complains because such a commit doesn't follow conventional commits.


adolsalamanca (approved):

alejandrodnm (changes requested):
We actually have 2 options, squash and rebase (I used it by default), if you merge this then people that use rebase might cause problems.
If you want to do this, then you’d have to remove rebase as an option.
I’m fine either way, I usually only have one commit per PR.

smoya:
> If you want to do this, then you’d have to remove rebase as an option.

I'm happy setting this up for only allowing "squash and merge" strategy.

PR #448, opened 2025-02-06
feat: vectorizer row loader, default chunking

Askir:
- Removes the chunk_column from chunking configs
- Makes recursive_chunking the default
- Adds required loading config with `loading_row(column_name -> 'column')`

Docs are missing. I'd do those in a separate PR because it will communicate a different API and merging that before a release is bound to cause confusion.


projects/extension/tests/vectorizer/test_vectorizer.py
```diff
@@ -260,37 +266,6 @@ def test_vectorizer_timescaledb():
             actual = cur.fetchone()[0]
             assert actual == 3
 
-            # bob should have select on the source table
```
cevian, line 263: why were these deleted?

projects/pgai/pgai/vectorizer/configuration.py
```diff
@@ -62,13 +61,25 @@ class ChunkingRecursiveCharacterTextSplitterConfig(SQLArgumentMixin):
...
     function_name: ClassVar[str] = "ai.chunking_recursive_character_text_splitter"
 
-    chunk_column: str
     chunk_size: int | None = None
     chunk_overlap: int | None = None
     separators: list[str] | None = None
     is_separator_regex: bool | None = None
 
 
+@dataclass
```
cevian, line 70: how was this missing before and things worked?

cevian (changes requested):
Mostly looks good. 2 questions + I think there need to be doc changes

projects/pgai/pgai/vectorizer/configuration.py
Askir: I think litellm via alembic just isn't possible right now., there is no test for it either.
We only merged litellm recently, and I think forgot this step.
Tbf both litellm and alembic were in development at the same time, so it wasn't 100% clear who would merge first and the other one would have to update this. I think I mentioned this to james, but with so many things to keep track of it can be forgotten anyway.

projects/extension/tests/vectorizer/test_vectorizer.py
Askir: This was done in the commits I based my changes onto and explained here:
https://github.com/timescale/pgai/pull/407#discussion_r1942965198

Apparently those assertions are done twice right now.

projects/extension/sql/incremental/016-migrate-existing-vectorizers-to-loading.sql
```diff
@@ -0,0 +1,34 @@
...
+    _vectorizer RECORD;
+    _chunking jsonb;
+    _chunk_column text;
+    _config jsonb;
+BEGIN
+    -- Loop through all vectorizers
+    FOR _vectorizer IN SELECT id, config FROM ai.vectorizer
+    LOOP
+        -- Extract the chunking config and chunk_column
+        _chunking := _vectorizer.config->'chunking';
```
smoya, line 12:
```suggestion
        _chunking := _vectorizer.config operator(pg_catalog.->) 'chunking'; 
```

(applies for the rest of the file)

Based in our codebase, we use operators instead. However, I'm not 100% sure it is mandatory; I agree this implementation looks more readable than with operators. Ping @jgpruitt who can shed some light on this.

projects/pgai/pgai/vectorizer/generate/docker-compose.yaml
```diff
@@ -1,7 +1,10 @@
 name: pgai
 services:
   db:
-    image: timescale/timescaledb-ha:pg17
+    build:
+      context: ../../../../extension
+      dockerfile: Dockerfile
+      target: pgai-test-db
```
smoya, line 7: 💯 I was always doing this change locally but not committing 😅

smoya: WDYT about creating a separate PR with this change targeting `main` branch so other devs can benefit from it asap.

smoya (approved):
LGTM. Let's move forward and merge this PR into the integration branch.

projects/extension/tests/vectorizer/test_vectorizer.py
adolsalamanca: yes, those tests are now redundant.

projects/extension/sql/incremental/016-migrate-existing-vectorizers-to-loading.sql
jgpruitt: `->` is considerably more readable.
`operator(pg_catalog.->)` is "safer". It explicitly refers to the `->` operator that exists in the `pg_catalog` schema which is built-in. Only a superuser could change the definition of `->` in `pg_catalog`.
`->` alone resolves the operator based on the current search_path. To execute malicious code, a malicious user could define a new `->` operator in a schema other than `pg_catalog`. This only really comes in to play if the function is `security definer`. A `security definer` function defined in an extension is executed as superuser.

PR #461, opened 2025-02-10
feat: add parsing and implement load document

Askir:
I tried writing a test with a png, but I realized that:
1. filetype doesn't recognize pngs
2. Even if I hardcode filetype "png", pymupdf actually has no OCR capabilities out of the box you have to separately enable that:
https://pymupdf.readthedocs.io/en/latest/installation.html#installation-ocr

We could still add a test for:
- mobi/epub
- file from `bytea` column
though.

Other than that, only docs are missing I believe.


projects/pgai/pgai/vectorizer/parsing.py
```diff
@@ -0,0 +1,51 @@
...
+                "use parsing_auto or parsing_none instead"
+            )
+        file_like = io.BytesIO(payload)
+        kind = filetype.guess(file_like)  # type: ignore
+
+        if kind is None:
+            raise ValueError("Could not determine file type")
+
+        file_like.seek(0)  # Reset buffer position after guessing file type
+        with pymupdf.open(stream=file_like, filetype="pdf") as pdf_document:
```
smoya, line 48: are we assuming it's always pdf? 🤔

PR #462, opened 2025-02-10
feat(text_to_sql): add explain to validate generated query

jgpruitt:
WIP: experimenting with using `EXPLAIN` to check that the query generated by the LLM is valid.


jgpruitt:
see https://github.com/timescale/pgai/pull/503 instead

PR #466, opened 2025-02-11
feat: support queue item retries 

adolsalamanca:
As agreed in the design of the s3 document support, we're including retries and retry_after fields in the queue tables generated for each vectorizer.

Also add a couple of minor changes, rename func used in upgrades, and of vars of the loading migration.


smoya (approved):
LGTM! 🚀🌔

Next step is actually implementing the backoff, right?

PR #467, opened 2025-02-11
feat: add required changes for supporting multiple document file types

smoya:
This PR adds the base code required for supporting parsing other document types rather than pdf.


projects/pgai/pgai/vectorizer/parsing.py
```diff
@@ -22,30 +22,47 @@ def parse(self, row: dict[str, Any], payload: str | bytes) -> str:  # noqa: ARG0
...
+        self, row: dict[str, Any], payload: bytes | str, column_name: str | None
+    ) -> str:  # noqa: ARG002
         if isinstance(payload, str):
             raise ValueError(
                 "Column content must be bytes to be parsed by pymupdf "
                 "use parsing_auto or parsing_none instead"
             )
-        file_like = io.BytesIO(payload)
-        kind = filetype.guess(file_like)  # type: ignore
+        with io.BytesIO() as file_like:
```
smoya, line 47: Ensuring we close the file when the leaving the context

projects/pgai/pgai/vectorizer/parsing.py
```diff
@@ -22,30 +22,47 @@ def parse(self, row: dict[str, Any], payload: str | bytes) -> str:  # noqa: ARG0
...
+            file_like.write(memoryview(payload))
```
smoya, line 48: This way, we are not really duplicating the content but reusing the payload

PR #469, opened 2025-02-11
test: add epub and markdown tests + byta column

Askir:
I am not sure if doing the filetype guessing in the loader is cleaner or not, but by having the dataclass to pass around we avoid having to inject the config into the parser at least, which I think is definitely better.


projects/pgai/pgai/vectorizer/parsing.py
```diff
@@ -22,47 +22,34 @@ def parse(self, _1: dict[str, Any], payload: str | bytes, _2: str | None) -> str
...
-        self, row: dict[str, Any], payload: bytes | str, column_name: str | None
-    ) -> str:  # noqa: ARG002
+    def parse(self, row: dict[str, Any], payload: LoadedDocument | str) -> str:  # noqa: ARG002
         if isinstance(payload, str):
             raise ValueError(
-                "Column content must be bytes to be parsed by pymupdf "
+                "Column content must be a document to be parsed by pymupdf "
                 "use parsing_auto or parsing_none instead"
             )
-        with io.BytesIO() as file_like:
```
Askir, line 47: You don't need to close bytesIO objects they are completely contained in memory and simply cleaned up by the garbage collector. The reason other IO objects have to be closed is because they make use of OS resources like file descriptors which are not freed by the GC itself, but BytesIO just "pretends" to be a file so that the interface is consistent.

smoya: Using the context manager here was to immediately free the in-memory usage, considering the amount of memory we could use in this exact function. It does not matter if it requires i/o from system or not, it is a matter of releasing the memory as fast as we can; common when handling with files and easy to maintain with contexts.

However, It made sense in a prior version (mine) of the code, where there was more code right after the pymupdf parsing. With the current code, there is no benefit on ensuring the `file_like` gets freed because we are actually referring to that var until the very end of the function.

projects/pgai/pgai/vectorizer/parsing.py
```diff
@@ -22,47 +22,34 @@ def parse(self, _1: dict[str, Any], payload: str | bytes, _2: str | None) -> str
...
 
-            if kind:
-                extension = kind.extension
-            elif column_name is not None:
-                # determine extension if possible
-                extension = row[column_name].split(".")[-1]
-            else:
-                raise ValueError("No file extension could be determined")
+        if payload.file_type is None:
+            raise ValueError("No file extension could be determined")
```
smoya, line 43:
```suggestion
            raise ValueError("No file type could be determined")
```

smoya (approved):
LGTM! 🚀🌔

PR #470, opened 2025-02-11
fix: record exceptions in embedding.setup()

cevian:
Previously exceptions in self.vectorizer.config.embedding.setup() were not recorded in the db.

This PR catches exceptions one stack higher to catch more of them.


jgpruitt (approved):

smoya (approved):
LGTM! 🚀🌔

PR #474, opened 2025-02-12
feat(text_to_sql): add a func to resolve ambiguous object references

jgpruitt:
Adds a func to resolve ambiguous object references.

We need the user's search_path to make sampling table rows and running EXPLAIN work when the database objects are not fully-qualified. So, we capture it and pass it down.

### DIFF
--- a/projects/extension/ai/semantic_catalog.py
+++ b/projects/extension/ai/semantic_catalog.py
@@ -8,16 +8,15 @@ class GeneratedDescription(TypedDict):
     description: str
 
 
-def render_sample(plpy, relation: str, total: int = 5) -> str:
+def render_sample(plpy, fq_relation: str, total: int = 5) -> str:
     """
     Given a table ore view name, return a sample of rows from it.
     """
-    ident = ".".join([plpy.quote_ident(part) for part in relation.split(".")])
     # do not let LLM select 0 or less and no more than 10 rows, so as to not overwhelm
     # context window.
-    result = plpy.execute(f"select * from {ident}", min(max(total, 1), 10))
+    result = plpy.execute(f"select * from {fq_relation}", min(max(total, 1), 10))
 
-    ret_obj = f"""<data id="{relation}">"""
+    ret_obj = f"""<data id="{fq_relation}">"""
     for r in result:
         values = []
         for v in r.values():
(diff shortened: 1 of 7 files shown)


projects/extension/sql/idempotent/906-text-to-sql-anthropic.sql
```diff
@@ -409,7 +410,7 @@ begin
                             ;
                         when 'request_table_sample' then
                             raise debug 'tool use: request_table_sample: %', _message.input;
-                            select _samples || jsonb_build_object(_message.input->>'name', ai.render_sample((_message.input->>'name')::regclass, (_message.input->>'total')::int4))
+                            select _samples || jsonb_build_object(_message.input->>'name', ai.render_sample((_message.input->>'name')::regclass, (_message.input->>'total')::int4, search_path))
```
MasterOdin, line 413: I don't think this will work how we want it to? I believe that this should error if `(_message.input->>'name')::regclass` isn't fully qualified as it'll use the current function search path, and so we won't even enter `ai.render_sample`.

I think we'd either need to wrap calling the function with the `set search path` stuff, or maybe add a second variant of the function that's defined as:

```sql
create or replace function ai.render_sample(name text, total int4, search_path string) returns text
as $func$
    -- call set local search_path
    -- call and return select ai.render_sample(name::regproc, total)
$func$ language sql stable security invoker
;
```

projects/extension/ai/semantic_catalog.py
```diff
@@ -8,10 +8,18 @@ class GeneratedDescription(TypedDict):
...
 
 
-def render_sample(plpy, relation: str, total: int = 5) -> str:
+def render_sample(
+    plpy, relation: str, total: int = 5, search_path: str = "pg_catalog, pg_temp"
+) -> str:
     """
     Given a table ore view name, return a sample of rows from it.
     """
+    # set the search path. this won't work for schemas that contain a comma. please don't do that
```
cevian, line 17: do we really want this? don't we want the rendering to schema-qualify everything it possibly can?

jgpruitt: Yes. We will render everything fully qualified. If the LLM asks for sample data but fails to provide a fully qualified reference, this code attempts to be able to find the table anyway by using the user's search path. We still render it fully qualified

projects/extension/sql/idempotent/906-text-to-sql-anthropic.sql
jgpruitt: Oh yeah. I think you're right. Lemme think about that.

projects/extension/ai/semantic_catalog.py
jgpruitt: looking at this with fresh eyes. I think I understand what you mean. I'm gonna take another stab at this

projects/extension/sql/idempotent/906-text-to-sql-anthropic.sql
jgpruitt: fixed

projects/extension/ai/semantic_catalog.py
jgpruitt: fixed

cevian (approved):

PR #475, opened 2025-02-12
feat: add verbose flag to model calls

cevian:


projects/extension/sql/incremental/016-add-verbose-args.sql
```diff
@@ -0,0 +1,43 @@
+-- add verbose boolean to all model calling functions
+
+-- drop all functions that have verbose args before the arg was added.
+-- list generated by the following query:
+-- SELECT 'DROP FUNCTION IF EXISTS ' || ns.nspname || '.' || proname || '(' || 
+--   array_to_string(proargtypes[0:array_length(proargtypes, 1)-2]::regtype[], ', ') || ');' as drop_command 
+-- FROM pg_proc p 
+-- JOIN pg_namespace ns ON p.pronamespace = ns.oid 
+-- WHERE proargnames @> ARRAY['verbose'];
```
jgpruitt, line 9: love it!

jgpruitt (approved):
LGTM

projects/extension/ai/utils.py
```diff
@@ -7,3 +10,21 @@ def get_guc_value(plpy, setting: str, default: str) -> str:
     if val is None:
         val = default
     return val
+
+
+class VerboseTrace:
```
smoya, line 15: If only used for requests, I would suggest we rename the class.
```suggestion
class VerboseRequestTrace:
```

Otherwise, change the message in

```
self.plpy.info(f"sending a request for {self.name}...")
```

smoya (approved):
Only a minor, LGTM otherwise so I'm approving anyway

PR #476, opened 2025-02-12
test(openai): do not check content length when stopped for max tokens

jgpruitt:
The test_openai_chat_complete_with_tokens_limitation_on_reasoning_models test is failing randomly. Sometimes the LLM returns content when it is stopped for length reasons. Sometimes it doesn't. Drop the assertion.


Askir (approved):
If this fixes it, feel free to merge for now.
I don't really understand the test, the max_tokens is set to 1/100 why are we asserting that there is content length 0? shouldn't it at least be 1 token?

PR #477, opened 2025-02-13
feat: docling parser

smoya:
Docling requires some models for analysing and extracting data from documents. Those models are automatically downloaded the first time you run the parser if those are not found locally. However, we can prefetch them. See [docs](https://ds4sd.github.io/docling/usage/#model-prefetching-and-offline-usage). 

Should we do prefetch?:
- YES:
  - If we do it in the Docker image, the image size will increase by around ~300 MB.
  - Alternatively, we can download the calls in the conftest.py calling `docling.utils.model_downloader.download_models()`
- NO:
  - The models are downloaded on demand
  - However, tests will try to catch the response (the actual models) in the cassettes. We should then add hugging face host to the vcrpy **`ignore_hosts` **for this particular test**.

I decided to do prefetching in conftest.py, for tests. Users will download the models under demand but just once. 

Need to check if it's possible to cache those models in GH runner (we don't run self-hosted runners, so no idea yet if possible)


projects/pgai/tests/vectorizer/cli/test_vectorizer_document.py
```diff
@@ -18,6 +18,7 @@
     "sample_txt.txt",
     "basic-v3plus2.epub",
     "test.md",
+    "NVDA.pdf",
```
Askir, line 21: I am not sure we can just add this file under our main license?. Sure it's a publicly readable google doc with public data in it and unlikely someone complains but I didn't create this so I'm a bit careful about that 😅

smoya: Yeah don't worry, this pr won't be merged with that spreadsheet but instead will create a custom one.
Thanks for pointing that out.

smoya: Replaced by `lego_sets.pdf`, a PDF based from a free dataset from https://mavenanalytics.io/data-playground.
I imported the CSV in Google Spreadsheets, added an image (to test OCR on/off) and a title, and limited the dataset to few Star Wars lego sets to reduce the size.

smoya:
Tests are failing because we will need to ignore HF host for all tests and not only for those.

projects/pgai/pgai/vectorizer/parsing.py
```diff
@@ -24,32 +36,91 @@ class ParsingAuto(BaseModel):
...
-            # Convert to markdown using pymupdf4llm
-            md_text = pymupdf4llm.to_markdown(pdf_document)  # type: ignore
+            return pymupdf4llm.to_markdown(pdf_document)  # type: ignore
+
+
+class ParsingDocling(BaseDocumentParsing):
+    """Document parsing implementation using Docling."""
+
+    implementation: Literal["docling"]  # type: ignore[reportIncompatibleVariableOverride]
+    cache_dir: Path | str = Path.home().joinpath(".cache/docling/models")
```
smoya, line 106: In the future we could make this configurable via env var. Just using the default value from docling atm.

adolsalamanca (approved):
LGTM

PR #479, opened 2025-02-13
ci(pgai): add format-fix, lint-fix and fix cmds to justfiles

smoya:
- `just format-fix` will directly fix the formatting issues instead of showing the diff as `just format` does. Useful for development.
- `just lint-fix` will do the same but for linting issues.
- `just fix` will run both commands above.

This PR also adds pgspot as dev dependency instead of building from source in Docker. Unless there is a specific reason, this alternative allows running pgspot from host machine and not only from the Docker container.


adolsalamanca (approved):

cevian (approved):
Maybe also add a fix command that does lint-fix and format-fix?

projects/extension/Dockerfile
```diff
@@ -65,12 +65,6 @@ USER root
 
 RUN pip install --break-system-packages uv==0.5.20
 
-# install pgspot
-RUN set -eux; \
```
smoya, line 69: @jgpruitt Is there any reason I'm not aware for building pgspot from source instead of installing as a dev dependency ?

projects/extension/pyproject.toml
```diff
@@ -30,6 +30,7 @@ dev = [
     "fastapi-cli==0.0.5",
     "psycopg[binary]==3.2.1",
     "uv==0.5.20",
+    "pgspot>=0.9.0",
```
smoya, line 33: dev dependency

smoya:
@adolsalamanca @cevian I made another round, please check the description of the PR with an updated list of the changes. 🙏 🙏

projects/extension/Dockerfile
jgpruitt: There was at least one time when pgspot had a bug that broke us. I installed it this way because I was able to go fix the bug, but I didn't want to wait on the publishing process.

projects/extension/build.py
```diff
@@ -338,42 +338,48 @@ def lint_sql() -> None:
         """runs pgspot against the `ai--<this_version>.sql` file"""
         cmd = " ".join(
             [
-                "pgspot --ignore-lang=plpython3u",
+                "uv run pgspot --ignore-lang=plpython3u",
```
jgpruitt, line 341: needs the `--no-project` flag to `uv run`

jgpruitt (approved):

adolsalamanca (approved):

projects/extension/Dockerfile
smoya: I guess it's fixed now 👍

projects/pgai/justfile
```diff
@@ -33,23 +33,30 @@ uninstall:
 
 # Run pytest test suite
 test:
-	@uv run pytest
+	@uv run --no-project pytest
```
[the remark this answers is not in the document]
jgpruitt: awesome

PR #483, opened 2025-02-14
fix(vectorizer): use sql name for data types in source_pk

jgpruitt:
pg_type.typname works fine in SQL but chokes up psycopg3. Use pg_catalog.format_type to get the SQL name of a type instead.

1. create a source table with a primary key with an array type e.g. text[]
2. the `typname` from `pg_type` will be `_text`
3. SQL seems to handle `_text` fine as if it were an alias to `text[]`
4. `format_type()` will translate this back to `text[]`
5. the vectorizer worker is using the binary copy format and has to call `set_types(...)` in psycopg3
6. psycopg3 barfs all over `_text`. doesn't know what it is
7. psycopg3 does recognize `text[]`

- `pg_type.typname` is not unique. `typname` + `typnamespace` is unique
- We didn't have enough information in the original implementation of source_pk to identify a type uniquely
- The original implementation would fail if a type is defined in a schema off the search_path
- Using `format_type()` gives us the type in SQL parlance with schema if necessary

### DIFF
--- a/projects/extension/tests/vectorizer/test_vectorizer.py
+++ b/projects/extension/tests/vectorizer/test_vectorizer.py
@@ -65,7 +65,7 @@ VECTORIZER_ROW = r"""
             "pknum": 2,
             "attnum": 3,
             "attname": "published",
-            "typname": "timestamptz"
+            "typname": "timestamp with time zone"
         }
     ],
     "view_name": "blog_embedding",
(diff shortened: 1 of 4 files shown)


cevian (changes requested):
Can we add a test with a text[] pk column?

projects/extension/tests/vectorizer/test_vectorizer.py
```diff
@@ -1815,3 +1815,67 @@ def test_create_vectorizer_privs():
             , grant_to=>null
             );
             """)
+
+
+def test_weird_primary_key():
```
smoya, line 1820: WDYT about adding some context to this test? like the reason of being, why we are testing this change, etc. I know we always have github, but it will play in favour of developers when reading this test.

jgpruitt:
see https://github.com/timescale/pgai/pull/494 and https://github.com/timescale/pgai/pull/495

PR #484, opened 2025-02-14
feat: document retries working

adolsalamanca:
Implementation of the document failure retries, both for s3 problems, and LLM provider.


projects/pgai/pgai/cli.py
```diff
@@ -247,6 +248,13 @@ def shutdown_handler(signum: int, _frame: Any):
     show_default=True,
     help="Exit immediately when an error occurs.",
 )
+@click.option(
+    "--doc-retries",
```
smoya, line 252: Would it make sense to make this option agnostic of the loader? Like `--loading-retries` or `--load-retries` or similar?

projects/pgai/pgai/vectorizer/loading.py
```diff
@@ -52,3 +52,11 @@ def load(self, row: dict[str, str]) -> LoadedDocument:
             file_path=row[self.column_name],
             file_type=guess_filetype(content, row[self.column_name]),
         )
+
+
+class DocumentLoadingError(Exception):
```
smoya, line 57: WDYT about having a `LoadingError` base exception? In that way we can always raise an exception no matters the loader type.

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -370,6 +370,33 @@ def is_vectorizer_disabled_query(self) -> sql.Composed:
...
+                SELECT
+                    id
+                FROM
+                    {queue_table}
+                WHERE
+                    id = {vectorizer_id}) AS source ON target.id = source.id
+            WHEN MATCHED
+                AND target.retries >= %s THEN
+                    DELETE
+            WHEN MATCHED THEN
```
smoya, line 387: What about adding an extra check so the query won't ever remove a row if there is a backoff in place?
```suggestion
            WHEN MATCHED
                AND retry_after <= now() THEN
```

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -370,6 +370,33 @@ def is_vectorizer_disabled_query(self) -> sql.Composed:
             self.vectorizer_table_ident,
         )
 
+    @cached_property
+    def requeue_or_remove_work_query(self) -> sql.Composed:
```
smoya, line 374: I would love this function to be documented, mentioning we implement linear backoff, etc 🙏 WDYT?

projects/pgai/pgai/vectorizer/vectorizer.py
adolsalamanca: I just wrote a test for this case, it's already working, as this is handled in the fetch_work_query func:

```sql
                WITH selected_rows AS (
                    SELECT {pk_fields}
                    FROM {queue_table}
                    WHERE retry_after is null or retry_after < now()
                    LIMIT %s
                    FOR UPDATE SKIP LOCKED
                )
```

adolsalamanca: just pasted the initial part of the query not to make it not to pollute the conversation

projects/pgai/pgai/vectorizer/vectorizer.py
adolsalamanca: added, thanks!

projects/pgai/pgai/vectorizer/vectorizer.py
smoya: I still believe it is part of this query to protect itself against an unexpected behaviour

projects/pgai/pgai/vectorizer/loading.py
adolsalamanca: Sounds good, I'll switch to that.

projects/pgai/pgai/cli.py
adolsalamanca: sounds good!

PR #487, opened 2025-02-17
feat: add litellm to alembic and python vectorizer creation

Askir:
In the merge race when litellm and sqlalchemy/alembic was merged we forgot to regenerate the config classes, so did that now.


smoya (approved):
LGTM! 🚀🌔

PR #492, opened 2025-02-18
fix(build): use --no-project flag to uv run for build tasks

jgpruitt:
This PR addresses the bug below that we have been experiencing frequently in CI. When we use `uv run` to run tasks in CI, those tasks do NOT depend on the pgai project being installed. It seems `uv run` will install the pgai package by default since we are running `uv run` in a directory with a pyproject.toml. This PR adds the `--no-project` flag to `uv run` to avoid this behavior.

https://docs.astral.sh/uv/guides/scripts/#running-a-script-without-dependencies

image

### DIFF
--- a/projects/extension/build.py
+++ b/projects/extension/build.py
@@ -314,7 +314,11 @@ class Actions:
     def test() -> None:
         """runs the tests in the docker container"""
         subprocess.run(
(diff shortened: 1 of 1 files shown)


Askir (approved):
lets see if this helps. Still messy if that's whats needed. I think we aren't understanding this well enough rn.

PR #494, opened 2025-02-18
fix: load target column types' oids to use in binary copy cmd

jgpruitt:
We use the `COPY` command with the `BINARY` format in Vectorizer to load embeddings into the target table. To do so, we have to tell psycopg the types of the columns we are copying into. The `set_types` function can take type names or oids of types. We used type names but discovered this is incompatible with some types/domains we encountered.

This PR switches to using OIDs.

The pgai extension can continue to send types, but the vectorizer worker will ignore them. This is important as we need the worker to be backward-compatible with old extension versions. Instead, it will look up the OIDs of the types on the fly using the postgres catalog. It will do this once -- the first time it attempts to do a COPY.

I will do a follow-up PR to have the extension stop sending the type info.

https://www.psycopg.org/psycopg3/docs/api/copy.html#psycopg.Copy.set_types

This STILL does not work for EVERY type. I found that `bit(4)` did not work. psycopg loads a `bit(4)` into a Python str but will not support COPYing a str into a bit(4) via binary copy. Similarly, a `bpchar` type could not be loaded with pyscopg. To make these work, you must (possibly write) configure and register appropriate dumpers and loaders.

We need to think about how to enable the use of these types.

https://www.psycopg.org/psycopg3/docs/advanced/adapt.html

### DIFF
--- a/projects/pgai/pgai/vectorizer/vectorizer.py
+++ b/projects/pgai/pgai/vectorizer/vectorizer.py
@@ -54,11 +54,9 @@ class PkAtt:
 
     Attributes:
         attname (str): The name of the attribute (column).
-        typname (str): The type of the attribute.
     """
 
     attname: str
-    typname: str
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


cevian (approved):
I think this is a pareto improvement so approving. Would like to dig in a bit more why the other types fail though

PR #495, opened 2025-02-19
refactor(vectorizer): drop source_pk column and gen pk info dynamically

jgpruitt:
Use format_type() to get the correct rendering of the SQL data types for a vectorizer's primary key columns. This supports more types. Do not store this in the vectorizer table as it may change. Instead, generate it on the fly when needed.


jgpruitt:
see #497

PR #498, opened 2025-02-19
feat(text_to_sql): validate LLM authored sql statements by running EXPLAIN

jgpruitt:
also fixes an issue with `ai._search_semantic_catalog_obj`


projects/extension/sql/idempotent/909-text-to-sql-openai.sql
```diff
@@ -455,11 +466,40 @@ begin
...
+                                _prompt_err = concat_ws
+                                ( E'\n'
+                                , _prompt_err -- keep prior bad queries. append new ones
+                                , '<invalid-sql-statement>'
+                                , format(E'/* the following sql statement is not valid. error:\n%s\n*/', _err_msg)
+                                , _answer
+                                , '</invalid-sql-statement>'
+                                );
+                            end if;
+                        end if;
```
MasterOdin, line 495: I think we want to decrement `_max_iter` at least once when we hit a query error to allow for another iteration that hopefully the second generated query works.

MasterOdin (approved):
Changes make sense, one small comment, but can tackle in follow-up.

projects/extension/sql/idempotent/909-text-to-sql-openai.sql
jgpruitt: oops. that's a bug.

MasterOdin: Hah, I missed the bug you fixed (that the return would happen even on error), but I was meaning that let's say `_max_iter = 3`, then on the third iteration of the loop (`__iter_remaining = 0`), when we hit the continue and go back up to the top, then we'll exit the while loop and return `null`. I think that we should allow the LLM one final iteration to get it right and still force it to return _something_.

jgpruitt: ahhh! yes. okay. I'll put that in my next PR

PR #504, opened 2025-02-21
feat(text_to_sql): rank objs by relevancy and include obj name list i…

jgpruitt:
- adds a `usage` column to the `ai.semantic_catalog_obj` which gets incremented each time the obj is referenced in a text-to-sql result
- adds a list of N object names to the prompt ordered by relevancy rank desc
- include sample data by default when a table/view is described in the prompt
- render descriptions as SQL COMMENTs
- update the anthropic implementation of text-to-sql to be on par with openai


MasterOdin (approved):
I don't really like calling the column `relevancy_rank` as I feel like we're conflating what "relevancy" means as we also have the relevancy search for database objects / queries, which is different.

I'd maybe just call it "usage" or something? :shrug:

PR #507, opened 2025-02-24
build(extension): remove tmp dir before cloning into it

jgpruitt:
If you install and then uninstall and then try to install again (which I do during development), the second install will fail. We should delete the dir (if exists) we are cloning into before cloning into it.

image

### DIFF
--- a/projects/extension/build.py
+++ b/projects/extension/build.py
@@ -134,6 +134,7 @@ class Actions:
             if version_target_dir.exists():
                 continue
             tmp_dir = Path(tempfile.gettempdir()).joinpath("pgai", version)
+            shutil.rmtree(tmp_dir, ignore_errors=True)
             tmp_dir.mkdir(parents=True, exist_ok=True)
             branch = git_tag(version)
             subprocess.run(


MasterOdin (approved):

JamesGuthrie (commented):
We already have `shutil.rmtree(tmp_dir)` on line 156, suggesting that we should (in addition to this change) wrap the subsequent steps in a `try ... finally`:

```python3
        try:
            branch = git_tag(version)
            subprocess.run(
                f"git clone https://github.com/timescale/pgai.git --branch {branch} {tmp_dir}",
                shell=True,
                check=True,
                env=os.environ,
            )
            tmp_src_dir = tmp_dir.joinpath("projects", "extension").resolve()
            bin = "pip3" if shutil.which("uv") is None else "uv pip"
            cmd = f'{bin} install -v --compile --target "{version_target_dir}" "{tmp_src_dir}"'
            subprocess.run(
                cmd,
                check=True,
                shell=True,
                env=os.environ,
                cwd=str(tmp_src_dir),
            )
        finally:
            shutil.rmtree(tmp_dir)
```

PR #512, opened 2025-02-25
ci: check markdown links

smoya:


jgpruitt (approved):
If this works, it would be nice to add this to the dev docker images and add a `just` command for it. That way we can catch it during development.

https://github.com/lycheeverse/lychee

PR #514, opened 2025-02-26
feat: mount the whole repo in the pgai-ext container and add ai.version()

jgpruitt:
We want to add a way to get the git commit hash of the extension's build from a SQL function. To do this, the `.git` directory must be inside the dev container. With this PR, we now mount the whole repo into the extension dev container. This lets us access git information during the build process.

Additionally, this allows us to access the `examples` directory from the dev container, which is helpful in testing the examples (especially the text-to-sql one).

I moved the `.venv` python virtual environment to `/py/.venv` inside the dev container. This is NOT mounted to the host. Thus, the `.venv` inside the container and the `.venv` on the host cannot possibly clash.


projects/extension/Dockerfile
```diff
@@ -65,10 +66,9 @@ USER root
...
 
-WORKDIR /pgai
+## install only our python dependencies
+COPY pyproject.toml uv.lock /py/
+RUN uv sync --directory /py --no-install-project --frozen
+ENV PATH="/py/.venv/bin:$PATH"
 
-## install our test python dependencies
-COPY pyproject.toml uv.lock requirements-lock.txt build.py /pgai
-COPY ai /pgai/ai
```
MasterOdin, line 72: Don't we still want to copy in all of the other files of `ai`?

.github/workflows/ci.yml
```diff
@@ -46,22 +46,28 @@ jobs:
         run: just ext docker-run
 
       - name: Build
-        run: docker exec pgai-ext just build
+        run: docker exec 
+          -w /pgai/projects/extension
```
MasterOdin, line 50: Specifying `-w` for all usages feels like a mistake?

projects/extension/Dockerfile
jgpruitt: nah. The code in the ai module gets installed in `/usr/local/lib/pgai/<version>/ai`

Actually, I *think* all we need here are the dev/test dependencies.

.github/workflows/ci.yml
jgpruitt: yeah, i'm not sure it's necessary. I'm being paranoid.

projects/extension/Dockerfile
jgpruitt: yeah. all we need are the dependencies from the dev group

.github/workflows/ci.yml
jgpruitt: yep, now that i set the workdir in the dockerfile, it should work without this noise

.github/workflows/ci.yml
```diff
@@ -122,4 +124,5 @@ jobs:
       - name: CI pipeline. Install dependencies, run linters and formatters, execute tests and build the project", 
         run: just pgai ci
         env:
-          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
\ No newline at end of file
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          GITHUB_SHA: ${{ github.sha }}
```
adolsalamanca, line 128: nit, missing end of file new line

projects/extension/DEVELOPMENT.md
```diff
@@ -37,13 +37,14 @@ To make changes to pgai:
...
+2. Run the container:
    ```bash
    just ext docker-run
    ```
-   The `projects/extension` directory is mounted to `/pgai` in the running container.
+   The root of the repo directory is mounted to `/pgai` in the running container.
+   The working directory is at `/pgai/projects/extension`.
 
-1. Work inside the container:
+4. Work inside the container:
```
adolsalamanca, line 47: nit, we have skipped the number 3 here, and the same with the subsequent points too, all of them are +1 what they should be.

projects/pgai/justfile
```diff
@@ -33,7 +33,7 @@ uninstall:
 
 # Run pytest test suite
 test:
-	@uv run --no-project pytest
+	@uv run --no-project pytest -s
```
adolsalamanca, line 36: Is this intentional?

adolsalamanca (commented):
left a few comments

.github/workflows/ci.yml
jgpruitt: fixed

projects/extension/DEVELOPMENT.md
jgpruitt: fixed

projects/pgai/justfile
jgpruitt: it was for debugging, but i've removed.

adolsalamanca (approved):
neat

PR #518, opened 2025-02-27
ci: allow forks to use secrets under aproval

smoya:


.github/workflows/ci.yml
```diff
@@ -13,7 +13,16 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  authorize:
+    environment:
+      ${{ github.event_name == 'pull_request_target' &&
+      github.event.pull_request.head.repo.full_name != github.repository &&
+      'external-contributors' || 'internal-contributors' }}
```
adolsalamanca, line 20:
```suggestion
      'external-contributors'}}
```

adolsalamanca (commented):
This might be needed not to require approval for internal contributors requiring CI runs

.github/workflows/ci.yml
smoya: Nope, that shouldn't be the cause because that environment doesn't require so.
CleanShot 2025-02-27 at 12 29 40

adolsalamanca (approved):

cevian (approved):

PR #522, opened 2025-02-27
build: ensure temp dir deleted regardless of errors in install_prior_py

jgpruitt:
addresses @JamesGuthrie comment: https://github.com/timescale/pgai/pull/507#pullrequestreview-2637546410

### DIFF
--- a/projects/extension/build.py
+++ b/projects/extension/build.py
@@ -128,32 +128,32 @@ class Actions:
     def install_prior_py() -> None:
         """installs the extension's python package for prior versions"""
         for version in prior_versions():
-            if os.sep in version:
-                fatal(f"'{os.sep}' in version {version}. this is not supported")
             version_target_dir = python_install_dir().joinpath(version)
             if version_target_dir.exists():
                 continue
             tmp_dir = Path(tempfile.gettempdir()).joinpath("pgai", version)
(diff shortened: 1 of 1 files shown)


projects/extension/build.py
```diff
@@ -128,32 +128,32 @@ def install_sql(version: str | None = None) -> None:
     def install_prior_py() -> None:
         """installs the extension's python package for prior versions"""
         for version in prior_versions():
-            if os.sep in version:
-                fatal(f"'{os.sep}' in version {version}. this is not supported")
             version_target_dir = python_install_dir().joinpath(version)
             if version_target_dir.exists():
                 continue
             tmp_dir = Path(tempfile.gettempdir()).joinpath("pgai", version)
             shutil.rmtree(tmp_dir, ignore_errors=True)
```
MasterOdin, line 135: I think part of his suggestion was that we don't then need this line as the `finally` below will take care of it, and that the only way this folder should exist is if you created the folder manually.

JamesGuthrie (approved):

projects/extension/build.py
jgpruitt: yeah, we *shouldn't need it, but it doesn't hurt to have it

PR #534, opened 2025-03-03
chore: fix concurrency & add new workflow to rule them all

adolsalamanca:
Fixes concurrency relying on `github.event.pull_request.head.ref` given that
the current reference doesn't exist, but couldn't blame the author, github docs aren't great,
retrieved the info for pull_request events from [this](https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28#get-a-review-comment-for-a-pull-request) doc.

Then adds a new workflow so we can make it the only mandatory,
it'll make sure that all the triggered workflows have passed. 
This way we can avoid having to bypass checks merging docs,
which are no longer triggering code builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ permissions:
   contents: read
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.client_payload.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.head.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/merge-gatekeeper.yaml
+++ b/.github/workflows/merge-gatekeeper.yaml
@@ -1,0 +1,19 @@
+name: Merge Gatekeeper
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ permissions:
   contents: read
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.client_payload.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.head.ref }}
```
Askir, line 12: Does this end up being the branch name or the commit sha? If it's the later it means the PR workflow would not be cancelled for new pushes, not the end of the world but not quite perfect :D

Askir (approved):

.github/workflows/ci.yml
adolsalamanca: it's the branch name
```
      "head": {
        "label": "timescale:adol/add-merge-gatekeeper",
        "ref": "adol/add-merge-gatekeeper",
```

I investigated with this action:

```
steps:
  - name: Dump GitHub context
  uses: crazy-max/ghaction-dump-context@v1
```

PR #536, opened 2025-03-03
fix: respect OpenAI token limit

JamesGuthrie:
The OpenAI API has an undocumented limit on the maximum number of tokens that can be sent in a single API call: 600'000.

This change improves the batching logic to adhere to both constraints of max_chunks_per_batch and max_tokens_per_batch.

### DIFF
--- a/projects/pgai/pgai/vectorizer/embeddings.py
+++ b/projects/pgai/pgai/vectorizer/embeddings.py
@@ -1,4 +1,3 @@
-import math
 import time
 from abc import ABC, abstractmethod
 from collections.abc import Awaitable, Callable, Sequence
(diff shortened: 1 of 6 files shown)


projects/pgai/pgai/vectorizer/embeddings.py
```diff
@@ -159,6 +205,13 @@ def _max_chunks_per_batch(self) -> int:
         :return: int: the max chunk count
         """
 
+    def _max_tokens_per_batch(self) -> int | None:
+        """
+        The maximum number of tokens that can be embedded per API call
+        :return: int: the max token count
+        """
+        return None
```
Askir, line 213: This whole abstraction of embedders seems to fall apart, with openai just being a super special case because we do tokenization. I feel like a simple embed function per provider, would be cleaner than the inheritance and overrides etc.

But nothing for now ofc.

Askir (approved):
lgtm!

projects/pgai/pgai/vectorizer/embeddings.py
JamesGuthrie: I agree that it's a PITA.

adolsalamanca (approved):

PR #549, opened 2025-03-06
fix: error "Object of type UUID is not JSON serializable"

JamesGuthrie:
When an error occurs while chunking, the error is packaged and stored in the database with some additional metadata. During this process, we attempted to put a UUID value into a json object to store in a column. Psycopg doesn't support this by default, so we need to transform the uuid to text.


JamesGuthrie:
Note: I went back and forth on whether to fix this globally or locally.

The actual error comes from one place, here:

https://github.com/timescale/pgai/blob/0cb8296adcfc2d0a5efb40cc2b127899ecb43a4c/projects/pgai/pgai/vectorizer/vectorizer.py#L802-L812

It would be possible to use the `dumps=<dumper>` parameter in the construction of `Jsonb()` which would fix this error locally. I considered that we may accidentally end up doing the same thing again in another place, causing silent breakage, which made me err on fixing it globally.

projects/pgai/pgai/cli.py
```diff
@@ -269,6 +284,8 @@ def vectorizer_worker(
     dynamic_mode = len(vectorizer_ids) == 0
     valid_vectorizer_ids = []
 
+    set_json_dumps(partial(json.dumps, cls=UUIDEncoder))
```
Askir, line 287: It seems strange to me that psycopg successfully parses the postgres datatype to a python UUID object but then struggles to convert it to json.

The way you fix it seems very specific to UUIDs which I guess will be a common type for primary keys but there could be other that are missing a conversion right?

Should we just resort to always do a string conversion if an exception is thrown during json conversion? Or should the errors table maybe contain the primary keys as columns directly instead of nested in a jsonb?

Askir (approved):

projects/pgai/pgai/cli.py
JamesGuthrie: > It seems strange to me that psycopg successfully parses the postgres datatype to a python UUID object but then struggles to convert it to json.

Yes, this is because psycopg has a built-in "[uuid adaptation](https://www.psycopg.org/psycopg3/docs/basic/adapt.html#uuid-adaptation)", for column values. But it doesn't have a built-in uuid-in-json adaptation.

> The way you fix it seems very specific to UUIDs which I guess will be a common type for primary keys but there could be other that are missing a conversion right?

Correct.

> Should we just resort to always do a string conversion if an exception is thrown during json conversion?

We could also just always do a string conversion of all values - it's in a jsonb, I don't think that preserving the type is that important.

```python
 Jsonb( 
     { 
         "pk": { 
             pk_attname: str(record[i])
             for i, pk_attname in enumerate(self.queries.pk_attnames) 
         }, 
         "chunk_id": record[-2], 
         "chunk": record[-1], 
         "error_reason": chunk_error.error_details, 
     } 
 ), 

```

> Or should the errors table maybe contain the primary keys as columns directly instead of nested in a jsonb?

I think that would be the most robust solution, but I hate to think of how we would handle compatibility between the vectorizer worker and the db schema.

JamesGuthrie: @jgpruitt any thoughts?

Askir: > We could also just always do a string conversion of all values

That also works, but I'd then tie this more specifically to the errors insert I think. For other jsonbs the type might be more important and a generic "everything is a string" handling might be a bit surprising?

jgpruitt: I don't know that we can stringify everything. I think we'd want to preserve numbers, true, false, and null as those specific json values (i.e. not strings). Also, stringifying a Json value or an array might be problematic or at least annoying.

JamesGuthrie:
Merging as-is. Can always improve later.

PR #560, opened 2025-03-12
fix(text_to_sql): deal with changes to ai.openai_chat_complete

jgpruitt:
The signature of the `ai.openai_chat_complete` function changed and broke text-to-sql. This fixes it

### DIFF
--- a/projects/extension/sql/idempotent/909-text-to-sql-openai.sql
+++ b/projects/extension/sql/idempotent/909-text-to-sql-openai.sql
@@ -449,7 +449,7 @@ begin
             end
         , api_key=>_api_key
         , api_key_name=>_api_key_name
-        , base_url=>_base_url
+        --, base_url=>_base_url
         , frequency_penalty=>_frequency_penalty
         , logit_bias=>_logit_bias
         , logprobs=>_logprobs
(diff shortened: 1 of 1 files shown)


projects/extension/sql/idempotent/909-text-to-sql-openai.sql
```diff
@@ -460,7 +460,7 @@ begin
         , stop=>_stop
         , temperature=>_temperature
         , top_p=>_top_p
-        , openai_user=>_openai_user
+        --, openai_user=>_openai_user
```
MasterOdin, line 463: Would the fix here just be `client_config => jsonb_build_object('base_url', _base_url, 'openai_user', _openai_user)` vs totally removing these?

Or is it that we need to not pass these values in if `null`?

MasterOdin (approved):
Changes seem fine, I think fine to not support `base_url` or `openai_user` if it's at all difficult to do given the API is probably going to change anyway in the future.

projects/extension/sql/idempotent/909-text-to-sql-openai.sql
jgpruitt: Yes, that would be the best fix, but since this is all changing and we aren't using this feature, I just did the fastest thing possible to get it working again

PR #563, opened 2025-03-14
test: simplify cohere tests

smoya:
Since this test is failing randomly very often, and based on the confirmation we got from Cohere team, the result of the chat complete can't be deterministic at 100%. It means this test will keep failing randomly in time. 

 In our particular case, we don't need to evaluate the result but just checking we do have a result.

### DIFF
--- a/projects/extension/tests/test_cohere.py
+++ b/projects/extension/tests/test_cohere.py
@@ -1,5 +1,4 @@
 import os
-from time import sleep
 
 import psycopg
 import pytest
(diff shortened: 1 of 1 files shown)


projects/extension/tests/test_cohere.py
```diff
@@ -364,26 +363,19 @@ def test_cohere_rerank_simple(cur_with_api_key):
...
+            ( 'role', 'user'
+            , 'content', 'How much wood would a woodchuck chuck if a woodchuck could chuck wood?'
+            )
+          )
+        , seed=>42
+        , temperature=>0.0
+        )->'message'->'content'->0->>'text'
+    """)
+    result = cur_with_api_key.fetchone()[0]
+    assert result != "" or None
```
alejandrodnm, line 381: You can just do `assert result`, it'll fail if the result is "" or None.

alejandrodnm (approved):

projects/extension/tests/test_cohere.py
alejandrodnm: `""` `[]` `None` will resolve to `False`

PR #569, opened 2025-03-17
fix(vectorizer): always use column ordering from table rather than primary key ordering

jgpruitt:
A table can have columns in order `A, B, C, D` and a primary key defined in order `A, D, C, B`. When we select items from the source table, we get the records using table column order (not primary key order). When copying into the target table, we were using primary key ordering to construct the SQL command, but table column order to provide the values. This is a bug.

### DIFF
--- a/projects/pgai/pgai/vectorizer/vectorizer.py
+++ b/projects/pgai/pgai/vectorizer/vectorizer.py
@@ -3,9 +3,8 @@ import json
 import os
 import threading
 import time
-from collections.abc import Callable
+from collections.abc import Callable, Sequence
 from functools import cached_property, partial
-from itertools import repeat
 from typing import Any, TypeAlias
 from uuid import UUID
 
@@ -61,6 +60,8 @@ class PkAtt:
     """
 
     attname: str
+    pknum: int
+    attnum: int
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


smoya (approved):

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -717,6 +742,8 @@ async def _copy_embeddings(
             conn (AsyncConnection): The database connection.
             records (list[EmbeddingRecord]): The embedding records to be copied.
         """
+        if self.copy_types is None:
+            await self._load_copy_types(conn)
```
JamesGuthrie, line 746: This duplicates logic in the block below. Either move this there, or remove the whole `if self.copy_types is None` below.

JamesGuthrie (changes requested):
Looks good in general - just one cleanup.

JamesGuthrie (commented):
Also: I wasn't able to run the `test_vectorizer_weird_pk` without adding `pytest-asyncio` as a dev dependency, and adding the following line:

```
asyncio_mode = "auto"
```

to the existing `[tool.pytest.ini_options]` block.

How are you running that test?

projects/pgai/pgai/vectorizer/vectorizer.py
jgpruitt: whoops. doh.

JamesGuthrie (approved):

PR #570, opened 2025-03-18
test: add vectorizer benchmark tools

alejandrodnm:
Add 2 new benchmark recipes for the vectorizer:

- `just pgai benchmark-mem`: runs the vectorizer on a queue of 500 items, generates a memray profile, and a flamegraph.

- `just pgai benchmark-cpu`: runs the vectorizer on a queue of 500 items, generates a flamegraph using py-spy.

The benchmark DB uses the wiki.dump stored in the repository, creates an openAI vectorizer, and runs the benchmark tool storing the results in `projects/pgai/benchmark/results`.


alejandrodnm:
I'm working on adding a new cli command that runs the vectorizer but uses vcr to replay the openAI responses. That way we can remove openAI's response time when measuring how fast we can process embeddings. 

The cassette is ~500MB so I'm going to try git large files support, or maybe hosting in S3.

projects/pgai/.python-version
```diff
@@ -1 +1 @@
-3.10
+3.12.9
```
JamesGuthrie, line 1: Why did this change?

projects/pgai/pgai/cli.py
```diff
@@ -220,6 +323,7 @@ def cli():
 
 vectorizer.add_command(vectorizer_worker)
 vectorizer.add_command(download_models)
+vectorizer.add_command(vectorizer_worker_benchmark)
```
JamesGuthrie, line 326: What is the purpose of this?

JamesGuthrie: Can we move this functionality into the "benchmark" project?

projects/pgai/.python-version
alejandrodnm: I'm reverting it back, I updated it for a test. 

We are using 3.12 for the docker container and I wanted the same,

projects/pgai/pgai/cli.py
alejandrodnm: Moved it to the benchmark dir.

alejandrodnm:
@JamesGuthrie I moved the part of vcr to the benchmark repo, and updated the recipes and create_vectorizer.sql to be more customizable.

PR #571, opened 2025-03-19
perf: use openAI raw response to reduce CPU consumption

alejandrodnm:
The openAI regular client API performs a lot of typecast operations when parsing API responses. On big responses, like what the vectorizer is expected to handle, it uses considerable CPU time to typecast and assert types.

The openAI embedder now uses an API that returns raw responses. The raw responses are decoded into a dict instead of custom types, which uses considerable less CPU.

https://www.loom.com/share/13043b12c2094871bad55a1fcef1cfc9?sid=383f3e28-83ac-4900-a9a5-c0a618f81b98


JamesGuthrie (approved):
🎉

PR #572, opened 2025-03-19
feat: add vectorizer.run method

smoya:
This PR moves the `cli.run_worker` function to a `.run` method in the Vectorizer class. 
This new method can be called from any port, such as CLI or a Lambda, so this becomes the main entry point for processing a vectorizer.

For example this is the main difference in the changes on the methods API: 

Previous call:
```python
await cli.run_vectorizer(
    db_url, vectorizer_actual, 1, features, worker_tracking
)
```

New call:
```python
await vectorizer_actual.run(db_url, features, worker_tracking, 1)
```

### DIFF
--- a/projects/pgai/pgai/cli.py
+++ b/projects/pgai/pgai/cli.py
@@ -20,7 +20,7 @@ from pytimeparse import parse  # type: ignore
 from .__init__ import __version__
 from .vectorizer.embeddings import ApiKeyMixin
 from .vectorizer.features import Features
-from .vectorizer.vectorizer import Vectorizer, Worker
+from .vectorizer.vectorizer import Vectorizer
 from .vectorizer.worker_tracking import WorkerTracking
 
 load_dotenv()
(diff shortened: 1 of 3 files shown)


adolsalamanca (approved):

projects/pgai/pgai/cli.py
```diff
@@ -250,6 +213,13 @@ def shutdown_handler(signum: int, _frame: Any):
     show_default=True,
     help="Exit after processing all available work (implies --exit-on-error).",
 )
+@click.option(
+    "-c",
+    "--concurrency",
+    type=click.IntRange(1),
+    default=None,
```
smoya, line 220: Making it optional, since we are either either using the concurrency configured in the vectorizer config or its default (1).

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -122,6 +122,62 @@ class Vectorizer:
...
+            worker_tracking: Tracking instance for worker stats
+            concurrency: Number of concurrent workers
+                (overrides vectorizer config if provided)
+            should_continue_processing_hook: Optional callback to
+                control processing flow
+
+        Returns:
+            Number of items processed
+        """
+        concurrency = concurrency or self.config.processing.concurrency
```
smoya, line 147: Here it is the fallback I mentioned in https://github.com/timescale/pgai/pull/572/files#r2005248136

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -122,6 +122,62 @@ class Vectorizer:
     schema: str = "ai"
     table: str = "vectorizer"
 
+    async def run(
```
smoya, line 125: This is merely a copy of the old `cli.run_vectorizer` method but grabbing the data from the vectorizer itself instead of requiring it as arguments as the old function was doing.

PR #573, opened 2025-03-19
fix: make features more robust

cevian:


JamesGuthrie (approved):
I like the direction of this change.

PR #574, opened 2025-03-19
fix: exclude future links from being analyzed

adolsalamanca:
Our markdown links checker is failing on releases, due to release-please generating yet non existing links. Exclude those types of links from being analyzed.
Also stop running CI for mere changes in workflows.

Workflow failing on release, due to future, non existing links:
https://github.com/timescale/pgai/actions/runs/13950810027

And passing after testing the fix:
https://github.com/timescale/pgai/actions/runs/13950938884/job/39049990520

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,8 @@ on:
   pull_request_target:
     paths-ignore:
       - "**.md"
+      - ".github/workflows/*.yml"
+      - ".github/workflows/*.yaml"
   workflow_dispatch:
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
adolsalamanca, line 1: btw, we should probably converge to just using yaml for the workflows

.github/workflows/ci.yml
```diff
@@ -4,6 +4,8 @@ on:
   pull_request_target:
     paths-ignore:
       - "**.md"
+      - ".github/workflows/*.yml"
+      - ".github/workflows/*.yaml"
```
JamesGuthrie, line 8: Why did you add this ignore? Is it because these changes to the workflows won't be run in a `pull_request_target` action anyway?

adolsalamanca: to prevent CI from running in case we're merely adjusting the CI, like in this PR.

PR #576, opened 2025-03-20
feat: allow failing on first exception on vectorizer.run

smoya:
This PR allows configuring the behaviour on the `vectorizer.run` method in terms of failing fast or rather waiting for all workers to finish their work. 

For example, the cli.py current behaviour is to wait for all workers to finish and then process their exceptions. However, other software may want to use another behaviour and fail fast.


adolsalamanca:
We have agreed not do do this and instead merged [this](https://github.com/timescale/pgai/pull/577) PR, closing.

PR #579, opened 2025-03-21
Fix huggingface ci

adolsalamanca:
it's currently broken given the repo wasn't cheked out.


adolsalamanca:
The only green runs are those where the checkout and the right cache path was specified, [link](https://github.com/timescale/pgai/actions/workflows/huggingface-dataset.yml)

Askir (approved):

PR #581, opened 2025-03-24
fix: force runtime chunk exclusion in fetch_work_query

JamesGuthrie:
Using lateral join allows for constraints to be pushed into the query against the source table, which can be used to perform runtime chunk exclusion of chunks in a (compressed) hypertable, decreasing time take to fetch a chunk of work items.


adolsalamanca (commented):
could it make sense to exclude those items in the part of the statement where we acquire the advisory lock?

JamesGuthrie:
> could it make sense to exclude those items in the part of the statement where we acquire the advisory lock?

No, the chunk exclusion is only applicable to the source table, but the part where we're acquiring the advisory lock is operating only on the queue table.

cevian (approved):

PR #582, opened 2025-03-25
feat: add destination config, allows saving embeddings to source table

Askir:
Based on: ~https://github.com/timescale/pgai/pull/592 (which needs some discussion still)~ #589 
Docs not yet done, will do in a separate one.


projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -997,23 +1026,39 @@ async def _copy_embeddings(
...
-            await self._load_copy_types(conn)
-        async with (
-            conn.cursor(binary=True) as cursor,
-            cursor.copy(self.queries.copy_embeddings_query) as copy,
-        ):
-            assert self.copy_types is not None  # ugh. make pyright happy
-            copy.set_types(self.copy_types)
-            for record in records:
-                await copy.write_row(record)
+        if self.vectorizer.config.destination.implementation == "default":
```
cevian, line 1038: nit: I'd break this up into two submethods called by this "switch"

cevian: Later (not in this PR) we should even refactor into separate classes

projects/pgai/db/sql/incremental/026-remove-destination-columns.sql
```diff
@@ -0,0 +1,54 @@
...
+    end loop;
+end;
+$block$;
+
+-- These will be recreated by the idempotent migrations in new form that work despite the dropped columns
+drop view if exists ai.vectorizer_status; 
+drop event trigger if exists _vectorizer_handle_drops;
+drop function if exists ai._vectorizer_handle_drops;
+
+alter table ai.vectorizer
```
cevian, line 44: does this auto-drop the unique constraint? How does that dropping work?

projects/pgai/db/sql/incremental/026-remove-destination-columns.sql
```diff
@@ -0,0 +1,54 @@
...
+                'target_schema': _target_schema,
+                'target_table': _target_table,
+                'view_schema': _view_schema,
+                'view_name': _view_name
+        ));
+
+        -- Update the vectorizer with new config
+        update ai.vectorizer 
+        set config = _config
+        where id operator(pg_catalog.=) _vectorizer.id;
```
cevian, line 34: do we need to update the version of the config here?

projects/pgai/db/sql/idempotent/011-vectorizer-int.sql
```diff
@@ -96,6 +96,144 @@ language plpgsql volatile security invoker
 set search_path to pg_catalog, pg_temp
 ;
 
+-------------------------------------------------------------------------------
+-- _vectorizer_create_destination
+create or replace function ai._vectorizer_create_destination
```
cevian, line 101: I would def have separate functions for each destination here

projects/pgai/db/sql/idempotent/010-destination.sql
```diff
@@ -0,0 +1,69 @@
...
+    , 'view_name': view_name
+    absent on null
+    )
+$func$ language sql immutable security invoker
+set search_path to pg_catalog, pg_temp
+;
+
+-------------------------------------------------------------------------------
+-- destination_source
+create or replace function ai.destination_source
```
cevian, line 28: maybe this should be `destination_embedding_column` or  `destination_column`

cevian (changes requested):
Overall great work. 2 high-level comments
- I think we need better names than destination_default and destination_source
- I'd like to have some of the functions that now have a giant top-level switch on destination to distination_belonging functions (which should be reflected in name). so `ai._vectorizer_create_destination` should become `ai._destination_source_create` and `ai._destination_default_create`. The idea is we'd have a bunch of functions that "belong" to one class or the other`. That will make the code easier to read and reason about. 

We also discussed creating a unique name for each vectorizer but that can be a followup pr

projects/pgai/pgai/vectorizer/vectorizer.py
Askir: done ✅ I also moved the destination into the params so we can rely on typing rather than is instance checks. this ensure we don't accidentally call the wrong methods. But yeh this needs to be refactored further eventually. Probably when we do customization we touch this anyway.

projects/pgai/db/sql/idempotent/011-vectorizer-int.sql
Askir: done ✅

projects/pgai/db/sql/idempotent/010-destination.sql
Askir: done ✅

projects/pgai/db/sql/incremental/026-remove-destination-columns.sql
Askir: We do this in the 022 migration already which hasn't been released yet. So I don't think we have to do it here as well. That's honestly a bit of a mess... but it works.

I still feel like just not updating the versions and making the migration system simply work on the existence of keys instead of distinct versions, would be more robust. But I got voted out on that :D

projects/pgai/db/sql/idempotent/011-vectorizer-int.sql
```diff
@@ -444,6 +581,8 @@ create or replace function ai._vectorizer_build_trigger_definition
 , queue_table pg_catalog.name
 , target_schema pg_catalog.name
 , target_table pg_catalog.name
+, source_schema pg_catalog.name
```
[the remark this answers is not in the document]
Askir: I'm not so sure about this either.
I thought about separating it out but I'd prefer to not have all the interpolation and especially the _relevant_columns_check defined twice. I could pull them out into all separate functions and compose from there but then we have a lot of very tiny functions which does not seem to be the style we follow.

Alternatively I could define the two template functions separately and then call the same replacement logic on top in this function, but I feel like that is also hard to follow control flow wise. Open for suggestions, so far I think having everything in one place might look a bit messy but is the easiest way to understand what's going on imo.

projects/pgai/db/sql/idempotent/011-vectorizer-int.sql
```diff
@@ -463,59 +604,109 @@ begin
...
+        _func_def := replace(_func_def, '$TARGET_TABLE$', quote_ident(target_table));
+        _func_def := replace(_func_def, '$RELEVANT_COLUMNS_CHECK$', _relevant_columns_check);
+        _func_def := replace(_func_def, '$TRUNCATE_STATEMENT$', _truncate_statement);
+    
+    else
+        _func_def := $def$
+        begin
+            if (TG_LEVEL = 'ROW') then
+                if (TG_OP = 'UPDATE') then
+                    if $RELEVANT_COLUMNS_CHECK$ then
```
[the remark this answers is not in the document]
cevian: don't we delete rows in the embedding table if the rows changed?  Can you check with @jgpruitt I think we used to.

projects/pgai/db/sql/idempotent/011-vectorizer-int.sql
```diff
@@ -617,6 +822,15 @@ begin
     )
     loop
         raise notice 'Recreating trigger function for vectorizer ID %s', _vec.id;
+        
+        _destination_type := _vec.config->'destination'->>'implementation';
+        if _destination_type = 'table' then
+            _target_schema := _vec.config->'destination'->>'target_schema';
+            _target_table := _vec.config->'destination'->>'target_table';
+        else
+            raise notice 'Skipping vectorizer ID %s because it has a non-table destination', _vec.id;
```
cevian, line 831: wait we have a trigger on destination_column vectorizers so how is this right?

projects/pgai/db/sql/idempotent/011-vectorizer-int.sql
```diff
@@ -967,13 +1197,22 @@ begin
     set local search_path = pg_catalog, pg_temp;
 
     -- if the conditions are right, create the vectorizer index
-    if ai._vectorizer_should_create_vector_index(_vec) then
+    if ai._vectorizer_should_create_vector_index(_vec) and _vec.config operator(pg_catalog.->) 'destination' operator(pg_catalog.->>) 'implementation' operator(pg_catalog.=) 'table' then
```
cevian, line 1200: nit: don't love that we have to call `ai._vectorizer_should_create_vector_index` twice: in the if and the elseif. Can we "cache" that in a variable?

projects/pgai/db/sql/idempotent/012-vectorizer-api.sql
```diff
@@ -603,8 +574,8 @@ create or replace view ai.vectorizer_status as
 select
   v.id
 , pg_catalog.format('%I.%I', v.source_schema, v.source_table) as source_table
-, pg_catalog.format('%I.%I', v.target_schema, v.target_table) as target_table
-, pg_catalog.format('%I.%I', v.view_schema, v.view_name) as "view"
+, pg_catalog.format('%I.%I', v.config operator(pg_catalog.->) 'destination' operator(pg_catalog.->>) 'target_schema', v.config operator(pg_catalog.->) 'destination' operator(pg_catalog.->>) 'target_table') as target_table
```
cevian, line 577: should the target_table or "view" be the source table in the dest_column case? should we add an "embedding_column" info column to the view?

cevian (changes requested):
some changes/questions mostly minor

projects/pgai/db/sql/idempotent/011-vectorizer-int.sql
```diff
@@ -96,6 +96,151 @@ language plpgsql volatile security invoker
...
+        join pg_catalog.pg_class c on a.attrelid = c.oid
+        join pg_catalog.pg_namespace n on c.relnamespace = n.oid
+        where n.nspname = source_schema
+        and c.relname = source_table
+        and a.attname = embedding_column
+        and not a.attisdropped
+    ) into _column_exists;
+
+    if _column_exists then
+        raise exception 'embedding column %I already exists in %I.%I', embedding_column, source_schema, source_table;
```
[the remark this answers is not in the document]
Askir: I changed it. My only concern is that this doesn't guarantee that the user uses the correct type (especially the dimension count).

projects/pgai/db/sql/idempotent/011-vectorizer-int.sql
Askir: This is just the recreation logic, since we don't yet have any destination column vectorizers this wasn't really necessary. But it actually is easy now since the definition just relies  on target table and target schema being null to create the correct trigger definition. So fixed now.

projects/pgai/db/sql/idempotent/012-vectorizer-api.sql
Askir: I added the embedding_column info column, I'm not so sure about putting the source table also into the target table column, that seems a bit unintuitive to me?

cevian (approved):

PR #590, opened 2025-03-27
chore: Use BaseModel everywhere instead of mixing it with pydantic dataclasses

Askir:


projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -150,7 +145,7 @@ class Vectorizer:
     queue_failed_table: str | None = None
     errors_schema: str = "ai"
     errors_table: str = "vectorizer_errors"
-    schema: str = "ai"
+    schema_: str = Field(alias="schema", default="ai")
```
JamesGuthrie, line 148: Why do we have to rename this attribute?

JamesGuthrie (approved):
LGTM, one question.

projects/pgai/pgai/vectorizer/vectorizer.py
Askir: Because pydantics base models have a schema method. The method is deprecated and using the field name is actually fine but I was getting type errors for it.
Instead of having a random type ignore I took this approach. We don't even seem to use the field so I think we can also just remove it?

PR #592, opened 2025-03-27
chore: compatibility with old vectorizers without migration framework

Askir:
This just uses pydantics before validator to directly wrangle the input dict into a compatible format.
If you want to add more migrations you'd just add another validator at the appropriate level. For ordering see:
https://docs.pydantic.dev/latest/concepts/validators/#ordering-of-validators


projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -94,22 +93,29 @@ class Config(BaseModel):
...
-
-        logger.warning("Unable to migrate configuration: raw data type is unknown")
-        return data  # type: ignore[reportUnknownVariableType]
+        if (
+            isinstance(data, dict)
+            and all(isinstance(key, str) for key in data)  # type: ignore
+            and data.get("loading", None) is None  # type: ignore
+        ):  # type: ignore
+            try:
+                column = data.get("chunking").get("chunk_column")  # type: ignore
```
JamesGuthrie, line 111: Should we not also delete this key?

Askir: Pydantic doesn't care about additional keys (unless you configure it to do so) so I don't think that's necessary.

Askir: I'm not quite sure I understand what you are getting at with this:
> make assumptions about the structure of the data they're working with

- We don't write this back to the db (that's what the sql migrations do)
- Another (later or deeper) validator might also want to assume that if the vectorizer is old, the old keys are still there and it can read from there.

I find removing it actually more confusing. When I'm writing a new validator not only would I need to make sure it works for the format that is in the db, but I'd also need to understand all the mutations the other validators are already doing.
Sure there might be cases where we e.g. change the type of a field in which this is unavoidable but in general I'd rather read from the original fields (which is impossible if another validator deletes it).

Either way it feels all very hypothetical, I don't feel that strongly about it. I can also add the delete and if later on it annoys me I remove it again.

Askir: I think there is a difference between the validator output being compatible with the new schema and actual equivalence of the dicts.

E.g. when creating a new vectorizer we always add the autoparser into the config dict. But old vectorizers simply work because the parser also has a default value. We didn't write a validator migration for this, should we do that? If yes we should then also run pdyantic in strict mode and remove any kind of coercing: https://docs.pydantic.dev/latest/concepts/strict_mode/

I don't think so though, I think it makes our life easier to not enforce a strict structure and instead be happy with compatibility.

Askir:
We had a call, I wasn't able to convince sergio and james that this wont be very chaotic. So we decided to go with #589

PR #593, opened 2025-03-28
chore: submit plain (not tokenized) input to OpenAI

JamesGuthrie:
The OpenAI embedding API supports submitting both plain text, or pre-tokenized inputs. No other embedding providers support submitting tokenized input, and the internal complexity required to support both doesn't seem worth it. For this reason, we're switching to submitting plain input to the OpenAI API.


projects/pgai/pgai/vectorizer/embedders/openai.py
```diff
@@ -121,24 +111,21 @@ async def embed(self, documents: list[StringDocument]) -> Sequence[EmbeddingVect
...
         encoder = self._encoder
         context_length = self._context_length
-        is_tokenized = False
-        processed_docs: list[Document] = []
         if encoder is not None and context_length is not None:
+            token_counts: list[int] = []
             # truncate all documents before submitting them to the API
-            is_tokenized = True
-            for document in documents:
+            for i, document in enumerate(documents):
```
Askir, line 117: Would it make sense to rename documents to chunks? considering we have the actual file-document support now I find the terminology a bit confusing.

Askir (approved):
looks and sounds great to me, we still have the cut-off in place so that's also good.

Askir:
tests are red due to some network issues probably on githubs side

JamesGuthrie:
Something is not right with the extension tests. I can reproduce the errors related to `ai.anthropic_list_models()`, but I have no idea what's going on. It's not related to this change, so I'm going to merge this as-is.

PR #597, opened 2025-04-01
fix: print error messages from grouped exceptions

### LINKED ISSUE #594: [Bug]: error messages not being logged
### What happened?

When an error happens, the exceptions are collected by an exception group. The exception group doesn't print the error messages from all the grouped exceptions, making it hard to debug vectorizer issues. 

### pgai extension affected

_No response_

### pgai library affected

_No response_

### PostgreSQL version used

All

All

Not applicable

Not applicable

```bash

```

```bash
Raise an exception
```

🦸 Yes , I will submit a PR soon!

alejandrodnm:
Grouped exceptions were not being printed, just a message saying that the vectorizer failed with a certain number of sub-exceptions.

Adds vectorizer_id to errors whenever possible.

closes #594

### DIFF
--- a/projects/pgai/pgai/vectorizer/processor.py
+++ b/projects/pgai/pgai/vectorizer/processor.py
@@ -2,6 +2,7 @@ import asyncio
 import datetime
 import os
 import random
+import sys
 import traceback
 from collections.abc import Sequence
 from dataclasses import dataclass
@@ -15,6 +16,12 @@ from .embeddings import ApiKeyMixin
 from .features import Features
 from .vectorizer import Vectorizer
 from .worker_tracking import WorkerTracking
+
+if sys.version_info >= (3, 11):
+    from builtins import BaseExceptionGroup
+else:
+    # For Python 3.10 and below, use the backport
+    from exceptiongroup import BaseExceptionGroup
 
 logger = structlog.get_logger()
 
@@ -132,7 +139,10 @@ class Processor:
         vectorizer_id: int | None,
         worker_tracking: WorkerTracking | None,
     ) -> Exception:
-        logger.error(error_message)
+        kwargs = {}
+        if vectorizer_id:
+            kwargs["vectorizer_id"] = vectorizer_id
+        logger.error(error_message, **kwargs)
         if worker_tracking is not None:
             await worker_tracking.save_vectorizer_error(vectorizer_id, error_message)
         if self.exit_on_error:
@@ -173,6 +183,7 @@ class Processor:
         worker_tracking = None
 
         while True:
+            vectorizer_id = None
             try:
                 if not can_connect or pgai_version is None:
                     with (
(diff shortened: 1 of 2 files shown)


JamesGuthrie (approved):

projects/pgai/pgai/cli.py
```diff
@@ -378,14 +388,24 @@ async def async_run_vectorizer_worker(
                 err_msg = f"unable to connect to database: {str(e)}"
             else:
                 err_msg = f"unexpected error: {str(e)}"
-            await handle_error(err_msg, None, worker_tracking, exit_on_error)
+            await handle_error(err_msg, vectorizer_id, worker_tracking, exit_on_error)
+        except BaseExceptionGroup as e:  # type: ignore
+            # catch any exceptions, log them, and keep on going
+            for exception in e.exceptions:  # type: ignore
+                for exception_line in traceback.format_exception(e):  # type: ignore
```
JamesGuthrie, line 395:
```suggestion
                for exception_line in traceback.format_exception(exception):  # type: ignore
```

JamesGuthrie (changes requested):

projects/pgai/pgai/cli.py
alejandrodnm: Wait, I'll double check this, because I think I'm running python 3.10 and in the builtin version the traceback (I think) shows. the subexceptions

projects/pgai/pgai/cli.py
```diff
@@ -378,14 +388,24 @@ async def async_run_vectorizer_worker(
...
-            await handle_error(err_msg, None, worker_tracking, exit_on_error)
+            await handle_error(err_msg, vectorizer_id, worker_tracking, exit_on_error)
+        except BaseExceptionGroup as e:  # type: ignore
+            # catch any exceptions, log them, and keep on going
+            for exception in e.exceptions:  # type: ignore
+                error_msg = str(exception)  # type: ignore
+                log.error(error_msg, vectorizer_id=vectorizer_id)
+            for exception_line in traceback.format_exception(e):  # type: ignore
+                for line in exception_line.rstrip().split("\n"):
+                    log.debug(line)
```
alejandrodnm, line 399: @JamesGuthrie We support python 3.10 as the minimum version, the builtin version of `BaseExceptionGroup` was added in 3.11.

When using the builtin version, the traceback includes the stacktraces of all the grouped exceptions. The issue is only with version from 3.10, which just has the stacktrace of where BaseExceptionGroup was raised. In either case we do still need to log the errors.

We are using python3.12 in the Docker image, so I'd just leave it like this. Maybe even raise our python version to 3.11.

projects/pgai/pgai/cli.py
alejandrodnm: Take a look at https://github.com/timescale/pgai/pull/597#discussion_r2022926476

projects/pgai/pgai/vectorizer/processor.py
```diff
@@ -245,7 +256,23 @@ async def run(self) -> Exception | None:
...
+                    err_msg, vectorizer_id, worker_tracking
+                )
+                if self.exit_on_error:
+                    return exception
+            except BaseExceptionGroup as e:  # type: ignore
+                # catch any exceptions, log them, and keep on going
+                for exception in e.exceptions:  # type: ignore
+                    error_msg = str(exception)  # type: ignore
+                    logger.error(error_msg, vectorizer_id=vectorizer_id)
+                for exception_line in traceback.format_exception(e):  # type: ignore
```
cevian, line 269: should we do this for all exceptions or only the group?

cevian (approved):

projects/pgai/pgai/vectorizer/processor.py
cevian: oh nvm I see the other comment

JamesGuthrie (approved):

PR #598, opened 2025-04-01
fix: exclude manually added columns from copy

Askir:


projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -979,6 +979,7 @@ async def _load_copy_types(self, conn: AsyncConnection) -> None:
                 and k.relname operator(pg_catalog.=) %s
                 and a.attname operator(pg_catalog.!=) 'embedding_uuid'
                 and a.attnum operator(pg_catalog.>) 0
+                and a.attnotnull = true
```
jgpruitt, line 982: We should use LIMIT rather than `attnotnull`. A person could add a `not null` column with a default value, and it should work just fine. So, grab the first N columns ordered by `attnum`.

jgpruitt:
> I think we should change this to look up types by column names, not just rely on the table prefix by attnum. That would be much more robust

Agreed. But if we do this, we must construct the `copy` command to allow for the arbitrary ordering of columns too. Maybe it almost handles that already. I'm not sure how you could "inject" a new column into the middle without some major finnagling.

Askir:
@cevian I updated the query. I don't really know how to write a test for this though, since you cant just add a column at any point in postgres, it's always added at the end.

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -979,11 +984,14 @@ async def _load_copy_types(self, conn: AsyncConnection) -> None:
...
-                order by a.attnum
             """,
-                (self.vectorizer.target_schema, self.vectorizer.target_table),
+                (
+                    self.vectorizer.target_schema,
+                    self.vectorizer.target_table,
+                    target_columns,
+                ),
             )
             self.copy_types = [row[0] for row in await cursor.fetchall()]
```
cevian, line 996: there is no guarantee right now of the order of the results. The query should return the column name, type and then we can finangle things into the right order

cevian (changes requested):

cevian (approved):

PR #601, opened 2025-04-02
fix: trigger def migration

cevian:
Previously trigger def migration was in the idempotent files, and thus would run with every migration, while we actually just want it to run once.

Moving it to the incremental file does that.


projects/pgai/db/sql/incremental/026-change_trigger_definition.sql
jgpruitt, line 1: The duplication of `ai._vectorizer_build_trigger_definition` isn't great, but not a dealbreaker. We kinda need the idea of `pre` and `post` incremental files (those that run before idempotent, and those that run after).

jgpruitt (approved):

Askir:
Yeh i did this intentionally to avoid the duplication of the trigger definition. I wrote the migration in an idempotent way. And also need the same setup again for the same table vectorizer (because all the triggers need to be recreated to avoid triggering on the embedding column).

I don't think recreating all triggers on every install is that bad? At least that ensures all the trigger functions are up to date and you don't need to write these one off migrations for it.

cevian:
@Askir recreating triggers takes very heavy locks on the underlying tables. It is not best practice to do that on every upgrade if it can be avoided. We could in the idempotent script create the code for the trigger, compare that to the existing code, and only recreate the trigger if there is a diff but I think it's more straightforward to just do an incremental upgrade, wdyt.

cevian:
@Askir great idea! changed pr to implement

Askir (approved):
lgtm!

PR #606, opened 2025-04-07
perf: maximaly defer library imports

JamesGuthrie:
A number of third-party libraries impose significant import-time overhead. Many are unlikely to be used in all vectorizer configurations. This change, driven by profiling data, pushes those imports as far down as possible.

This refactoring significantly reduces the CPU and time taken to import the `Processor` module from `pgai.vectorizer`.

The output of `time python -c "from pgai.vectorizer import Processor"`:

Before:

real    0m5.078s
user    0m7.520s
sys     0m0.376s

After:

real    0m0.325s
user    0m1.751s
sys     0m0.030s


smoya (approved):
Great job 💯 💯 Just out of curiosity, what are the computer specs you used for running the profiling?

JamesGuthrie:
> Great job 💯 💯 Just out of curiosity, what are the computer specs you used for running the profiling?

It's a desktop with an Intel i5-13600K CPU.

alejandrodnm (approved):

PR #608, opened 2025-04-08
feat: add proc processor

cevian:
The proc processor runs the processor in a separate process. This is useful for running a processor as a "sidecar". For example, you may want your FastAPI app to start a processor alongside the main app.


projects/pgai/pgai/vectorizer/processor.py
```diff
@@ -266,4 +280,152 @@ async def run(self) -> Exception | None:
...
+    def run_in_new_process(self):
+        # todo: monitor for process termination
+        self.result_queue: ProcProcessor.QueueType[ProcProcessor.MessageType] = (
+            mp.Queue()
+        )
+        self.process = mp.Process(
+            target=self._run_processor_inside_new_process,
+            args=(self, self.result_queue),
+        )
+        self.process.start()
```
Askir, line 367: Is that the correct way to let it run alongside the fastapi process? I would have thought we try to hook our existing worker into the fastapi eventloop instead of starting another process?
I feel like if you have two processes you probably should have two containers for it?

JamesGuthrie: TBH I would be slightly concerned with putting the vectorizer in fastapi's event loop. While we are using async, I don't think that we can know that we're (or our dependencies) not accidentally doing blocking calls somewhere.

I've looked for other examples of doing "background work" in FastAPI and found two:
- APScheduler: https://pypi.org/project/APScheduler/
- FastAPI Utils "Repeated Tasks": https://fastapi-utils.davidmontague.xyz/user-guide/repeated-tasks/

APScheduler offers both asyncio integration, and a thread-based background worker. FastAPI Utils also does both, from what I can tell, it runs async functions in the event loop, and synchronous functions in a thread pool.

projects/pgai/pgai/vectorizer/processor.py
```diff
@@ -266,4 +280,156 @@ async def run(self) -> Exception | None:
...
+                    # timeout means the sleep completed
+                    pass
+
+        if worker_tracking is not None:
+            await worker_tracking.force_last_heartbeat_and_stop()
+        logger.info("exiting processor.run()")
+        return None
+
+
+class ProcProcessor(Processor):
```
JamesGuthrie, line 304: Is there any reason why we subclass Processor? Could we not use composition here?

PR #614, opened 2025-04-10
perf: reduce vectorizer memory consumption

alejandrodnm:
### perf: optimize batches memory usage 
 
 - Convert embedding processing from list-based to generator-based approach
- Replace batch collections with async generators for incremental processing
- Process embeddings in chunks rather than keeping all in memory

These changes reduce memory consumption by avoiding the accumulation of
all embeddings in memory at once, instead processing them incrementally
as they become available.

### perf: reduce memory consumption by lazy loading openAI's response 

https://www.loom.com/share/a25ac5bf5d56409581d4c9da1517c257?sid=a2fdff86-3b52-4aad-b116-ede34c122084


alejandrodnm:
Flamegraph before improvements:

https://drive.google.com/file/d/1-GNrDkGbaFGCLzRUb40Wkoqk7EoQM_fC/view?usp=drive_link

Flamegraph after improvements:

https://drive.google.com/file/d/1QWZq2MaSWM8tdMkOmzHVAExjM-FAvjVj/view?usp=drive_link

projects/pgai/pgai/vectorizer/embedders/openai.py
```diff
@@ -35,6 +36,18 @@
 )
 
 
+class ResponseWithRead:
+    def __init__(self, response: "AsyncAPIResponse[types.CreateEmbeddingResponse]"):
+        default_ijson_buffer = 64 * 1024
```
JamesGuthrie, line 41: Can we extract this to a constant and also pass that value as the `buf_size` parameter to ijson.async_parse? I'm concerned that we're implicitly relying on them being the same.

JamesGuthrie (approved):
LGTM overall, one minor comment.

I was curious about how much of a difference using `ijson` alone makes, so I benchmarked it standalone, and the difference was surprising!

Memory profile before ijson:

Screenshot 2025-04-11 at 09 39 04

Memory profile after:

Screenshot 2025-04-11 at 09 38 27

Note: I was not using the pre-recorded responses for this test, so the timing differences are expected.

Peak Heap size dropped from ~450MB to ~250MB!

projects/pgai/pgai/vectorizer/embedders/openai.py
alejandrodnm: Good one 👍🏼

JamesGuthrie (approved):

PR #616, opened 2025-04-13
feat: add graceful shutdown to processor

cevian:
Usage now becomes e.g.

```
async def lifespan(_app: FastAPI):
    # start the processor in a new process running in the background
    processor = Processor(DB_URL)
    task = asyncio.create_task(processor.run())

   yield
    
    print("Shutting down...")
    
    print("gracefully shutting down processor...")
    await processor.request_graceful_shutdown()
    try:
        result = await asyncio.wait_for(task, timeout=20)
        if result is not None:
            print("Processor shutdown with exception:", result)
        else:
            print("Processor shutdown successfully")
    except asyncio.TimeoutError:
        print("Processor did not shutdown in time, killing it")
    
    print("Shutting down complete")

```

### DIFF
--- a/projects/pgai/pgai/vectorizer/processor.py
+++ b/projects/pgai/pgai/vectorizer/processor.py
@@ -56,6 +56,7 @@ class Processor:
         self.once = once
         self.exit_on_error = exit_on_error
         self.concurrency = concurrency
+        self.shutdown_requested = asyncio.Event()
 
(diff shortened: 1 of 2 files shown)


Askir (approved):
I like this implementation, you could still run it in a separate process/thread if you prefer I think an use the asyncio event to synchronize the shutdown.

JamesGuthrie (approved):

PR #622, opened 2025-04-16
feat: named vectorizers

Askir:


projects/pgai/db/sql/incremental/029-add-vectorizer-name-column.sql
```diff
@@ -0,0 +1,45 @@
+alter table ai.vectorizer add column name text;
```
jgpruitt, line 1:
```suggestion
alter table ai.vectorizer add column name name check (catalog_name ~ '^[a-z][a-z_0-9]*$');
```

Consider restricting the name to something that would make a reasonable postgres identifier that doesn't require double-quoting. There's a decent chance we may want to create other things in the future that include this name in the object name. It will also make the name "fit in" like other postgres objects.

jgpruitt: I am torn on whether "name" is a good column name since it's also a built-in type, but we've done it before, so maybe okay.

jgpruitt, line 1:
```suggestion
alter table ai.vectorizer add column name name check (catalog_name ~ '^[a-z][a-z_0-9]*$');
```

Consider restricting the name to something that would make a reasonable postgres identifier that doesn't require double-quoting. There's a decent chance we may want to create other things in the future that include this name in the object name. It will also make the name "fit in" like other postgres objects.

I'm also torn on whether "name" is a good column name since "name" is a built-in type. I'd like to use it, since it's easy and succinct. We've done it before, so maybe it's ok.

jgpruitt (changes requested):

projects/pgai/db/sql/idempotent/012-vectorizer-api.sql
```diff
@@ -145,6 +148,29 @@ begin
...
+
+    -- validate the name is available
+    select id from ai.vectorizer
+    where ai.vectorizer.name operator(pg_catalog.=) create_vectorizer.name
+    into _existing_vectorizer_id
+    ;
+    if _existing_vectorizer_id is not null then
+        if if_not_exists is false then
+            raise exception 'a vectorizer named % already exists.', name;
+        end if;
```
cevian, line 170:
```suggestion
        end if;
        raise notice `vectorizer % already exists`, name 
```

projects/pgai/db/sql/idempotent/010-destination.sql
```diff
@@ -64,6 +64,65 @@ begin
...
+        end;
+        view_schema = coalesce(view_schema, source_schema);
+        view_name = case
+            when destination operator(pg_catalog.->>) 'view_name' is not null then destination operator(pg_catalog.->>) 'view_name'
+            when destination operator(pg_catalog.->>) 'destination' is not null then destination operator(pg_catalog.->>) 'destination'
+            else pg_catalog.concat(source_table, '_embedding')
+        end;
+         -- make sure view name is available
+        if pg_catalog.to_regclass(pg_catalog.format('%I.%I', view_schema, view_name)) is not null then
+            raise exception 'an object named %.% already exists. specify an alternate destination explicitly', view_schema, view_name;
```
cevian, line 101: won't this break `if_not_exists`?

projects/pgai/db/sql/idempotent/010-destination.sql
```diff
@@ -64,6 +64,65 @@ begin
...
+        end if;
+    
+        -- make sure target table name is available
+        if pg_catalog.to_regclass(pg_catalog.format('%I.%I', target_schema, target_table)) is not null then
+            raise exception 'an object named %.% already exists. specify an alternate destination or target_table explicitly', target_schema, target_table;
```
cevian, line 106: same here

projects/pgai/db/sql/idempotent/012-vectorizer-api.sql
```diff
@@ -280,19 +308,34 @@ set search_path to pg_catalog, pg_temp
 
 -------------------------------------------------------------------------------
 -- disable_vectorizer_schedule
-create or replace function ai.disable_vectorizer_schedule(vectorizer_id pg_catalog.int4) returns void
+create or replace function ai.disable_vectorizer_schedule
```
cevian, line 311: so there are two approaches:
1) ai.disable_vectorizer_schedule(vectorizer_id int4, name text)
2) ai.disable_vectorizer_schedule(vectorizer_id int4) ai.disable_vectorizer_schedule(name text) (overwridding)

The advantage of 2 is that 
ai.disable_vectorizer_schedule('my_name') would work. Right now you'd have to do ai.disable_vectorizer_schedule(name=>'my_name'). 

so I think I'd prefer 2. WDYT?

projects/pgai/db/tests/vectorizer/test_named_vectorizer.py
```diff
@@ -0,0 +1,199 @@
...
+            , embedding=>ai.embedding_openai('text-embedding-3-small', 768)
+            , formatting=>ai.formatting_python_template('title: $title published: $published $chunk')
+            , scheduling=>ai.scheduling_timescaledb
+                    ( interval '5m'
+                    , initial_start=>'2050-01-06'::timestamptz
+                    , timezone=>'America/Chicago'
+                    )
+            , destination=>ai.destination_column('embedding1')
+            , chunking=>ai.chunking_none()
+            , if_not_exists => true
```
cevian, line 98: we need to test `if_not_exists=> true` with a destination_table as I think that's currently broken

cevian (changes requested):

projects/pgai/db/sql/idempotent/012-vectorizer-api.sql
jgpruitt: I prefer the override approach (2) as well. You're either using the id OR the name to specify, never both.
What would happen if someone specified the id from one vectorizer and the name of a different one?

projects/pgai/pgai/alembic/operations.py
```diff
@@ -22,18 +22,20 @@ def create_vectorizer(cls, operations: Operations, **kw: Any):
 
 
 class DropVectorizerOp(MigrateOperation):
-    def __init__(self, vectorizer_id: int | None, drop_all: bool):
+    def __init__(self, vectorizer_id: int | None, name: str | None, drop_all: bool):
```
smoya, line 25: Nit: for consistency
```suggestion
    def __init__(self, vectorizer_id: int | None, vectorizer_name: str | None, drop_all: bool):
```
Or rather change the `vectorizer_id` to just `id`.
Same applies to the rest of the code

projects/pgai/db/sql/idempotent/010-destination.sql
```diff
@@ -64,6 +64,79 @@ begin
...
     end if;
 end
+$func$ language plpgsql immutable security invoker
+set search_path to pg_catalog, pg_temp
+;
+
+
+-------------------------------------------------------------------------------
+-- evaluate_destination
+create or replace function ai._evaluate_destination
```
smoya, line 74: It's nice we have all this logic encapsulated now 💯

smoya (approved):
LGTM at a glance.

projects/pgai/db/sql/incremental/029-add-vectorizer-name-column.sql
cevian: added

PR #629, opened 2025-04-17
fix: send top level destination fields in updateembeddings request

Askir:
I tried to also create a test for this, but since the function is in the bundled ai library I have a hard time mocking the httpx call. I could call the function outside of the postgres context but then I need to mock the plpy call at which point... I'm not testing much anymore. Bit annoying, I guess we'll just live with it?

### DIFF
--- a/projects/extension/ai/vectorizer.py
+++ b/projects/extension/ai/vectorizer.py
@@ -32,6 +32,20 @@ def execute_vectorizer(plpy, vectorizer_id: int) -> None:
     embedding_api_key = (
         vectorizer.get("config", {}).get("embedding", {}).get("api_key_name", None)
     )
+    destination = vectorizer.get("config", {}).get("destination", None)
+    if destination is not None:
+        vectorizer["target_schema"] = destination.get(
+            "target_schema", vectorizer["source_schema"]
+        )
+        vectorizer["target_table"] = destination.get(
+            "target_table", vectorizer["source_table"]
+        )
+        vectorizer["view_schema"] = destination.get(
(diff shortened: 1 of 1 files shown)


projects/extension/ai/vectorizer.py
```diff
@@ -32,6 +32,12 @@ def execute_vectorizer(plpy, vectorizer_id: int) -> None:
     embedding_api_key = (
         vectorizer.get("config", {}).get("embedding", {}).get("api_key_name", None)
     )
+    destination = vectorizer.get("config", {}).get("destination", None)
+    if destination is not None:
+        vectorizer["target_schema"] = destination.get("target_schema", vectorizer.get("source_schema"))
+        vectorizer["target_table"] = destination.get("target_table", vectorizer.get("source_table"))
+        vectorizer["view_schema"] = destination.get("view_schema", vectorizer.get("source_schema"))
+        vectorizer["view_table"] = destination.get("view_table", vectorizer.get("source_table"))
```
JamesGuthrie, line 40: Why the fallback to `vectorizer.get`? If destination is not `None`, then we wouldn't expect those values to be populated?

JamesGuthrie, line 40:
```suggestion
        vectorizer["view_name"] = destination.get("view_name", vectorizer.get("source_table"))
```

Askir: yeh true makes no sense, null would break the receiving end anyways since not empty is asserted.

cevian (approved):

PR #630, opened 2025-04-17
fix: vectorizer_status with column dest

cevian:


projects/pgai/pgai/data/ai.sql
```diff
@@ -3871,10 +3871,16 @@ create or replace view ai.vectorizer_status as
 select
   v.id
 , pg_catalog.format('%I.%I', v.source_schema, v.source_table) as source_table
-, pg_catalog.format('%I.%I', v.config operator(pg_catalog.->) 'destination' operator(pg_catalog.->>) 'target_schema', v.config operator(pg_catalog.->) 'destination' operator(pg_catalog.->>) 'target_table') as target_table
-, pg_catalog.format('%I.%I', v.config operator(pg_catalog.->) 'destination' operator(pg_catalog.->>) 'view_schema', v.config operator(pg_catalog.->) 'destination' operator(pg_catalog.->>) 'view_name') as "view"
+, case when v.config operator(pg_catalog.->) 'destination' operator(pg_catalog.->>) 'implementation' = 'table' then
+    pg_catalog.format('%I.%I', v.config operator(pg_catalog.->) 'destination' operator(pg_catalog.->>) 'target_schema', v.config operator(pg_catalog.->) 'destination' operator(pg_catalog.->>) 'target_table')
+    else null
```
smoya, line 3876: Even though in the first pass i was expecting to be explicit here and show the name of the actual source table instead of null, I actually believe it looks way better when showing `null` instead.

smoya (approved):
LGTM! 🚀🌔

PR #631, opened 2025-04-18
fix: rendering cagg views in semantic catalog

MasterOdin:
PR fixes rendering continuous aggregate views, where previously the views would be rendered as:

```
<view>
CREATE VIEW postgres_air.events_daily AS
 SELECT name,
    bucket
   FROM _timescaledb_internal._materialized_hypertable_2;
/* column types
name text
bucket timestamp without time zone
*/
</view>
```

and now they'll be properly rendered as:

```
<view>
CREATE MATERIALIZED VIEW postgres_air.events_daily WITH (timescaledb.continuous) AS
 SELECT name,
    postgres_air.time_bucket('1 day'::interval, "time") AS bucket
   FROM postgres_air.events
  GROUP BY name, (postgres_air.time_bucket('1 day'::interval, "time"));
/* column types
name text
bucket timestamp without time zone
*/
</view>
```

which matches how it was defined.


projects/pgai/pgai/semantic_catalog/loader.py
```diff
@@ -96,15 +96,28 @@ async def load_views(con: psycopg.AsyncConnection, oids: list[int]) -> list[View
     # TODO: add support for continuous aggregates
     assert len(oids) > 0, "list of oids must not be empty"
     async with con.cursor(row_factory=dict_row) as cur:
+        await cur.execute("select 1 from pg_extension where extname = 'timescaledb'")
+        is_timescale = await cur.fetchone() is not None
```
MasterOdin, line 100: I debated caching this boolean value within the module as we'll probably need a similar value for `load_tables` (and others?), but ultimately decided against it for the time being given the async nature of the code, and that doing this query is largely negligible in overall runtime.

projects/pgai/tests/semantic_catalog/data/cagg.sql
```diff
@@ -0,0 +1,18 @@
+create extension if not exists timescaledb with schema postgres_air cascade;
+create extension if not exists timescaledb_toolkit with schema postgres_air cascade;
+
+create table if not exists postgres_air.events
```
MasterOdin, line 4: I had originally looked to create a hypertable on one of the existing tables, but I think that would require modifying their indexes, and so ultimately just created this placeholder table.

projects/pgai/tests/semantic_catalog/conftest.py
```diff
@@ -22,6 +22,17 @@ def load_postgres_air(container: PostgresContainer):
...
+    with (
+        psycopg.connect(
+            container.connection_string(database="postgres_air"),
+            autocommit=True,
+        ) as con,
+        con.cursor() as cur,
+    ):
+        for query in script3.split(";"):
+            if query.strip():
+                cur.execute(query)  # pyright: ignore [reportArgumentType]
```
MasterOdin, line 35: I was getting errors about cannot create materialized view within a transaction errors when I tried:
1. Sticking the execute in the above with block
2. Executing all the queries in one `cur.execute`.

Hence, I stuck this into its own context block (with `autocommit=True`) and had to do the `split(";")` so that it all worked.

projects/pgai/pgai/semantic_catalog/loader.py
jgpruitt: Yeah, i think cutting corners is fine for now. We can come back later and squeeze every ounce of performance out if needed.

projects/pgai/tests/semantic_catalog/conftest.py
jgpruitt: no problem

projects/pgai/tests/semantic_catalog/data/cagg.sql
```diff
@@ -0,0 +1,18 @@
+create extension if not exists timescaledb with schema postgres_air cascade;
+create extension if not exists timescaledb_toolkit with schema postgres_air cascade;
```
jgpruitt, line 2: Out of curiosity, why did you install them in the postgres_air schema?

projects/pgai/tests/semantic_catalog/data/cagg.sql
jgpruitt: this is good

jgpruitt (approved):

projects/pgai/pgai/semantic_catalog/builder.py
```diff
@@ -42,6 +42,9 @@ async def find_tables(
                 on (c.relnamespace = n.oid)
             where n.nspname not like 'pg_%%'
             and n.nspname != 'information_schema'
+            and n.nspname not like '_timescaledb_%%'
+            and n.nspname not like 'timescaledb_%%'
+            and n.nspname != 'toolkit_experimental'
```
MasterOdin, line 47: The full set of things to ignore I found was:

```
            and n.nspname != 'timescaledb_experimental'
            and n.nspname != 'timescaledb_information'
            and n.nspname != 'toolkit_experimental'
            and n.nspname != '_timescaledb_cache'
            and n.nspname != '_timescaledb_catalog'
            and n.nspname != '_timescaledb_config'
            and n.nspname != '_timescaledb_debug'
            and n.nspname != '_timescaledb_functions'
            and n.nspname != '_timescaledb_internal'
```

debated on using `not like` vs `not in` and explicitly listing everything, where the former feels more "forward facing" but could potentially disregard user created schemas that match that, but decided that it seemed unlikely end-users would create their own schemas with those prefixes.

MasterOdin:
@jgpruitt Made some changes to builder after you reviewed the PR, so re-requesting it.

projects/pgai/tests/semantic_catalog/test_builder.py
```diff
@@ -13,16 +13,778 @@
...
     "passenger",
     "phone",
 }
 
-ALL_VIEWS: set[str] = {"flight_summary", "passenger_details"}
+ALL_VIEWS: set[str] = {"events_daily", "flight_summary", "passenger_details"}
 
 ALL_PROCS: set[str] = {"advance_air_time", "update_flight_status"}
 
+TIMESCALE_PROCS: set[str] = {
```
jgpruitt, line 27: I think this is fine for now, but we may want to revisit so we aren't chasing changes to extensions that install stuff in the public schema. Maybe we should create an "empty" schema to use instead of public for these tests.

projects/pgai/pgai/semantic_catalog/builder.py
jgpruitt: Yeah, I agree with `!=` and excluding these for now. We may want to make it possible for users to describe these in the future. For now, let's move on.

jgpruitt (approved):

PR #632, opened 2025-04-18
quickstart rework py script

cevian:


avthars (changes requested):
Overall I like the minimal version, but left comments that could apply to both.

examples/simple_fastapi_app/README.md
```diff
@@ -0,0 +1,328 @@
...
+    ```python
+    worker = Worker(DB_URL)
+    task = asyncio.create_task(worker.run())
+    ```
+    
+    In this example, we run the Vectorizer worker inside the FastAPI app for simplicity. You can also run the vectorizer worker outside the FastAPI app, in a separate process or separate container. Please see the [vectorizer worker](/docs/vectorizer/worker.md) documentation for more information.
+    
+1. **Create the table and load the test dataset**
+
+    We'll create a table named `wiki` from a few rows of the english-language `wikimedia/wikipedia` dataset.
```
smoya, line 57: I notice we are mixing future and present tense. I think we should use present one always here.
```suggestion
    We create a table named `wiki` from a few rows of the english-language `wikimedia/wikipedia` dataset.
```

This applies to some other parts of this walkthrough

examples/simple_fastapi_app/README.md
```diff
@@ -0,0 +1,328 @@
...
+                    LIMIT %s
+                """, (embedding, limit))
+                
+                return await cur.fetchall()
+    ```
+    </details>
+    
+    This query selects from the `wiki_embedding` view that was created by the vectorizer and which contains all the columns from the `wiki` table plus the `embedding` column which contains the vector embeddings and the `chunk` column which contains the chunked text corresponding to the embedding. In this query, we are returning both the full text of the article and the chunk that is most relevant to the query (different applications may want to return only one or the other).
+
+     The `embedding <=> %s` is a PostgreSQL operator that computes the cosine distance between the stored embedding and the query embedding.
```
smoya, line 211: Worth to mention (kinda of again) this operator is available through pgvector?

README.md
```diff
@@ -39,351 +39,358 @@ Works with any PostgreSQL database, including Timescale Cloud, Amazon RDS, Supab
...
-1. **Start the docker compose file.**
-    ```
-    docker compose up -d
-    ```
+run the following code in the same directory as the `.env` file:
+
+<details>
+<summary>Click to expand code</summary>
+
+```python
```
smoya, line 86: Wouldn't make sense to move all this python code into an actual .py file so people can download it (curl) ?

README.md
```diff
@@ -39,351 +39,358 @@ Works with any PostgreSQL database, including Timescale Cloud, Amazon RDS, Supab
...
+```
+TODO
+```
+</details>
 
-### Create a FastAPI app that performs semantic search and RAG over wikipedia articles
 
-You can take a look at a simple [fastAPI Application](/examples/simple_fastapi_app/with_psycopg.py) to see how to
-setup an app to perform RAG with pgai Vectorizer. 
+## Key parts of the code
```
smoya, line 250: As in https://github.com/timescale/pgai/pull/632/files
```suggestion
## Code walkthrough
```

smoya (commented):
It looks pretty clear with the recent changes. 
On top of what others have said, I want to bring my concern about the abuse of duplication of code.
We are writing scripts and then copying snippets of them in order to show the walkthroughs and IMHO that adds a lot of maintanance burden, specially whenever we need to update those source scripts.
Maybe we should consider using some tools in our CI like https://github.com/zakhenry/embedme so we just embed the snippets from actual source files.
It could be done as a next step, not necessarly in this PR of course.

avthars (approved):
Awesome stuff, love the new flow

PR #635, opened 2025-04-21
chore: remove structlog

cevian:
It is considered best practice for Python libraries to use the standard logger package.


projects/pgai/pgai/logger.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Returns:
+        A Logger instance with the appropriate namespace
+    """
+    if name:
+        logger_name: str = f"pgai.{name}"
+    else:
+        logger_name: str = "pgai"
+
+    return logging.getLogger(logger_name)
```
JamesGuthrie, line 53: I don't understand how this is any different to "just" using `logging.getLogger(__name__)`?

projects/pgai/pgai/logger.py
```diff
@@ -0,0 +1,71 @@
...
+    Args:
+        level: The logging level (e.g., logging.INFO, logging.DEBUG)
+              or a string level name ('INFO', 'DEBUG', etc.)
+    """
+    if isinstance(level, str):
+        numeric_level: int = getattr(logging, level.upper(), logging.INFO)
+    else:
+        numeric_level = level
+
+    logging.getLogger("pgai").setLevel(numeric_level)
```
JamesGuthrie, line 71: I'm not sure I understand the utility of this. The caller of this could just as well use `logging.getLogger("pgai").setLevel()` themselves, which is the standard way of doing this in python logging.

projects/pgai/pgai/logger.py
```diff
@@ -0,0 +1,71 @@
...
+        self.kwargs: dict[str, Any] = kwargs
+
+    @override
+    def __str__(self) -> str:
+        renderer: RendererType = self._renderer or self.default_renderer
+        return renderer(self.message, self.kwargs)
+
+    @staticmethod
+    def default_renderer(msg: str, kwargs: dict[str, Any]) -> str:
+        return f"{msg} >>> {json.dumps(kwargs)}"
```
JamesGuthrie, line 28: This fails when the vectorizer worker runs, because json.dumps can't serialize UUID.

projects/pgai/pgai/logger.py
```diff
@@ -0,0 +1,71 @@
...
+
+from typing_extensions import override
+
+T = TypeVar("T", bound="StructuredMessage")
+
+# Type for the renderer function
+RendererType = Callable[[str, dict[str, Any]], str]
+
+
+class StructuredMessage:
```
JamesGuthrie, line 14: I don't understand why we're introducing the `StructuredMessage` class. I am aware that we had structured logging with structlog, but I don't understand if we're just trying to replicate what structlog had, or provide some functionality for our users.

projects/pgai/pgai/cli.py
JamesGuthrie, line 1: This change breaks logging for the install subcommand (no logs appear).

JamesGuthrie (changes requested):

projects/pgai/pgai/logger.py
cevian: it enforces the `pgai` prefix

projects/pgai/pgai/logger.py
cevian: This is really mostly for internal usage to be able to keep some of the structured data we already specified in the output

projects/pgai/pgai/logger.py
JamesGuthrie: The prefix is always `pgai` (except in `__main__.py`), because all files are in the pgai module.

PR #638, opened 2025-04-22
feat: add a way to sample tables and render to inserts

jgpruitt:
Adds a way to sample tables and views.
Samples N rows from a table/view and render as either inserts or as a copy command with text format.


projects/pgai/pgai/semantic_catalog/sample.py
```diff
@@ -0,0 +1,87 @@
...
+import psycopg
+from psycopg.sql import SQL, Composed, Identifier, Literal
+
+
+async def _sample_as_inserts(
+    con: psycopg.AsyncConnection, schema_name: str, object_name: str, limit: int = 3
+) -> str:
+    async with con.cursor() as cur:
+        await cur.execute(
+            SQL("select * from {}.{} limit %s").format(
```
cevian, line 12: shouldn't we use TABLESAMPLE here (if table not view)? Or do you think overkill for now?

jgpruitt: TABLESAMPLE would be good if we wanted more than a few rows and it would have the advantage of not just sampling the "head" of the table, but I think it's overkill for the moment. @MasterOdin found some literature that indicated 3 sample rows are plenty.

projects/pgai/pgai/semantic_catalog/templates/table.j2
```diff
@@ -12,4 +12,5 @@ ALTER TABLE {{ table.schema_name }}.{{ table.table_name }} ADD CONSTRAINT {{ con
 {% for column in (table.columns if table.columns is not none else []) -%}
 {% if column.description is not none %}COMMENT ON COLUMN {{ table.schema_name }}.{{ table.table_name }}{{ column.name }} IS $${{ column.description.description }}$$;{% endif -%}
 {% endfor -%}
+{% if table.sample is not none %}{{ table.sample }}{% endif -%}
```
cevian, line 15: should we have <samples> or anything here?

jgpruitt: Not sure I follow. What would we put if the sample is missing?

cevian: sorry should we have `<samples></samples>` here

projects/pgai/pgai/semantic_catalog/sample.py
MasterOdin: Yeah, it's not been clear to me whether `TABLESAMPLE` would give more meaningful results vs just grabbing the first couple of rows. Also yeah, for the generic case, I've found that only a few rows provide meaningful info to the LLM without blowing up its context window, except for very specialized cases, but generally in that case using the semantic catalog is probably better vs relying on sampling.

MasterOdin (approved):

PR #639, opened 2025-04-22
fix: turn off ddtrace by default

cevian:


jgpruitt (approved):

Askir (approved):
Are we manually enabling this for the lambda then?
edit: I checked the docker file and we do already set that env var 👍

PR #641, opened 2025-04-23
docs: update for name and destination changes

Askir:


docs/vectorizer-quick-start.md
```diff
@@ -90,8 +90,8 @@ Now we can create and run a vectorizer. A vectorizer is a pgai concept, it proce
     SELECT ai.create_vectorizer(
          'blog'::regclass,
          loading => ai.loading_column('contents'),
-         destination => 'blog_contents_embeddings',
-         embedding => ai.embedding_ollama('nomic-embed-text', 768),
+         destination => ai.destination('blog_contents_embeddings'),
```
cevian:
```suggestion
         destination => ai.destination_table('blog_contents_embeddings'),
```

cevian: also nit: but I'd put that as the last element to simulate the pipeline (it's the last output element)

docs/vectorizer/api-reference.md
```diff
@@ -111,14 +121,40 @@ SELECT ai.create_vectorizer(
...
+3. Creates a view `website.blog_embeddings` joining the source and embeddings.
+4. Loads the `contents` column.
+5. Uses the Ollama `nomic-embed-text` model to create 768 dimensional embeddings.
+6. Chunks the content into 128-character pieces with a 10-character overlap.
+7. Formats each chunk with a `title` and a `published` date.
+8. Grants necessary permissions to the roles `bob` and `alice`.
+
+#### Example 2: Column destination
+
+For simpler cases where you need just one embedding per row (e.g. because your application already did the chunking), you can add a column directly to the source table:
```
cevian, line 135:
```suggestion
Column destination place the embedding in a separate column in the source table. It can only be used when the vectorizer does not perform chunking because it requires a one-to-one relationship between the source data and the embedding. This is useful in cases where you know the source text is short (as is common if the chunking has already been done upstream in your data pipeline).

The workflow is that your application inserts data into the table with a NULL in the embedding column. The vectorizer will then read the row, generate the embedding and update the row with the correct value in the embedding column.
```

docs/vectorizer/api-reference.md
```diff
@@ -111,14 +121,40 @@ SELECT ai.create_vectorizer(
...
+);
+```
+
+This function call:
+1. Sets up a vectorizer named 'product_descriptions_vectorizer' for the `website.product_descriptions` table.
+2. Adds a column called `description_embedding` directly to the source table.
+3. Loads the `description` column.
+4. Uses OpenAI's embedding model to create 768 dimensional embeddings.
+5. Doesn't chunk the content (required for column destination).
+6. Grants necessary permissions to the role `marketing_team`.
```
cevian, line 155:
```suggestion
3. Loads the `description` column.
4. Doesn't chunk the content (required for column destination).
5. Uses OpenAI's embedding model to create 768 dimensional embeddings.
6. Writes the embedding into the `description_embedding` column by updating the row.
7. Grants necessary permissions to the role `marketing_team`.
```

docs/vectorizer/api-reference.md
```diff
@@ -128,7 +164,8 @@ in other management functions.
...
 | source           | regclass                                               | -                                 | ✔        | The source table that embeddings are generated for.                                                |
-| destination      | name                                                   | -                                 | ✖        | Set the name of the table embeddings are stored in, and the view with both the original data and the embeddings.<br>The view is named `<destination>`, the embedding table is named `<destination>_store`.<br>You set destination to avoid naming conflicts when you configure additional vectorizers for a source table.                              |
+| name             | text                                                   | Auto-generated                    | ✖        | A unique name for the vectorizer. If not provided, it's auto-generated based on the destination type:<br>- For table destination: `[target_schema]_[target_table]`<br>- For column destination: `[source_schema]_[source_table]_[embedding_column]`<br>Must follow snake_case pattern `^[a-z][a-z_0-9]*$`                              |
```
cevian, line 167: don't we need to add the `if_not_exists` parameter here as well?

docs/vectorizer/api-reference.md
```diff
@@ -149,7 +182,93 @@ in other management functions.
...
+
+A JSON configuration object that you can use in [ai.create_vectorizer](#create-vectorizers).
+
+### ai.destination_column
+
+You use `ai.destination_column` to store embeddings directly in the source table as a new column. This approach:
+- Adds a vector column directly to the source table
+- Does not create a separate view
+- Requires chunking to be set to `ai.chunking_none()` (no chunking)
+- Stores exactly one embedding per row
```
cevian, line 248:
```suggestion
You use `ai.destination_column` to store embeddings directly in the source table as a new column. This approach can only be used when the vectorizer does not perform chunking because it requires a one-to-one relationship between the source data and the embedding. This is useful in cases where you know the source text is short (as is common if the chunking has already been done upstream in your data pipeline).

This approach:
- Adds a vector column directly to the source table
- Does not create a separate view
- Requires chunking to be set to `ai.chunking_none()` (no chunking)
- Stores exactly one embedding per row

The workflow is that your application inserts data into the table with a NULL in the embedding column. The vectorizer will then read the row, generate the embedding and update the row with the correct value in the embedding column.
```

docs/vectorizer-quick-start.md
smoya, line 93: Isn't `destination_column` the name of the function 🤔 ?
```suggestion
         destination => ai.destination_column('blog_contents_embeddings'),
```

docs/vectorizer/api-reference.md
```diff
@@ -1384,23 +1516,30 @@ You use `ai.disable_vectorizer_schedule` to:
...
 SELECT ai.disable_vectorizer_schedule(1);
+
+-- Using name (recommended)
+SELECT ai.disable_vectorizer_schedule('public_blog_embeddings');
 ```
 
 #### Parameters
 
-`ai.disable_vectorizer_schedule` takes the following parameters:
+`ai.disable_vectorizer_schedule` takes one of the following parameters:
```
cevian, line 1531: this is misleading. There are now two functions one that takes an id and one that takes a name. Note one that takes an id or a name.

docs/vectorizer/api-reference.md
```diff
@@ -1357,23 +1482,30 @@ You use `ai.enable_vectorizer_schedule` to:
...
 SELECT ai.enable_vectorizer_schedule(1);
+
+-- Using name (recommended)
+SELECT ai.enable_vectorizer_schedule('public_blog_embeddings');
 ```
 
 #### Parameters
 
-`ai.enable_vectorizer_schedule` takes the following parameters:
+`ai.enable_vectorizer_schedule` takes one of the following parameters:
```
cevian, line 1497: misleading... there are two different functions

docs/vectorizer/api-reference.md
```diff
@@ -1466,9 +1609,12 @@ Examples:
 
 |Name| Type | Default | Required | Description |
 |-|------|-|-|-|
-|vectorizer_id| int  | -|✔|The identifier of the vectorizer you want to drop|
+|vectorizer_id| int  | -|✔*|The identifier of the vectorizer you want to drop|
```
cevian, line 1612: again having both in the same table is misleading

docs/vectorizer/api-reference.md
```diff
@@ -1576,19 +1726,26 @@ items.
...
+```
 
 #### Parameters
 
 `ai.vectorizer_queue_pending function` takes the following parameters:
 
 | Name          | Type | Default | Required | Description                                             |
 |---------------|------|---------|----------|---------------------------------------------------------|
-| vectorizer_id | int  | -       | ✔        | The identifier of the vectorizer you want to check      |
+| vectorizer_id | int  | -       | ✔*       | The identifier of the vectorizer you want to check      |
```
cevian, line 1743: misleading

docs/vectorizer/overview.md
```diff
@@ -150,7 +151,7 @@ into each chunk:
 SELECT ai.create_vectorizer(   
     'blog'::regclass,
     loading => ai.loading_column('contents'),
-    destination => 'blog_contents_embeddings',
+    destination => ai.destionation('blog_contents_embeddings'),
```
cevian, line 154:
```suggestion
    destination => ai.destination_table('blog_contents_embeddings'),
```

docs/vectorizer/overview.md
```diff
@@ -284,7 +285,7 @@ accordingly:
 SELECT ai.create_vectorizer(
     'blog'::regclass,
     loading => ai.loading_column('contents'),
-    destination => 'blog_contents_embeddings',
+    destination => ai.destination('blog_contents_embeddings'),
```
cevian:
```suggestion
    destination => ai.destination_table('blog_contents_embeddings'),
```

docs/vectorizer/overview.md
```diff
@@ -304,7 +305,7 @@ example uses a HNSW index:
...
```
cevian:
```suggestion
    destination => ai.destination_table('blog_contents_embeddings'),
```

docs/vectorizer-quick-start.md
Askir: that's a good catch I did a faulty find and replace 😓 thank you!

docs/vectorizer/overview.md
```diff
@@ -344,6 +345,56 @@ CREATE TABLE blog_contents_embeddings_store(
...
+```
+
+**When to use table destination:**
+- When you need multiple embeddings per row (chunking)
+- For large text fields that need to be split
+- You are vectorizing documents (which typically require chunking)
+
+### 2. Column Destination
+
+For simpler cases, you can add an embedding column directly to the source table:
```
cevian, line 377: can you move some of the language I suggested above down here? Let's make it crystal-clear this is the no-chunking case

examples/embeddings_from_documents/documents/pgai.md
```diff
@@ -120,7 +120,7 @@ Please note that using Ollama requires a large (>4GB) download of the docker ima
     ```sql
     SELECT ai.create_vectorizer(
          'wiki'::regclass,
-         destination => 'wiki_embeddings',
+         destination => ai.destionation('wiki_embeddings'),
```
cevian, line 123:
```suggestion
         destination => ai.destination_table('wiki_embeddings'),
```

examples/evaluations/litellm_vectorizer/README.md
```diff
@@ -54,7 +54,7 @@ The evaluation generates diverse question types (short, long, direct, implied, a
 
     SELECT ai.create_vectorizer(
         'paul_graham_essays'::regclass,
-        destination => 'essays_cohere_embeddings',
+        destination => ai.destination('essays_cohere_embeddings'),
```
cevian:
```suggestion
        destination => ai.destination_table('essays_cohere_embeddings'),
```

docs/vectorizer-quick-start.md
Askir: `ai.destination_table` is the function name that's equivalent to what we had before :D

examples/evaluations/ollama_vectorizer/README.md
```diff
@@ -61,7 +61,7 @@ Dataset Setup:
    SELECT ai.create_vectorizer(
       'pg_essays'::regclass,
       loading => ai.loading_column('text'),
-      destination => 'essays_nomic_embeddings',
+      destination => ai.destionation('essays_nomic_embeddings'),
```
cevian, line 64:
```suggestion
      destination => ai.destination_table('essays_nomic_embeddings'),
```

cevian (changes requested):

docs/vectorizer-quick-start.md
smoya: > `ai.destination_table` is the function name that's equivalent to what we had before :D

that's a good catch I did a faulty find ~and replace~ 😓 thank you!

😆

docs/vectorizer/api-reference.md
Askir: yeh I forgot, added!

docs/vectorizer/api-reference.md
Askir: okay I felt like two tables is more confusing than this, but changed it everywhere, let me know what you think.

cevian (approved):
some nits but approving

PR #643, opened 2025-04-23
chore: release extension 0.10.0

cevian:


docs/vectorizer/api-reference.md
```diff
@@ -81,6 +84,21 @@ Vectorizer offers the following APIs:
   performance of vectorizers.
 
 
+## Install or upgrade the database objects necessary for vectorizer
+
+You can install or upgrade the database objects necessary for vectorizer by running the following python code:
```
Askir, line 89: I'd mention the cli as well no?

docs/vectorizer/api-reference.md
```diff
@@ -81,6 +84,21 @@ Vectorizer offers the following APIs:
...
+
+```python
+import pgai
+
+pgai.install(DB_URL)
+```
```
Askir, line 95:
```suggestion
```python
import pgai

pgai.install(DB_URL)
```
Alternatively this cli command:
```bash
pgai install -d DB_URL
```

projects/extension/RELEASE_NOTES.md
```diff
@@ -1,6 +1,30 @@
...
+functionality to be used on more PostgreSQL cloud providers (AWS RDS, Supabase,
+etc.)
+
+We made this change in a way that will allow current users of the vectorizer to
+continue using the feature without interruption, but they will have to modify how they
+upgrade vectorizer functionality in the future. The upgrade to this extension
+version will detach the vectorizer catalog tables and functions from the
+extension but leave them in your database.  You can then manage them from the
+python library. In particular, you can then upgrade the vectorizer with
+`pgai.install(DB_URL)` as in the new python-library-based [workflow](/docs/vectorizer/api-reference.md#install-or-upgrade-database-dependencies).
```
Askir, line 18:
```suggestion
We made this change in a way that will allow current users of the vectorizer to
continue using the feature without interruption, but they will have to modify how they
upgrade vectorizer functionality in the future.

**Upgrading to 0.10.0:**
1. Run `ALTER EXTENSION name UPDATE TO '0.10.0'` to detach the vectorizer catalog tables and functions from the extension. This leaves them in your database in the ai schema, and the vectorizer will continue to work.
2. You can then manage them from the python library or cli by upgrading the vectorizer with `pgai install -d DB_URL` as described in the new python-library-based [workflow](/docs/vectorizer/api-reference.md#install-or-upgrade-database-dependencies).
3. If you don't use the model calling capabilities of pgai, you can then remove the pgai extension from your database.
```

Askir (approved):
I think we should focus on the cli since here since this will be more targeted at DB admins rather than application developers.

docs/vectorizer/api-reference.md
smoya: Definitely the cli command looks the defacto way for the users

PR #648, opened 2025-04-24
docs: update llms.txt

Askir:
The script doesn't 100% work anymore because it wasn't build for the world where pgai is just a library but oh well. Better than referencing files that don't exist. We can improve on it later.

### DIFF
--- a/llms.txt
+++ b/llms.txt
@@ -12,54 +12,43 @@ The project is maintained by Timescale, a PostgreSQL database company.
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


llms.txt
```diff
@@ -12,54 +12,43 @@ The project is maintained by Timescale, a PostgreSQL database company.
 
 ## Installation
```
JamesGuthrie, line 13: The installation section is empty now?

cevian (approved):
ok as a quick fix for now

llms.txt
Askir: Because they moved to the extension dir. Good catch, fixed that also in the script.

PR #651, opened 2025-04-25
chore: reduce default batch_size for loading_uri

Askir:
This wont work if you are using a BYTEA column for documents. Handling that would be a bit more complicated.
Also no idea how to write a test for this.

### DIFF
--- a/projects/pgai/pgai/vectorizer/vectorizer.py
+++ b/projects/pgai/pgai/vectorizer/vectorizer.py
@@ -905,6 +905,21 @@ class Executor:
 
             return len(items)
 
+    @cached_property
+    def _batch_size(self) -> int:
+        """Returns the batch size for processing.
+        Documents take way longer to process than simple text rows,
+        due to download and parsing overhead.
+        So when the vectorizer is processing documents
+        we use a smaller default batch size."""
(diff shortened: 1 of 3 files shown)


JamesGuthrie (approved):

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -905,6 +905,21 @@ async def _do_batch(self, conn: AsyncConnection) -> int:
...
+        due to download and parsing overhead.
+        So when the vectorizer is processing documents
+        we use a smaller default batch size."""
+        if self.vectorizer.config.processing.batch_size is not None:
+            return max(1, min(self.vectorizer.config.processing.batch_size, 2048))
+        else:
+            if isinstance(self.vectorizer.config.loading, UriLoading):
+                return 1
+            else:
+                return 50
```
smoya, line 921: Can we take this opportunity to document such a default values in https://github.com/timescale/pgai/blob/main/docs/vectorizer/api-reference.md#aiprocessing_default ? 🙏

cevian (approved):
approved but agree with Sergio to change docs

PR #671, opened 2025-04-29
feat: support aggregate functions

MasterOdin:
PR adds support for loading and rendering aggregate functions. The big annoyance is that as `pg_get_functiondef` errors out on aggregate functions, we need to manually put together the create DDL for the function using stuff from the [`pg_aggregate`](https://www.postgresql.org/docs/current/catalog-pg-aggregate.html) table.


projects/pgai/pgai/semantic_catalog/loader.py
jgpruitt, line 1: I'd like to circle back and try to make this query more readable, but I think it's fine for now

jgpruitt (approved):

PR #678, opened 2025-04-30
fix: quick-starts for new lib based workflow

Askir:


docs/vectorizer-quick-start.md
```diff
@@ -38,7 +38,19 @@ On your local machine:
...
+1. **Start the database**
+   ```shell
+    docker compose up -d db
+    ```
+
+1. **Install pgai in your database**
+   ```shell
+   docker run --rm --entrypoint python \
+   timescale/pgai-vectorizer-worker:latest \
+   -m pgai install --strict true -d "postgres://postgres:postgres@localhost:5432/postgres"
```
smoya, line 50: I believe you should use the name of the service instead 
```suggestion
   -m pgai install --strict true -d "postgres://postgres:postgres@db:5432/postgres"
```

docs/vectorizer-quick-start.md
```diff
@@ -38,7 +38,19 @@ On your local machine:
...
-1. **Start the services**
+1. **Start the database**
+   ```shell
+    docker compose up -d db
+    ```
+
+1. **Install pgai in your database**
+   ```shell
+   docker run --rm --entrypoint python \
+   timescale/pgai-vectorizer-worker:latest \
```
smoya, line 49: Why don't you use the container created by the docker compose? Instead a `docker compose run` command using the `vectorizer-worker` service

Askir: oh yeh that's a good idea! will change it

JamesGuthrie (approved):

PR #683, opened 2025-04-30
feat: add support for named vectorizers to ai.vectorizer_errors

smoya:


cevian:
I don't like having to materialize a name in this table. Maybe a better approach is to move the table to `ai._vectorizer_errors` and create a view `ai.vectorizer_errors` that does the join?

smoya:
> I don't like having to materialize a name in this table. Maybe a better approach is to move the table to `ai._vectorizer_errors` and create a view `ai.vectorizer_errors` that does the join?

Yeah, that sounds more clever 👍

projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -137,7 +137,7 @@ class Vectorizer(BaseModel):
     source_pk: list[PkAtt]
     queue_failed_table: str | None = None
     errors_schema: str = "ai"
-    errors_table: str = "vectorizer_errors"
+    errors_table: str = "_vectorizer_errors"
```
smoya, line 140: I'm not 100% convinced about this change. AFAIK `errors_table` is not configurable when creating a vectorizer yet. If we end up allowing this, we would need to create the error views for each configured errors_table.

projects/pgai/db/sql/incremental/030-add_vectorizer_errors_view.sql
```diff
@@ -0,0 +1,53 @@
...
+    and indexdef like '%id, recorded%';
+    
+  if _index_name is not null then
+    execute 'drop index if exists ai.' || quote_ident(_index_name);
+  end if;
+end
+$$;
+
+-- recreate the previous index to perform lookups by vectorizer id
+create index on ai._vectorizer_errors (id, recorded);
```
cevian, line 23: I would prefer renaming the index not recreating `ALTER INDEX...RENAME`

projects/pgai/db/sql/incremental/030-add_vectorizer_errors_view.sql
```diff
@@ -0,0 +1,53 @@
...
+        FROM information_schema.role_table_grants
+        WHERE table_schema = 'ai' 
+        AND table_name = 'vectorizer'
+    LOOP
+        to_user := rec.username;
+        EXECUTE 'GRANT SELECT ON ai._vectorizer_errors TO ' || quote_ident(to_user);
+        EXECUTE 'GRANT SELECT ON ai.vectorizer_errors TO ' || quote_ident(to_user);
+    END LOOP;
+END
+$$;
```
cevian, line 53: you don't need to add grants to the table as the rename transfers the grants. But you should apply any grants given to the table to the view. Notice in the code below I transfer the grants from _vectorizer_errors table 
```suggestion
-- grant privileges on _vectorizer_errors and vectorizer_errors to vectorizer users
DO $$
DECLARE
    to_user text;
    priv_type text;
    with_grant text;
    rec RECORD;
BEGIN
    -- find all users that have permissions on ai.vectorizer table and grant them to the errors ones
    FOR rec IN 
        SELECT DISTINCT grantee as username
        FROM information_schema.role_table_grants
        WHERE table_schema = 'ai' 
        AND table_name = '_vectorizer_errors'
    LOOP
        to_user := rec.username;
        priv_type := rec.privilege_type
        with_grant = ''
        if rec.is_grantable then
           with_grant = ' WITH GRANT OPTION';
        end if;
        EXECUTE format('GRANT %I ON ai.vectorizer_errors TO %I %s', priv_type, to_user, with_grant);
    END LOOP;
END
$$; 
```

projects/pgai/pgai/vectorizer/vectorizer.py
cevian: this change is correct

projects/pgai/db/sql/incremental/030-add_vectorizer_errors_view.sql
```diff
@@ -0,0 +1,53 @@
...
+create index on ai._vectorizer_errors (id, recorded);
+
+-- create a view including vectorizer name
+create or replace view ai.vectorizer_errors as
+select 
+  ve.*,
+  v.name
+from
+  ai._vectorizer_errors ve
+  join ai.vectorizer v on ve.id = v.id;
```
cevian, line 32: paranoia
```suggestion
  ai._vectorizer_errors ve
  left join ai.vectorizer v on ve.id = v.id;
```

cevian (changes requested):

projects/pgai/pgai/vectorizer/vectorizer.py
smoya: Yeah, technically it is at this moment. Just wanted to leave a statement here for reference.

projects/pgai/db/sql/incremental/030-add_vectorizer_errors_view.sql
smoya: The idea behind the previous block is to automatically find the name of the index instead of assuming the name pattern is not the autogenerated. But yeah, at this point, we know the name pattern so we can go directly with it and do a rename 👍

smoya:
All good to a last round of review 🚀

cevian (approved):
LGTM. As a foliowup we need upgrade tests

PR #685, opened 2025-04-30
support python 3.13

cevian:


cevian:
note github action run can be seen here with the new workflow https://github.com/timescale/pgai/actions/runs/14764818481

JamesGuthrie (approved):
LGTM!

PR #689, opened 2025-05-01
add import/export commands for yaml file

jgpruitt:


projects/pgai/pgai/cli.py
```diff
@@ -770,30 +770,119 @@ def load(
...
+    help="The path to a file to write log messages to.",
+)
+@click.option(
+    "--log-level",
+    type=click.Choice(
+        ["DEBUG", "INFO", "WARN", "ERROR", "FATAL", "CRITICAL"], case_sensitive=False
+    ),
+    default="INFO",
+)
+def export_catalog(
```
MasterOdin, line 842: Maybe add an option to export only certain objects from the catalog?

jgpruitt: yeah, let's do that in a follow-up. I'll add a ticket

PR #691, opened 2025-05-01
feat: return query plan with generated sql

jgpruitt:


projects/pgai/pgai/cli.py
```diff
@@ -245,7 +245,6 @@ def install(db_url: str, strict: bool) -> None:
     import pgai
 
     pgai.install(db_url, strict=strict)
-    log.info(f"pgai {__version__} installed")
```
MasterOdin, line 248: Did you mean to remove this given it was only added today in https://github.com/timescale/pgai/commit/a8f5f8d66d25ee93427617e60d06383101593d5b?

projects/pgai/pgai/semantic_catalog/gen_sql.py
```diff
@@ -270,12 +276,14 @@ def record_sql_statement(  # pyright: ignore [reportUnusedFunction]
                     error=error,
                 )
             else:
+                logger.info("the sql statement is valid")
```
MasterOdin, line 279: Won't this be printed as part of `validate_sql_statement` as well?

MasterOdin (approved):

projects/pgai/pgai/cli.py
jgpruitt: nope. good catch. thanks

projects/pgai/pgai/semantic_catalog/gen_sql.py
jgpruitt: yep. i'll remove

projects/pgai/pgai/cli.py
jgpruitt: stupid merge conflicts

PR #692, opened 2025-05-01
fix: better error for old extension version

cevian:


projects/pgai/pgai/_install/install.py
```diff
@@ -45,6 +50,18 @@ def _get_vector_extension_schema_sql() -> sql_lib.SQL:
     """)
 
 
+def get_ai_extension_version_sql() -> sql_lib.SQL:
+    return sql_lib.SQL("select extversion from pg_extension where extname = 'ai'")
+
+
+def check_extension_version_result(result: Any | None) -> None:
```
JamesGuthrie, line 57: I think we should rename this to make it clear that it raises:
```suggestion
def raise_on_old_extension_version(result: Any | None) -> None:
```

projects/pgai/pgai/_install/install.py
```diff
@@ -11,6 +12,10 @@
 
 log = structlog.get_logger()
 
+OLD_AI_EXTENSION_EXCEPTION = Exception(
+    "You have an old version of the ai extension installed. You must upgrade the ai extension to version 0.10.0 or greater before installing pgai"  # noqa: E501
```
JamesGuthrie, line 16: I would suggest that we include the old version in this error message, something like:

```suggestion
    "The ai extension is outdated and must be upgraded. Installed version: {version}, required version: {required_version}" # noqa: E501
```

This exception has one usage site, so I would suggest inlining it there.

JamesGuthrie (changes requested):

JamesGuthrie (approved):
LGTM, tests are failing though.

PR #700, opened 2025-05-05
fix: improve sc error message when pgai not installed

MasterOdin:
PR improves it so that when trying to run a command that requires the semantic catalog when pgai hasn't been installed, so that instead of the seeing the following exception message:

```
psycopg.errors.UndefinedTable: relation "ai.semantic_catalog" does not exist
LINE 2:             from ai.semantic_catalog
```

will now see:

```
AssertionError: Semantic catalog is not installed, please run: pgai semantic-catalog create
```

which provides actionable advice to the user on how to fix their issue.

### DIFF
--- a/projects/pgai/pgai/semantic_catalog/semantic_catalog.py
+++ b/projects/pgai/pgai/semantic_catalog/semantic_catalog.py
@@ -343,7 +343,24 @@ class SemanticCatalog:
         await async_export_to_yaml(yaml, load_from_catalog(catalog_con, self.id))
 
 
+async def _check_installed(con: CatalogConnection) -> None:
+    async with con.cursor() as cur:
+        await cur.execute(
+            """\
+            select count(*)
+            from information_schema.tables
+            where table_schema = %s and table_name = %s
+        """,
+            ("ai", "semantic_catalog"),
+        )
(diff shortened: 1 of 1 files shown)


projects/pgai/pgai/semantic_catalog/semantic_catalog.py
```diff
@@ -307,7 +307,24 @@ async def export_catalog(
...
+            ("ai", "semantic_catalog"),
+        )
+        row = await cur.fetchone()
+        assert (
+            row is not None and row[0] > 0
+        ), "Semantic catalog is not installed, please run: pgai semantic-catalog create"
+
+
 async def from_id(con: CatalogConnection, id: int) -> SemanticCatalog:
+    await _check_installed(con)
```
jgpruitt, line 327: Instead of running an extra query to actively check, can we catch a "relation does not exist" exception specifically?

MasterOdin: Changed in 8bdde3e8d6d67492b2313a3e768bbfa5de0bbf96 to show the message off `psycopg.errors.UndefinedTable`, which should give the same behavior, though it's not quite as DRY.

PR #702, opened 2025-05-05
Lychee PR

JamesGuthrie:


.github/workflows/check-markdown-links.yml
```diff
@@ -10,25 +10,10 @@ jobs:
...
-          path: .lycheecache
-          key: cache-lychee-${{ github.sha }}
-          restore-keys: cache-lychee-
-
       - name: Link Checker
         id: lychee
         uses: lycheeverse/lychee-action@v2
         with:
-          args: --concurrency 100 --cache --max-cache-age 1d --exclude-loopback --accept '403' --exclude "http://0.0.0.0:8000/*" --exclude "https://github.com/timescale/pgai/compare/pgai-v.*" --exclude "https://github.com/timescale/pgai/commit/*"  --verbose --no-progress ..
+          args: --exclude-loopback --accept '403' --exclude "http://0.0.0.0:8000/*" --exclude "https://github.com/timescale/pgai/compare/pgai-v.*" --exclude "https://github.com/timescale/pgai/commit/.*" --exclude "https://github.com/.*/.*/blob/.*" --verbose --no-progress ..
```
smoya, line 17:
```suggestion
          args: --exclude-loopback --accept '403' --exclude "http://0.0.0.0:8000/*" --exclude "https://github.com/timescale/pgai/compare/pgai-v.*" --exclude "https://github.com/timescale/pgai/commit/.*" --include "https://github.com/timescale/pgai/blob/.*" --exclude "https://github.com/.*/.*/blob/.*" --verbose --no-progress ..
```

JamesGuthrie: done

projects/extension/docs/model_calling/anthropic.md
```diff
@@ -12,7 +12,7 @@ Anthropic functions in pgai require an [Anthropic API key](https://docs.anthropi
 In production, we suggest setting the API key using an environment variable.
 During testing and development, it may be easiest to configure the key value
 as a [session level parameter]. For more options and details, consult the
-[Handling API keys](/docs/security/handling-api-keys.md) document.
+[Handling API keys](..//security/handling-api-keys.md) document.
```
smoya, line 15: Honest question: Is there a technical reason for mixing absolute paths (considering / as the root of our project) and relatives like this one?

JamesGuthrie: No, not really. There were two considerations that went through my mind when switching to relative paths like this:
1. The path becomes longer `/projects/extension/docs/security/handling-api-keys.md`
2. The relative link is less fragile w.r.t. documentation content moving around internally

smoya: I would go for either one or the other solution and not mix both. Moving files around becomes easier if you just need to find and replace one single path type rather than more than one.

smoya (approved):

PR #707, opened 2025-05-06
fix: missing implementation error when doing sc create

MasterOdin:
PR fixes doing `pgai semantic-catalog create`, such that the catalog will be created and you will not see the following error:

```
AssertionError: config is missing implementation specification
```

### DIFF
--- a/projects/pgai/pgai/cli.py
+++ b/projects/pgai/pgai/cli.py
@@ -661,7 +661,7 @@ def create(
     from pgai.semantic_catalog.vectorizer import embedding_config_from_dict
 
     d = dict(
-        provider=provider.lower(),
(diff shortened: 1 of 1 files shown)


projects/pgai/pgai/cli.py
```diff
@@ -661,7 +661,7 @@ def create(
     from pgai.semantic_catalog.vectorizer import embedding_config_from_dict
 
     d = dict(
-        provider=provider.lower(),
+        implementation=provider.lower(),
```
MasterOdin, line 664: I chose not to rename the option to this function as I think for a layperson, "provider" makes more sense than "implementation".

jgpruitt (approved):

PR #708, opened 2025-05-06
test: use pgai-db container instead of pgai-ext for db tests

smoya:


jgpruitt (dismissed):
These tests were designed to work with the build.py in the projects/pgai/db directory which creates a `pgai-ext` container. In CI, we use the `pgai-ext` container for all of these tests. If we want to change any of this, we should consider the whole setup.

smoya:
> These tests were designed to work with the build.py in the projects/pgai/db directory which creates a `pgai-ext` container. 

I see it builds and runs `pgai-db` instead 🤔 https://github.com/timescale/pgai/blob/6daa5021b54fdc6cdf643fdfa8d30fa19f79c403/projects/pgai/db/build.py#L180-L198

smoya:
> Weird. CI still references the `pgai-ext` container. How is this working?
> 
> I think we should consider a more comprehensive cleanup of this

But we use it in CI only for the `build-and-test-extension`, and the step that runs those are `build-and-test-pgai-db-module`. In other words, the changes made here don't apply to the extension but rather to the "db" module under [/projects/pgai/db](/projects/pgai/db)

JamesGuthrie (approved):
I haven't taken a very deep loop at this, but at the surface level it looks right.

I suspect that when we copy-pasted these files from the extension repository when setting up projects/pgai/db, we didn't change `pgai-ext` to `pgai-db`.

This isn't surfaced in CI because it runs everything in docker, so these paths are never hit.

PR #709, opened 2025-05-06
chore: reduce openai token limit

Askir:


JamesGuthrie (changes requested):
Please leave some information in the commit message about why we're making this change.

JamesGuthrie (approved):

PR #712, opened 2025-05-06
fix: use cwd for dotenv

cevian:
By default dotenv searches the python script path not cwd for the .env. We actually want the cwd.

### DIFF
--- a/projects/pgai/pgai/cli.py
+++ b/projects/pgai/pgai/cli.py
@@ -12,12 +12,12 @@ import click
 import psycopg
 import structlog
 from ddtrace import tracer
-from dotenv import load_dotenv
+from dotenv import find_dotenv, load_dotenv
 from pytimeparse import parse  # type: ignore
 
 from .__init__ import __version__
 
-load_dotenv()
+load_dotenv(dotenv_path=find_dotenv(usecwd=True))
 
 structlog.configure(wrapper_class=structlog.make_filtering_bound_logger(logging.INFO))
 log = structlog.get_logger()


projects/pgai/pgai/cli.py
```diff
@@ -12,12 +12,12 @@
...
 import structlog
 from ddtrace import tracer
-from dotenv import load_dotenv
+from dotenv import load_dotenv, find_dotenv
 from pytimeparse import parse  # type: ignore
 
 from .__init__ import __version__
 
-load_dotenv()
+load_dotenv(dotenv_path=find_dotenv(usecwd=True))
```
MasterOdin, line 20: Should we also look at the user's home directory as well for any dotenv files? Could make it easier to use pgai from anywhere on your computer, vs having to configure it per specific location.

jgpruitt (approved):

projects/pgai/pgai/cli.py
smoya: Correct me if I'm wrong @MasterOdin but that will force env vars to be uniquely named (e.g. prefixed always with PGAI_) in order to avoid setting env vars for other applications in the home dir that could be taken by pgai.

smoya (approved):

projects/pgai/pgai/cli.py
cevian: yeah let's not be too clever. This will already work if we are calling the program in a subdir of the home dir. I don't think we need any extra logic

PR #715, opened 2025-05-07
ci: install pgai when running dev env through compose-dev.yaml

smoya:
We didn't update the `compose-dev.yaml` file since the repackaging. That means we were not installing `pgai` in the DB and we were still asking the user to install the extension.

Additionally, we were using the wrong postgres PGDATA path. In particular, we were setting `/home/postgres/pgdata/data` instead of the default in postgres `/var/lib/postgresql/data` because is the one we use in [timescaledb-docker-ha](https://github.com/timescale/timescaledb-docker-ha/blob/master/Dockerfile#L437). 
That means data was never persisted in the host volume.

### DIFF
--- a/projects/pgai/compose-dev.yaml
+++ b/projects/pgai/compose-dev.yaml
@@ -16,10 +16,27 @@ services:
         required: false
       - path: ./db.env
         required: false
+  pgai-installer:
+    build: .
+    depends_on:
+      - db
+    entrypoint: sh
+    command:
+      - -c
+      - |
+        for i in $(seq 1 10); do
+          echo "Attempt $$i: installing pgai..."
+          # always overwriting previous installation here.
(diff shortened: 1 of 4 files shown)


projects/pgai/compose-dev.yaml
```diff
@@ -16,10 +16,27 @@ services:
...
+    depends_on:
+      - db
+    entrypoint: sh
+    command:
+      - -c
+      - |
+        for i in $(seq 1 10); do
+          echo "Attempt $$i: installing pgai..."
+          # always overwriting previous installation here.
+          python -m pgai install -d postgres://postgres:postgres@db:5432/postgres && exit 0
```
smoya, line 30: Not using `--strict` here as this is a dev env.

Askir (approved):
lgtm

PR #726, opened 2025-05-09
test: avoid modifying template1

cevian:
Modifying template1 has weird, unexpected effects on subsequent tests in other modules. I sunk more time than I want to admit to figuring out what was going on.


cevian:
Should fix tests in #692

JamesGuthrie (approved):

smoya (approved):

PR #730, opened 2025-05-12
fix: openai per batch limit

### LINKED ISSUE #728: [Bug]: OpenAI - Reaching max tokens per request. Count discrepancy between local count tokens and the count given by OpenAI API
## Description
We've identified a significant discrepancy between the token counting performed by the local `tiktoken` library and the actual token counting performed by the OpenAI Embeddings API. This issue causes pgai to unexpectedly hit max token limit per request when generating embeddings.

It seems some undocumented pessimistic estimation happens when the request reaches OpenAI API servers. Didn't find any official docs, just [this very brief issue in their forums](https://community.openai.com/t/max-total-embeddings-tokens-per-request/1254699).

## Details
When generating documents that reach approximately 300K tokens (the OpenAI embeddings API limit per request):

1. We use `tiktoken` encoder to count locally, we generate as close as possible to 300K tokens.
2. When these documents are sent to the OpenAI API, it reports a much higher token count (approximately 50% more)
3. This results in errors like: `openai.BadRequestError: Error code: 400 - {'error': {'message': 'Requested 454148 tokens, max 300000 tokens per request', 'type': 'max_tokens_per_request', 'param': None, 'code': 'max_tokens_per_request'}}`

**Key Observation**: The discrepancy appears to be related to the length of the input array, not just only the total token count. When sending fewer items in the array (even when approaching the same token limit), the discrepancy is much smaller or nonexistent. The more items in the input array, the more pessimistic the API's token estimation becomes compared to local counting.

## Impact

1. When large datasets are used for vector embeddings, we might hit unexpected rate limits or token limits
2. The pgai vectorizer might fail to process large documents that appear to be under the token limit based on local counting
3. Users may need to implement more complex chunking strategies than expected based solely on local token counts

## Potential Solutions

1. Add a safety margin. Like applying some coefficient to the max token limit when using local counting.   
2. Reduce the number of items in input arrays by consolidating smaller inputs into larger chunks, since fewer array items appears to result in less token count inflation.
3. Any other I can't think atm.

Regardless, we might need to update our documentation to warn users about this discrepancy.

### pgai extension affected

_No response_

### pgai library affected

_No response_

### PostgreSQL version used

17

Any

Not applicable

Not applicable

```bash

```

```bash
import tiktoken, openai

encoder = tiktoken.encoding_for_model("text-embedding-ada-002")

raw_inputs = [
    # long
    "This is the larger Lorem ipsum. Lorem ipsum dolor sit amet consectetur adipiscing elit rutrum eros ultricies, convallis porttitor commodo purus urna id duis mattis. Dictum pellentesque sollicitudin dignissim quam in eu scelerisque donec neque, fames metus vestibulum torquent sed erat velit facilisis magna vivamus, mauris nam cras aptent euismod sagittis quis sociis. Eleifend imperdiet commodo nisl phasellus non sodales nisi interdum blandit, turpis luctus viverra erat vivamus ac mus et, sem hendrerit dis natoque sollicitudin conubia eros himenaeos.",
    
    # short
    "This is shorter Lorem ipsum doc. Lorem ipsum dolor sit amet consectetur adipiscing elit rutrum eros ultricies, convallis porttitor commodo purus urna id duis mattis."
]

for template_index, doc_base in enumerate(raw_inputs):
    print(f"\n----- input {template_index + 1} -----")
    print(f"Input: {doc_base[:40]}...")
    
    inputs = []
    encoded_inputs = []
    encoded_length = 0
    
    while encoded_length < 300000:
        new_input = doc_base
        encoded = encoder.encode(new_input)
        new_tokens = len(encoded)
        
        if encoded_length + new_tokens <= 300000:
            inputs.append(new_input)
            encoded_inputs.append(encoded)
            encoded_length += new_tokens
        else:
            # close enough, stop adding
            break
    
    
    print(f"Local calculated tokens: {encoded_length}")
    print(f"Number of inputs: {len(inputs)}")
    
    try:    
        resp = openai.embeddings.create(model="text-embedding-3-small", input=inputs) 
        print(f"API counted: {resp.usage.prompt_tokens} tokens")
        
    except Exception as e:
        print(f"Request errored: {e}")
```

None

Askir:
Fixes #728 

The magic here is that openai has a token estimator in front of their actual API this one simply counts utf-8 bytes and for each one calculates 0.25 tokens.

This is what the 300k limit per batch request limit applies to. The actual tokens don't matter. A request with 1 million tokens and 1.2 million bytes will go through just fine as the estimator thinks this is exactly 300k "tokens".
Likewise a request with 200 000 tokens but 1.5 million bytes will fail.


JamesGuthrie (commented):
I would prefer it if we would not duplicate the implementation of `batch_indices`, and would instead make the existing `batch_indices` take an optional `estimated_chunk_token_lengths` parameter.

projects/pgai/pgai/vectorizer/embedders/openai.py
```diff
@@ -150,6 +150,18 @@ async def call_embed_api(self, documents: list[str]) -> EmbeddingResponse:
...
+        Estimates token count based on UTF-8 byte length.
+        """
+
+        total_estimated_tokens = 0
+        for char in document:
+            byte_length = len(char.encode("utf-8"))
+            total_estimated_tokens += byte_length * 0.25  # 0.25 tokens per byte
+
+        return total_estimated_tokens
+
```
JamesGuthrie, line 164: I would suggest that we do this, and then remove the type changes that this propagates:

```suggestion
    def _estimate_token_length(self, document: str) -> int:
        """
        Estimates token count based on UTF-8 byte length.
        """

        total_estimated_tokens = 0
        for char in document:
            byte_length = len(char.encode("utf-8"))
            total_estimated_tokens += byte_length * 0.25  # 0.25 tokens per byte

        return ceil(total_estimated_tokens)
```

Askir: I don't think that works because that would mean it rounds each individual documents tokens not the whole batch. The api however counts the bytes of everything first and then rounds.

E.g. if you send 100 batches of just "a" it would be estimated with 100 tokens this way but the api only assigns 25 (which makes no sense but well that's how it works)

JamesGuthrie: Hmmm... interesting. I didn't realise that on first inspection.

JamesGuthrie (approved):

PR #733, opened 2025-05-13
test: add image ocr test case

smoya:
I realized we had an actual png image in the tests directory but we didn't add the test for OCR.


smoya:
Closed in favor of https://github.com/timescale/pgai/pull/743

PR #735, opened 2025-05-14
chore: revert "feat: add support for named vectorizers to ai.vectorizer_errors"

smoya:
Reverts timescale/pgai#683

We are reverting this because it will cause a BC with previous versions that are relying on `ai.vectorizer_errors` table and don't have the new `ai._vectorizer_errors`. 

I'm working on a fix for this, in the meantime let's revert it so we don't block the release pipeline.


JamesGuthrie:
Can you explain why we're reverting this?

smoya:
> Can you explain why we're reverting this?

Updated the description.

Askir (approved):

PR #751, opened 2025-05-16
docs: document how chunk overlap works

Askir:


docs/vectorizer/api-reference.md
```diff
@@ -617,6 +617,15 @@ You use it to recursively split text into chunks using multiple separators.
 
 A JSON configuration object that you can use in [ai.create_vectorizer](#create-vectorizers).
 
+### Understanding Overlap
+The recursive character text splitter creates overlaps only when splitting within words (using the empty string separator "").
+When splitting at explicit separators like periods or spaces, chunks connect with no overlap.
+Always include "" as the final separator or chunks may exceed the specified size.
+The chunk_overlap parameter is a target, not a guarantee, and actual overlap sizes will vary based on where splits occur.
```
smoya, line 624:
```suggestion
The `chunk_overlap` parameter is a target, not a guarantee, and actual overlap sizes will vary based on where splits occur.
```

docs/vectorizer/api-reference.md
```diff
@@ -617,6 +617,15 @@ You use it to recursively split text into chunks using multiple separators.
...
+
+Both chunking implementations use langchain's Text splitters under the hood: [[1]](https://python.langchain.com/docs/how_to/character_text_splitter/), [[2]](https://python.langchain.com/docs/how_to/recursive_text_splitter/).
+
+
```
smoya, line 628: WDYT about adding an example here?

docs/vectorizer/api-reference.md
```diff
@@ -617,6 +617,15 @@ You use it to recursively split text into chunks using multiple separators.
 
 A JSON configuration object that you can use in [ai.create_vectorizer](#create-vectorizers).
 
+### Understanding Overlap
+The recursive character text splitter creates overlaps only when splitting within words (using the empty string separator "").
```
JamesGuthrie, line 621: Is this statement correct? In the following example, the "" splitter is not used, but there is still overlapping text:

```python
text_splitter = RecursiveCharacterTextSplitter(separators=["\n\n", "\n", " ", ""],chunk_size=20,chunk_overlap=10)
text_splitter.create_documents(["Lorem Ipsum is simply dummy text of the printing and typesetting industry."])
```
Produces:
```
['Lorem Ipsum is', 'Ipsum is simply', 'is simply dummy', 'dummy text of the', 'of the printing and', 'and typesetting', 'industry.']
```

Askir: Your code actually contains the "" splitter. But I tried without it and you're correct that it still produces overlap. I can't quite figure out how this works. Some chunks seems to have overlap and some don't. Maybe linking to langchain is enough.

PR #752, opened 2025-05-16
chore: reproduce error when deleting during processing

Askir:


projects/pgai/tests/vectorizer/cli/test_vectorizer_core.py
```diff
@@ -514,3 +514,115 @@ def test_chunking_none(
...
+        assert (
+            source_row["count"] == 0
+        ), f"Source table should be empty, was {source_row['count']}"
+        # Check that the embedding table is empty (no dead chunks)
+        cur.execute(f"SELECT COUNT(*) FROM {embedding_table_name}")
+        embedding_row = cur.fetchone()
+        assert embedding_row is not None
+        assert (
+            embedding_row["count"] == 0
+        ), f"Embedding table should be empty, was {embedding_row['count']}"
```
Askir, line 628: This assert fails but the one above succeeds meaning you end up with dead chunks.
It's a bit iffy not sure if this will work in CI because the timing has to be configured correctly for the delete to be executed at the right time.

PR #760, opened 2025-05-21
feat(extension): support named vectorizers in ai.execute_vectorizer

smoya:


projects/extension/ai/vectorizer.py
```diff
@@ -28,7 +28,29 @@ def execute_vectorizer(plpy, vectorizer_id: int) -> None:
...
+    try:
+        plan = plpy.prepare("""
+            select pg_catalog.to_jsonb(v) as vectorizer
+            from ai.vectorizer v
+            where name = $1
+            limit 1
+        """, ["text"])
+        result = plan.execute([vectorizer_name], 1)
+    except Exception as e:
+        if 'column "name" does not exist' in str(e):
```
smoya, line 43: An alternative to this would be to check for the column via `information_schema.columns` table, but that will mean we would be doing one extra query always, even in the case where `name` column is presence (last versions of pgai lib)

projects/pgai/db/tests/vectorizer/test_named_vectorizer.py
```diff
@@ -9,18 +11,7 @@ def db_url(user: str) -> str:
...
-    ) as con:
-        with con.cursor() as cur:
-            cur.execute("create extension if not exists timescaledb")
-            cur.execute("select to_regrole('bob') is null")
-            if cur.fetchone()[0] is True:
-                cur.execute("create user bob")
-            cur.execute("select to_regrole('adelaide') is null")
-            if cur.fetchone()[0] is True:
-                cur.execute("create user adelaide")
+def setup_website_blog_table():
```
smoya, line 14: Did a little bit of refactor here to reuse some logic across tests.

PR #765, opened 2025-05-28
feat: add sc api to manage facts

MasterOdin:
PR adds an API for managing facts in the semantic catalog so that you can:
1. add a fact
2. drop an existing fact
3. edit an existing fact

The expectation would be that for (1) and (3) that the end user would run the `vectorize` function afterwards.

Assuming we're fine with this, I'd then go to add API for objects and sql examples, but wanted to start with just one for now.


projects/pgai/pgai/semantic_catalog/semantic_catalog.py
```diff
@@ -410,6 +410,121 @@ async def search_sql_examples(
...
+                )
+            """,
+                dict(catalog_name=self.name, description=description),
+            )
+            row = await cur.fetchone()
+            if row is None:
+                raise RuntimeError("Failed to add fact to semantic catalog")
+            return Fact(id=row["sc_add_fact"], description=description)
+
+    async def edit_fact(
```
jgpruitt, line 473: let's call this `update_fact`.

projects/pgai/pgai/semantic_catalog/semantic_catalog.py
```diff
@@ -410,6 +410,121 @@ async def search_sql_examples(
...
+                    from ai.semantic_catalog_embedding e
+                    where e.semantic_catalog_id = %(catalog_id)s
+                """,
+                dict(catalog_id=self.id),
+            )
+            row = await cur.fetchone()
+            count = row["count"] if row else 0
+            cols: list[Composable] = []
+            for i in range(count):
+                cols.append(SQL("{} = null").format(Identifier(f"emb{i + 1}")))
```
jgpruitt, line 507: This is not necessary. There is a trigger on the fact table that will null out the vector columns when the row is updated

jgpruitt (changes requested):

projects/pgai/pgai/semantic_catalog/semantic_catalog.py
MasterOdin: Good to know, removed in 64a8744f5516ae72eacf6e6c2d1955a3591cf01b.

projects/pgai/pgai/semantic_catalog/semantic_catalog.py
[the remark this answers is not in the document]
MasterOdin: Yeah, agreed. I just wanted to make sure I had buy-in from you on the basic approach before tackling those.

PR #774, opened 2025-05-30
feat: log file size

alejandrodnm:
I'm not really sure about this since it adds an extra call to s3. Maybe we can put it behind a debug flag? Although I think it might be useful while we make the feature more robust, specially in cloud.

### DIFF
--- a/projects/pgai/pgai/vectorizer/loading.py
+++ b/projects/pgai/pgai/vectorizer/loading.py
@@ -3,8 +3,11 @@ from dataclasses import dataclass
 from io import BytesIO
 from typing import Any, Literal
 
+import structlog
 from filetype import filetype  # type: ignore
 from pydantic import BaseModel
+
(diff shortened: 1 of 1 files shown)


projects/pgai/pgai/vectorizer/loading.py
```diff
@@ -80,11 +84,17 @@ def load(self, row: dict[str, str]) -> LoadedDocument:
...
-        content = BytesIO(
-            smart_open.open(  # type: ignore
-                file_path, "rb", transport_params=transport_params
-            ).read()
-        )
+            s3_resource = session.resource("s3")  # type: ignore
+        if file_path.startswith("s3://") and not self.aws_role_arn:
+            import boto3
+
+            s3_resource = boto3.resource("s3")  # type: ignore
```
JamesGuthrie, line 91: What is the purpose of this change?

PR #776, opened 2025-06-02
feat: support named vectorizers in ai.execute_vectorizer

smoya:
This PR adds a new signature for the `ai.execute_vectorizer` function that accepts a vectorizer name as input instead of an id.

### DIFF
--- a/projects/pgai/db/tests/vectorizer/test_named_vectorizer.py
+++ b/projects/pgai/db/tests/vectorizer/test_named_vectorizer.py
@@ -1,3 +1,5 @@
+import os
+
 import psycopg
 import pytest
 from psycopg.rows import namedtuple_row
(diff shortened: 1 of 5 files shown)


projects/pgai/db/tests/vectorizer/test_named_vectorizer.py
```diff
@@ -51,9 +37,50 @@ def test_named_vectorizer():
                 , ('how to make stir fry', '2022-01-06'::timestamptz, 'pick up the phone and order takeout')
             """)
 
-            # drop the drop_me column
-            cur.execute("alter table website.blog drop column drop_me")
 
+@pytest.mark.skipif(os.getenv("PG_MAJOR") == "15", reason="does not support pg15")
```
cevian, line 41: why does this not work on pg15?

cevian (approved):
one question about tests but non-blocking

projects/pgai/db/tests/vectorizer/test_named_vectorizer.py
smoya: Not sure if more errors will raise besides that one

PR #777, opened 2025-06-02
chore: switch to the low-level pydantic api for greater control

jgpruitt:
- switch to low-level "direct model request" API in pydantic ai for greater control
- disallow text replies from the LLM (force it to use a tool)
- revoke the search tool on the last iteration (force it to respond with a SQL answer)
- excludes prior matches from the semantic search so that context has a better chance of expanding
- includes all descriptions associated with a table/view in the prompt even when only a subset were matched semantically
- increases iteration limit from 5 to 10
- adds columns to semantic catalog tables to track usage of described items
- lists in the prompt the top N most frequently used objects


projects/pgai/pgai/semantic_catalog/search.py
```diff
@@ -31,23 +32,31 @@ async def search_objects(
...
             from ai.{table} x
-            order by x.{column} <=> %s::vector({dimensions})
-            limit %s
+            {filter}
+            order by x.{column} <=> %(query)s::vector({dimensions})
+            limit %(limit)s
         """).format(
             table=Identifier(f"semantic_catalog_obj_{catalog_id}"),
-            dimensions=SQL(str(int(config.dimensions))),  # pyright: ignore [reportArgumentType]
+            dimensions=Literal(int(config.dimensions)),
```
MasterOdin, line 53: Is doing the `int` cast necessary given that `config.dimensions` is typed as being an int and pydantic should error if any other type is assigned to it?

```suggestion
            dimensions=Literal(config.dimensions),
```

projects/pgai/pgai/semantic_catalog/search.py
```diff
@@ -75,23 +85,31 @@ async def search_sql_examples(
...
             select x.*
             from ai.{table} x
-            order by x.{column} <=> %s::vector({dimensions})
-            limit %s
+            {filter}
+            order by x.{column} <=> %(query)s::vector({dimensions})
+            limit %(limit)s
         """).format(
             table=Identifier(f"semantic_catalog_sql_{catalog_id}"),
             dimensions=SQL(str(int(config.dimensions))),  # pyright: ignore [reportArgumentType]
```
MasterOdin, line 106: To match above?

```suggestion
            dimensions=Literal(config.dimensions),
```

projects/pgai/pgai/semantic_catalog/search.py
```diff
@@ -119,23 +138,31 @@ async def search_facts(
...
             from ai.{table} x
-            order by x.{column} <=> %s::vector({dimensions})
-            limit %s
+            {filter}
+            order by x.{column} <=> %(query)s::vector({dimensions})
+            limit %(limit)s
         """).format(
             table=Identifier(f"semantic_catalog_fact_{catalog_id}"),
-            dimensions=SQL(str(int(config.dimensions))),  # pyright: ignore [reportArgumentType]
+            dimensions=Literal(int(config.dimensions)),
```
MasterOdin, line 159: Same here:

```suggestion
            dimensions=Literal(config.dimensions),
```

projects/pgai/pgai/semantic_catalog/gen_sql.py
```diff
@@ -554,127 +719,166 @@ async def generate_sql(
...
+                                    "asking llm to fix the invalid sql statement"
+                                )
+                                _continue = True
+                    else:
+                        logger.error(
+                            f"unrecognized tool name: {tool_call_part.tool_name}"
+                        )
+                case "text":
+                    text_part: TextPart = part
+                    logger.info(f"model response: {text_part.content}")
```
MasterOdin, line 872: As we're specifying `allow_text_output=False,` to the request options, hitting this case _should_ be impossible, right?

projects/pgai/pgai/semantic_catalog/gen_sql.py
```diff
@@ -554,127 +719,166 @@ async def generate_sql(
...
+                case _:
+                    logger.error(f"unrecognized model response kind: {part.part_kind}")
 
     return GenerateSQLResponse(
-        sql_statement=answer,
+        sql_statement=answer or "MISSING",
```
MasterOdin, line 877: Is this just to avoid a type error since I _think_ that the only way we should exit the loop above is when `type(answer) == str` or we've raised an `IterationLimitExceededException`?

If so, I think that we should just add a check above this that if `answer is None` then raise a `RuntimeError` as that would be an error within the library vs something we want downstream users to have to worry about.

projects/pgai/pgai/semantic_catalog/gen_sql.py
```diff
@@ -554,127 +719,166 @@ async def generate_sql(
...
+                            for irrelevant_id in (
+                                ctx.objects.keys() - relevant_object_ids
+                            ):
+                                ctx.objects.pop(irrelevant_id)
+                                ctx.rendered_objects.pop(irrelevant_id)
+                            await _increment_obj_usage(
+                                catalog_con,
+                                catalog_id,
+                                [id for id in ctx.objects],
+                            )
```
MasterOdin, line 829: Would it make sense to wait to do this block after we'e confirmed that there's no syntax errors in the query?

Minimally I think we should wait to do any increment till after we've verified that.

projects/pgai/pgai/semantic_catalog/gen_sql.py
```diff
@@ -437,18 +454,173 @@ async def initialize_database_context(
...
+                "command_type",
+                "relevant_object_ids",
+                "relevant_sql_example_ids",
+                "relevant_fact_ids",
+            ],
+        },
+    )
+
+
+async def _increment_obj_usage(
```
MasterOdin, line 565: I think the `increment_*_usage` functions should be methods off the `SemanticCatalog` class. I could see external applications tying into this concept, e.g. in PopSQL we track usage for showing "popular" tables, and so we might want to also increment the semantic-catalog usage counters as well.

projects/pgai/pgai/semantic_catalog/gen_sql.py
jgpruitt: eb3d0daa82733ae52115567b2e8fde045f89c1dd

projects/pgai/pgai/semantic_catalog/gen_sql.py
jgpruitt: I created a ticket for this to do later

PR #779, opened 2025-06-03
fix: incorrect error of column does not exist

alejandrodnm:
If the loading column had a unsupported type the error message that was being shown was related to the column not existing. Now an invalid type error messages is shown with the supported types.

### DIFF
--- a/projects/pgai/db/tests/vectorizer/test_vectorizer.py
+++ b/projects/pgai/db/tests/vectorizer/test_vectorizer.py
@@ -201,7 +201,9 @@ def psql_cmd(cmd: str) -> str:
     return str(proc.stdout).strip()
 
 
-@pytest.mark.skipif(os.getenv("PG_MAJOR") == "15", reason="does not support pg15")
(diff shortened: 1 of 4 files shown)


JamesGuthrie (commented):
Can you add a test for this as well?

projects/pgai/db/tests/vectorizer/test_vectorizer.py
```diff
@@ -201,7 +201,9 @@ def psql_cmd(cmd: str) -> str:
     return str(proc.stdout).strip()
 
 
-@pytest.mark.skipif(os.getenv("PG_MAJOR") == "15", reason="does not support pg15")
```
alejandrodnm, line 204: I don't understand these changes, because the format that's red is not even what's in `main` so 🤷🏼

JamesGuthrie: you need to rebase your change against main

PR #797, opened 2025-06-06
chore: update dependencies

JamesGuthrie:


JamesGuthrie:
I don't think that any of them are significant enough to justify a release.

PR #800, opened 2025-06-10
chore: split deps into separate extras

Askir:
This splits the worker and semantic-catalog into 2 separate extras. This setup would allow to install pgai via `pip install pgai` and use it to install all the SQL via `pgai install -d ...`

If then you try to run `pgai vectorizer worker` you'll get an error:
```
vectorizer-worker extra is not installed, please install it with `pip install pgai[vectorizer-worker]`
```

Same for any semantic catalog command.

The error messages use rich and tiktoken to decide if the extra was installed, since it is non-trivial/not cleanly possible to determine if an extra is installed, pip doesn't really store this metadata.

Note: This PR does unintentionally update a bunch of dependencies. I don't really know how to prevent that since uv doesn't have a command to move an existing dependency between extras. Since tests still pass I think we are fine though.


docs/semantic_catalog/quickstart.md
```diff
@@ -92,7 +92,7 @@ Once the semantic catalog is loaded with embedded descriptions, you start genera
    ```bash
    git clone https://github.com/timescale/pgai.git -b jgpruitt/semantic-catalog
    cd pgai/projects/pgai
-   uv sync
+   uv sync --all-extras
```
smoya, line 95: do you need all extras?

smoya:
I like the change so far.

Askir:
@smoya @JamesGuthrie Do you think we should proceed with this? If yes, what is missing?

alejandrodnm (approved):

smoya (approved):

smoya:
> @smoya @JamesGuthrie Do you think we should proceed with this? If yes, what is missing?

Happy proceeding 👍

PR #826, opened 2025-06-17
test: add upgrade tests for pgai lib

smoya:


alejandrodnm (approved):

smoya:
> Should we keep the latest state in version control? That's being my experience with working with snapshots, that way you can see the changes that are being added.

Can be a good thing, however my concern is that we will forgot at some point untracking old versioned files. So mixed feelings. WDYT?

PR #828, opened 2025-06-18
chore: remove git lfs

alejandrodnm:


projects/extension/tests/test_litellm.py
```diff
@@ -37,7 +37,7 @@ def cur() -> psycopg.Cursor:
         "name": "cohere/embed-english-v3.0",
         "dimensions": 1024,
         "api_key_name": "COHERE_API_KEY",
-        "exception": """CohereException - {"message":"no api key supplied"}""",
```
alejandrodnm, line 40: This exception changed and it includes `"id": "foobar"` in the body.

PR #843, opened 2025-07-02
fix: skip empty strings when batching documents

alejandrodnm:


projects/pgai/pgai/vectorizer/vectorizer.py
```diff
@@ -1188,13 +1188,19 @@ async def _generate_embeddings(
                     loading_errors.append((item, (LoadingError(e=e))))
                 continue
 
+            if isinstance(payload, str) and not payload.strip():
```
smoya, line 1191: I believe whitespace-only strings should be taken as they are and not skip them.
Additionally, whitespace-only strings don't break the API call to openai in this case.

JamesGuthrie (approved):

PR #867, opened 2025-08-18
Treat oid's as bigint

feikesteenbergen:
oids are 4 byte unsigned integers, and oids > 2^31 will cause errors when these functions are called.

Oids > 2^31 are a regular occurence in PostgreSQL, especially when temporary tables are used, or a lot of chunks/partitions are created and dropped.


MasterOdin:
This was superseded by #868 and no longer needed I think.

PR #879, opened 2025-10-07
fix: make file extension guessing with uri query parameters

smoya:


projects/pgai/pgai/vectorizer/loading.py
```diff
@@ -15,13 +16,19 @@ class LoadedDocument:
 
 
 def guess_filetype(file_like: BytesIO, file_path: str | None = None) -> str | None:
-    guess = filetype.guess(file_like)  # type: ignore
+    extension = filetype.guess_extension(file_like)  # type: ignore
```
smoya, line 19: This is a simplification. Calling `guess_extension` calls `guess` internally and just returns the extension, which is all we need.

projects/pgai/pgai/vectorizer/loading.py
```diff
@@ -15,13 +16,19 @@ class LoadedDocument:
...
+    if extension is None:
         if file_path is None:
             return None
-        return file_path.split(".")[-1]
-    return guess.extension
+        try:
+            parsed = urlparse(file_path)
+        except Exception:
+            return None
+        split = parsed.path.split(".")
```
smoya, line 28: As we use urlparse, this give us just the path without the query parameters.

projects/pgai/pgai/vectorizer/loading.py
```diff
@@ -15,13 +16,21 @@ class LoadedDocument:
 
 
 def guess_filetype(file_like: BytesIO, file_path: str | None = None) -> str | None:
-    guess = filetype.guess(file_like)  # type: ignore
+    extension = filetype.guess_extension(file_like)  # type: ignore
```
smoya, line 19: This is a simplification. Calling guess_extension calls guess internally and just returns the extension, which is all we need.

cevian (approved):

projects/pgai/pgai/vectorizer/loading.py
```diff
@@ -15,13 +16,22 @@ class LoadedDocument:
 
 
 def guess_filetype(file_like: BytesIO, file_path: str | None = None) -> str | None:
-    guess = filetype.guess(file_like)  # type: ignore
+    data = file_like.read()
```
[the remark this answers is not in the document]
smoya: Yeah, my attempt here was to make everything more explicit. But at the end, this will imply reading the headers using the same lenght as the library, and it is redundant and not very maintainable.
I'm reverting this change.

projects/pgai/pgai/vectorizer/loading.py
```diff
@@ -15,13 +16,21 @@ class LoadedDocument:
 
 
 def guess_filetype(file_like: BytesIO, file_path: str | None = None) -> str | None:
-    guess = filetype.guess(file_like)  # type: ignore
+    guess = filetype.guess(file_like)  # type: ignore[reportUnknownArgumentType,reportUnknownMemberType]
```
smoya, line 19: This error wasn't popping previously but now it does. Im just skipping. Other solution would be to do explicit casts.

projects/pgai/pgai/vectorizer/loading.py
```diff
@@ -15,13 +16,21 @@ class LoadedDocument:
...
+    if file_path is None:
+        return None
+
+    try:
+        parsed = urlparse(file_path)
+        _, ext = os.path.splitext(parsed.path)
+    except Exception:
+        return None
+
+    return ext[1:].lower() if ext else None
```
smoya, line 33: Added the `.lower()` so we ensure always to go with lowercase extensions. Just for consistency.

arajkumar (approved):

PR #898, opened 2025-11-16
vectorizers(litellm): Support setting max chunks per batch using env var

NickCrews:
I am using gemini as my provider. This wasn't listed as one of the providers, so I found I was getting the default of 5, and a warning message. This was slow.

So I added gemini to the code so it was supported.

But, more in general, I don't want some other user to be forced to modify the code in order to support their provider, so now you can configure this param with even more control using an env var


alejandrodnm:
Hey @NickCrews thanks for the PR. I'll make some time to review. Can you please rebase against main. I was fixing some things for pg18.

